PR #185, opened 2026-03-18
Add validation workflows for accessibility review

ewheeler:
This PR adds packaged accessibility assessment and a first-class Validation step to ADT Studio. It surfaces book-level accessibility summary/config in Validation, adds Preview accessibility inspection with highlighting, and moves accessibility out of the Debug pane. It also adds reviewer validation sessions, per-page checklists, progress summaries, smart N/A defaults, and resume-review flows across Preview and Validation. Upstream template-caused accessibility findings are intentionally left for a follow-up PR.

### DIFF
--- a/apps/api/src/routes/reviewer-validation.test.ts
+++ b/apps/api/src/routes/reviewer-validation.test.ts
@@ -10,6 +10,18 @@ import { createReviewerValidationRoutes } from "./reviewer-validation.js"
 
 const label = "validation-book"
 
+
+function enableReviewerValidation(configPath: string) {
+  fs.writeFileSync(configPath, [
+    "text_types:",
+    "  heading: Heading",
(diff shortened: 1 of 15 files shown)


nicpottier:
I pushed a few UI tweaks. I think we are close.

Couple more things:
 * config should live under a gear icon like it does for the other stages
 * do we want to show more details of the validation errors on the summary page. Maybe show which pages they occur on? 
  * from the validation page I also kind of want to be able to click on each of the severity levels and see explicitly what/where those errors are. Otherwise I'm left kind of wondering what all of them are?
  
  Otherwise I think we are good with the caveat that ya, we want to tune this and fix our own outputs so that hopefully we don't have any errors here? (or at least none we can't show the user how to fix?)

PR #268, opened 2026-04-08
Split Translate and Speech into separate stages with TTS word timestamps

elasticsounds:
## Summary
- **Split Translate/Speech stages**: The monolithic Translate stage is now two independent pipeline stages (Translate and Speech) with separate DAG nodes, settings panels, and run controls
- **Word-level timestamps**: Whisper-based word timestamp generation with inline playback highlighting, editable multi-column timecode tables with custom TimecodeInput controls, and background task queue with progress tracking
- **Whisper accuracy**: Source text passed as `prompt` parameter to Whisper API for better transcription alignment
- **Language picker**: Country lists updated to show only linguistically relevant countries when browsing; all countries still searchable by typing
- **UI polish**: Visual separator between run card controls and config sections; task progress messages displayed in sidebar

## Test plan
- [ ] Verify Translate and Speech stages appear independently in the sidebar and can be run separately
- [ ] Run Speech generation, then calculate timestamps — confirm background task shows progress (e.g. 42/632)
- [ ] Play audio and verify word-by-word highlighting syncs with playback
- [ ] Expand timestamp viewer, edit timecodes, save — confirm edits persist
- [ ] Re-run translations and verify pipeline completes without errors
- [ ] Open language picker and confirm only relevant countries appear per language


gbergengruen:
@elasticsounds, I saw that the language and speech steps are independent now, but you need to use the language step to run the speech step, as it seems to do some required step to generate the tts. The timestamps are amazing. I loved it.

PR #271, opened 2026-04-09
Fix MCQ validation, styling, consistency, and add multi-group support

elasticsounds:
## Summary
- Fix answer validation bug where string `"false"` was treated as truthy, causing all options to show as correct
- Replace destructive `className` replacement (`setLetterAppearance`) with non-destructive `classList` toggling (`setOptionState`) to prevent layout instability when selecting/validating options
- Add self-healing JS that auto-patches missing `.activity-option` classes and `role="radiogroup"` attributes on LLM-generated HTML
- Simplify MCQ prompt to a minimal functional skeleton, giving the LLM creative freedom on non-interactive elements while locking down required interactive structure
- Add cross-page visual consistency pass: a second LLM call rewrites subsequent activity sections to match the first section's styling
- Add accessible feedback display with opaque `bg-white` background for WCAG AA contrast compliance
- Support multiple question groups per page with per-group selection tracking, validation, and localStorage persistence
- Extract side effects (sound, email, attempt counter) into `recordAttemptResult` so they fire once per submit, not once per group

## Test plan
- [ ] Generate MCQ pages and verify correct/incorrect feedback displays properly
- [ ] Verify radio button selection doesn't cause layout shifts
- [ ] Generate multi-section MCQ pages and verify visual consistency across sections
- [ ] Test pages with multiple question groups — each group should track independently
- [ ] Verify keyboard shortcuts (1-9) work on single-group pages and are disabled on multi-group pages
- [ ] Check feedback text contrast on various background colors

### DIFF
--- a/assets/adt/modules/activities/multiple_choice.js
+++ b/assets/adt/modules/activities/multiple_choice.js
@@ -325,11 +325,6 @@ const selectOption = (option) => {
     groupOptions.forEach(opt => {
         opt.setAttribute('aria-checked', 'false');
         setOptionState(opt, 'default');
-        const feedback = opt.querySelector('.feedback-container');
-        if (feedback) {
-            feedback.classList.add('hidden');
-            feedback.classList.remove('bg-white', 'rounded-md', 'px-3', 'py-2', 'mt-3');
-        }
     });
 
     selectClickedOption(option);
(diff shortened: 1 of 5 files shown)


ksokolovic:
I reviewed the exact PR head (`596be90d`) against the current `main` branch. My final verdict is to close this PR without merging.

The main blockers are:

- `main` has deleted the vanilla-JavaScript activity runtime modified here and replaced it with the TypeScript runtime under `apps/adt-runtime`; resolving the current conflicts would risk restoring retired code.
- The new consistency LLM pass rewrites HTML after answer generation without revalidating the structure or regenerating the answers, so option IDs and stored answers can silently diverge.
- The MCQ-specific consistency prompt is applied to every repeated `activity_*` type.
- The advertised cross-page consistency is only within one `renderPage()` call, so it does not cross pages.
- The PR adds no regression tests for these paths.

I verified the PR head with Node 24: typecheck, build, the focused rendering suite, and the full historical test suite pass. Those checks do not cover the new browser-runtime or consistency-pass behavior.

The original `"false"` truthiness concern is still worth fixing, but it should be handled in a small new PR against the current `apps/adt-runtime` implementation, with a focused regression test.

Issue #215, opened 2026-03-30
[Pipeline/Export] Cannot delete/choose generated assets ie. TTS


elasticsounds:
At present, if you generate a pipeline stage, you cannot delete/reverse this generation. This means if you generate TTS, there is no way to go back forcing you to keep this in the final package. Add the ability to choose or delete various pipeline stages before packaging for export.

Options for exploration:
- Choose what to include in exports.
[PRO] You can customize your export without deleting assets depending on your target distribution scenario.
- Delete pipeline stages entirely.
[PRO] Simpler than the first route

Eliezir:
Added at #272

PR #272, opened 2026-04-10
Export scopes

Eliezir:
## Context
Previously the export stage had a single "ADT Export" that bundled everything the pipeline produced. Another problem was that once the user added a feature in the pipeline (TTS, quizzes, etc.), there was no way to remove them from the export. This PR splits exports into dedicated formats, lets the user toggle features per export, and introduces a project import flow so books can be moved between instances.

### Export types
Replaced the old single ADT export with four distinct formats, each with its own filename suffix so downloads don't collide:
- **Project Archive** (`<label>-project.zip`) — Full backup of the book directory (database, PDF, pipeline outputs). For transferring books between instances. This is what was previously called "ADT Export".
- **Web Export** (`<label>.webpub`) — The ADT web bundle (HTML pages, images, audio, quizzes, web app).
- **SCORM Package** (`<label>-scorm.zip`) — Web export wrapped as a SCORM 1.2 package.
- **ADT Package** (`<label>-adt.zip`) — ADT-flavored bundle.

Each format has its own card in the export view with a dedicated color, icon, and description.

### Export dialog with feature toggles
Clicking "Export" on any format opens a new two-panel dialog:

**Left sidebar** shows book metadata — cover image, title, authors, publisher, page count, and languages. Output languages are shown as a summary pill with a popover listing all translations.

**Right panel** changes based on the format:

For **Web/SCORM/WebPub** exports, it shows:
- **Feature toggles** for glossary, read-aloud (TTS), and quizzes — users can turn these on/off per export.
- **Accessibility checklist** — image descriptions, table of contents, and sign language are **not toggleable**. They're treated as accessibility requirements and always included when available. If any of them haven't been generated yet, the checklist shows a recommendation to run those pipeline steps, with WCAG reference codes explaining why they matter.

For **Project Archive**, it shows what's included in the archive (database, pipeline features count).

### Project import
An "Import project" button on the home page opens a dedicated `/books/import` route that:
- Accepts a `.zip` project archive via click or drag-and-drop (drop target is the whole viewport).
- Validates the archive has a `.db` and `.pdf` at the root (throws `Invalid project archive` otherwise).
- Shows a **preview card** before import — book title, authors, publisher, page count, cover, languages, and which pipeline stages are already done.
- Auto-renames the book on label conflict (`label` → `label-2`).
- Surfaces friendly errors for corrupt ZIPs vs. non-ADT archives.

## Test plan
- [ ] Export Web with all features on → open the exported `config.json` → verify `glossary: true`, `readAloud: true`, `activities: true`
- [ ] Export Web with all features off → verify they're `false`
- [ ] Open export dialog for a book missing accessibility features → verify checklist shows recommendations with WCAG codes
- [ ] Export each format → verify each downloads with its distinct filename suffix (`-project.zip`, `.webpub`, `-scorm.zip`, `-adt.zip`)
- [ ] Export Project Archive → import on a fresh instance → verify book loads
- [ ] Import a ZIP with a label that already exists → verify it auto-renames to `label-2`
- [ ] Import a ZIP missing the `.db` or `.pdf` → verify "Invalid project archive" error
- [ ] Import a corrupt/non-ZIP file → verify friendly "damaged archive" error

---
Addresses https://github.com/unicef/adt-studio/issues/215


elasticsounds:
Issue with languages in final package:
config.output_languages only stores translations, not the base language, and neither code path adds the base language back in.

Frontend — apps/studio/src/components/pipeline/stages/export/ExportDialog.tsx:573-579
const configLanguages = Array.from(
  new Set(
    ((config?.output_languages as string[] | undefined) ?? []).map((code) =>
      normalizeLocale(code),
    ),
  ),
);
This is the list that populates the modal’s LANGUAGES section. It omits the base language entirely, so with base=es and translations=[en,fr], the list becomes [en, fr]. The “First language is the default” label (line 451) then marks English as DEFAULT.

Note: the BookCover sub-component at line 73 does read book?.languageCode correctly — that’s why the “Spanish +2 translations” pill in the left panel is right, while the list on the right is wrong.

Backend — apps/api/src/services/export-service.ts:92-98
const outputLanguages = Array.from(
  new Set(
    (config.output_languages && config.output_languages.length > 0
      ? config.output_languages
      : [language]).map((code) => normalizeLocale(code))
  )
)
Same shape: the base language (Spanish) is only used as a fallback when output_languages is empty. Since the user had translations configured, Spanish is dropped from the final bundle — matching the observation that only English/French play in the exported package.

Correct pattern (already in the codebase)
apps/api/src/services/stage-runner.ts:1550 does it right:

const outputLanguages = Array.from(
  new Set(
    [language, ...(config.output_languages ?? [])].map((code) => normalizeLocale(code))
  )
)
Fix
Both sites should prepend the base language (book.languageCode / config.editing_language / metadata.language_code) to output_languages and dedupe, mirroring the stage-runner pattern. In the modal, the base language should also be the one marked DEFAULT (since it’s the book’s editing language), regardless of array order.

PR #275, opened 2026-04-12
Add empty-state screens for pruned books and storyboard-dependent stages

### LINKED ISSUE #273: Add proper empty-state screen for books with no pages (all pages pruned)
When all pages of a book are pruned, the Preview stage currently shows a raw JSON error: {"error":"ADT has no pages"}. There is no user-friendly feedback, and the Export stage is also broken in this state since there is nothing to export.

Expected behavior:

A clear empty-state screen should be shown when a book has no remaining pages after pruning
The screen should explain what happened and suggest next steps (e.g., go back to Extract and adjust the page pruning settings)
Export should either be disabled or show an equivalent message
Steps to reproduce:

Open a book where all pages have been pruned
Navigate to Preview or Export
I'll fix this — just filing so I don't forget.

Image

Eliezir:
Closes #273

## Summary

- Extracted reusable `StoryboardRequired` and `AllPagesPruned` empty-state components from inline implementations scattered across Preview, Export, Validation, and Sign Language views into a shared `StageEmptyStates.tsx` module
- Added user-friendly empty states for Preview and Export when all pages have been pruned, replacing the raw `{"error":"ADT has no pages"}` JSON error
- Extended the `StoryboardRequired` guard to all remaining pipeline stages that depend on a completed storyboard: **Captions, Glossary, Quizzes, ToC, Translate, and Speech**

## Future improvement

Once we have dedicated landing pages for each pipeline step, we can replace these empty-state screens with informative landing pages that explain what the step does. Instead of blocking the entire view, we would just disable the run button and show a message indicating the storyboard needs to be run first — giving users better context about each stage while still preventing premature execution.

## Test plan

- [ ] Open a book where **all pages have been pruned** — verify Preview and Export show the new empty-state screen instead of a JSON error
- [ ] Open a book that has **not run the storyboard** yet — navigate to each dependent stage (Captions, Glossary, Quizzes, ToC, Translate, Speech) and verify the "Storyboard required" empty state is shown
- [ ] Run the storyboard, then verify each stage loads its normal content
- [ ] Check all three locales (en, pt-BR, es) render translated empty-state messages

### DIFF
--- a/apps/studio/eslint-suppressions.json
+++ b/apps/studio/eslint-suppressions.json
@@ -44,11 +44,6 @@
       "count": 3
     }
   },
-  "src/components/pipeline/stages/PreviewView.tsx": {
-    "lingui/no-unlocalized-strings": {
-      "count": 7
-    }
-  },
   "src/components/pipeline/stages/storyboard/components/BookPreviewFrame.tsx": {
     "lingui/no-unlocalized-strings": {
       "count": 2
--- a/apps/studio/eslint.config.js
+++ b/apps/studio/eslint.config.js
@@ -131,10 +131,11 @@ export default [
             "*.info",
             "*.debug",
 
-            // --- Error constructors ---
+            // --- Error constructors & state error setters ---
             "Error",
             "TypeError",
             "RangeError",
+            "setError",
 
             // --- Module system ---
             "require",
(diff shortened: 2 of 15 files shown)


Eliezir:
Implemented at #421

PR #276, opened 2026-04-13
V1 Translation Evaluation

simba-mudz:
This PR adds a first V1 implementation of translation evaluation in ADT Studio using the project-native TypeScript API and existing `@adt/llm` client.

Based on review feedback, I removed the separate Python `eval-service` and MLflow dependency. Translation evaluation now runs inside the existing API process, and normalized evaluation results are stored as versioned `node_data` in each book’s SQLite database.

## What’s included

- translation evaluation config and schemas
- API routes for listing, running, and reading translation evaluations
- TypeScript evaluation runner using `@adt/llm`
- versioned local persistence for evaluation results in book SQLite storage
- Validation UI for running and reviewing evaluations
- progress display while evaluation tasks are running
- lightweight evaluation status in Text & Speech
- documentation for the current architecture and request flow

## Current design

The current V1 shape is:

1. Studio owns settings, run controls, and result review UI
2. API owns payload construction, task execution, judge calls, and persistence
3. `@adt/llm` handles cached/logged LLM judge calls
4. Each book SQLite database stores normalized evaluation results as versioned `translation-evaluation` node data
5. No separate Python service, MLflow server, or eval-specific Docker service is required

Evaluations are optional and non-blocking. They do not block packaging or replace manual reviewer validation.

## Notable updates from the earlier draft

- Removed the Python `eval-service`
- Removed MLflow and eval-service Docker Compose wiring
- Removed legacy MLflow metadata from the shared result schema
- Stored evaluation results directly in ADT Studio book SQLite storage
- Passed the OpenAI API key explicitly into `@adt/llm` for translation evaluation instead of mutating process-wide environment state
- Cleaned up docs, tests, locale strings, and stale service references

## Review focus

Main feedback I’m looking for:

- Does the TypeScript-native evaluation flow fit ADT Studio’s architecture?
- Is the `translation-evaluation` versioned storage shape appropriate?
- Are the UI states clear enough for current, stale, running, and failed evaluations?
- Does this feel like a reasonable foundation for future evaluation workflows?

## Key files

- `packages/types/src/translation-evaluation.ts`
- `apps/api/src/routes/translation-evaluations.ts`
- `apps/api/src/services/translation-evaluation-runner.ts`
- `apps/api/src/services/translation-evaluation-service.ts`
- `apps/studio/src/hooks/use-translation-evaluation.ts`
- `apps/studio/src/components/validation/TranslationEvaluationTab.tsx`
- `apps/studio/src/components/validation/TranslationEvaluationSettingsTab.tsx`
- `docs/TRANSLATION-EVALUATION.md`

## Validation

Ran focused API/UI/schema/runner tests, `pnpm build`, Studio lint, and `docker compose config`.
```


simba-mudz:
Update: based on the feedback so far, I removed the separate Python eval-service and MLflow dependency from this PR.

Translation evaluation now runs inside the existing TypeScript API using `@adt/llm`, and results are stored as versioned `translation-evaluation` node data in each book’s SQLite database.

I updated the PR description to reflect the current architecture.

nicpottier:
I think the architecture looks good!

 * I ran this on momo and it found a few issues that I agreed with, which I think is encouraging that even for a simple book there are things that can be improved with this kind of method.
 * Running evaluation took quite a bit, it looks like evaluation is done one string at a time as opposed to doing them in parallel with a set number of workers. Doing the latter should dramatically speed things up. It would probably be worth seeing whether batching many together in the same request/response would still preserve accuracy as that would also dramatically improve things. I fear right now it might be too slow for users to opt into despite there being benefits to it.
 * the UI could use a pass to simplify things and make things more obvious, these are all pretty simple and you are probably aware of them but for completeness:
   - it seems a bit odd that there is both a gate in settings to "enable" this but then also you have to manually run it for a language. It should be one or the other. Probably just being able to run it manually for a language is good enough for now without the additional settings gate.
   - the settings panel has some expanding sections. I don't think we do this anywhere else. In general I would look to simplify this a bit.
   - the actual results have a lot of repetition in that we put the rationale twice etc.. we should also probably default sort things that need attention.
   - it would be really nice to formalize the suggested "fix" and let the user accept it with a single click. The corrections are clearly calls for action but there's no easy way to take that action.
   - exactly where this lives and how it looks is probably still a work in progress. Maybe evaluation needs a landing page that bubbles up all these different types of evaluation and summarizes their results / allows you to run them? The small tab feels a bit hidden away at the moment.
   
Otherwise this all looks great to me and I think sets a good pattern that we could likely use elsewhere. The next obvious question is do we (and how do we) expose this on the translation tab itself.

@elasticsounds

PR #285, opened 2026-04-16
Fixed-layout EPUB rendering for illustrated storybooks

lest:
## Summary

Adds a fixed-layout rendering mode for illustrated PDFs (children's storybooks, picture books) where the original page composition matters: images placed by PDF coordinates, text overlaid as positioned spans for accessibility / search / TTS, blocks clustered at the speech-bubble level for translation. Ships alongside an EPUB 3 export format with **EPUB 3 Media Overlays (SMIL)** for word-level read-aloud sync.

### Fixed-layout pipeline mode
- New `fixed_layout` render strategy (first-class, alongside `llm`, `single_column`, `two_column_story`, etc.); selectable under the Storybook and Custom presets.
- Single-pass StructuredText walk extracts paragraphs with per-glyph font / size / colour data. Spread-aware (left + right pages stitched to a unified viewport).
- Block clustering groups speech-bubble lines into one logical paragraph; sentence boundaries split for TTS / translation chunking.
- Renderer pins each block to PDF coordinates with line-height derived from max font-size; per-segment contrast strokes against sampled background; `data-segments` JSON carries styling for runtime rebuild on language swap and inline edits.
- Shared `assets/adt/auto-fit.js` shrinks letter-spacing then font-size to fit text into bubble bounds; reschedules on font-swap events (`document.fonts.ready` + `FontFaceSet.loadingdone`).

### Word-level spans + viewer highlighting
- Renderer wraps each whitespace-separated token in `<span id="{paragraphId}_w{NNN}">` inside the existing per-segment style span. Whitespace stays outside the span so layout is preserved.
- `data-id` (paragraph) vs `id` (word) split: paragraph `data-id` continues to drive text-catalog / TTS / language swap; word `id` is consumed only by SMIL fragment refs and the viewer's word highlighter.
- Runtime translation rebuild (`translations.js`) preserves word spans through both source-language and translation paths, so DevTools shows them after preview render.
- Viewer (`tts_highlighter.js`) detects fixed-layout word spans via id pattern and switches between word-level and block-level highlight; binary count-guard falls back to block highlight when source span count diverges from Whisper word count.

### EPUB 3 export + Media Overlays (SMIL)
- New `packageEpub` built from the same ADT web output as WebPub.
- OPF emits `rendition:layout pre-paginated` for fixed-layout books.
- New `packages/pipeline/src/smil.ts` builder: per-page SMIL document with one `<seq>` per paragraph, one `<par>` per word when source span count exactly matches Whisper word count, paragraph-level `<par>` fallback otherwise. Explicit `clipBegin` / `clipEnd` on every `<par>` so EPUBCheck and readers don't have to infer audio duration.
- Manifest `<item media-overlay="..."/>` (per EPUB 3 spec — not on spine `<itemref>`); `media:duration` per overlay + total; `media:active-class` set to `-epub-media-overlay-active` with a CSS rule in the page stylesheet.
- Gated on the **Read-aloud / Speech** export toggle (`features.readAloud`), independent of fixed-layout — reflowable EPUBs also get SMIL when audio is present.
- Paragraph parsing uses jsdom to walk any `[data-id]` element except ``, so reflowable, fixed-layout, and mixed pages all flow through the same generator.
- Studio UI: EPUB card in the Export view; works in EPUB readers that support JavaScript (interactive features carry over).

### Wizard
- `fixed_layout` is selectable as a render strategy under the Storybook and Custom presets — no separate preset.
- Step 3 (content processing) shows a "no content processing needed" message when `fixed_layout` is selected, since activities detection, figure extraction, and image cropping/segmentation don't apply to that pipeline.
- New `FixedLayoutPreview` mockup tile in `LayoutPreview`.

### Studio preview
- `PreviewView` detects fixed-layout pages via the iframe `<body>`'s inline width/height and CSS-scales the iframe to fit the preview pane — mirrors the storyboard behaviour.

### Spread image stitching
- Horizontal stitching of left/right page images (PNG + JPEG); image filtering prunes original halves when a stitched spread exists.

## Storage

Schema v12 → v13: adds `bounds_x/y/w/h` columns to `images` for per-image PDF placement. Migration is additive (existing rows get NULL bounds).

## Demo: fixed layout EPUB read aloud in BookFusion reader
https://github.com/user-attachments/assets/e8dc4d55-8cf1-45ae-b124-22bf24e01c1b

### DIFF
--- a/apps/adt-runtime/src/features/audio/lib/word-highlight.ts
+++ b/apps/adt-runtime/src/features/audio/lib/word-highlight.ts
@@ -23,19 +23,89 @@ import {
   createApproximateWordTimestamps,
   type WordTimestamp,
 } from "@/features/audio/lib/tokenizer"
+import { parseSegments, styleToInline, type Segment } from "@/shared/lib/fl-segments"
 
 const ORIGINAL_HTML_ATTR = "data-tts-original-html"
 const WORD_HIGHLIGHT_CLASS = "bg-yellow-300"
 const BLOCK_HIGHLIGHT_CLASS = "tts-active-block"
 
 /**
+ * Fixed-layout word wrap: tokenise across the concatenated `data-segments`
+ * text, then for each word emit a `<span data-word-index="N">` containing
+ * the per-segment styled `<span style="…">` slices that overlap the word's
+ * character range. Preserves the styled-run structure (coloured letters,
+ * mixed fonts, contrast strokes) the renderer baked in.
+ */
+function wrapWordsFromSegments(element: HTMLElement, segments: Segment[]): boolean {
+  const plan = buildWordRenderPlan(segments.map((s) => s.text).join(""))
+  if (!plan.some((s) => s.type === "word")) return false
+
+  const doc = element.ownerDocument
+  const segOffsets: number[] = []
+  let cursor = 0
+  for (const s of segments) {
+    segOffsets.push(cursor)
+    cursor += s.text.length
+  }
+
+  const buildPieces = (start: number, end: number): HTMLSpanElement[] => {
+    const out: HTMLSpanElement[] = []
+    for (let i = 0; i < segments.length; i++) {
+      const segStart = segOffsets[i]
+      const segEnd = segStart + segments[i].text.length
+      if (segEnd <= start) continue
+      if (segStart >= end) break
+      const slice = segments[i].text.slice(
+        Math.max(start, segStart) - segStart,
+        Math.min(end, segEnd) - segStart,
+      )
+      if (!slice.length) continue
+      const span = doc.createElement("span")
+      const styleStr = styleToInline(segments[i].style)
+      if (styleStr) span.setAttribute("style", styleStr)
+      span.appendChild(doc.createTextNode(slice))
+      out.push(span)
+    }
+    return out
+  }
+
+  const fragment = doc.createDocumentFragment()
+  let charCursor = 0
+  for (const seg of plan) {
+    if (seg.type === "separator") {
+      if (seg.text.length > 0) fragment.appendChild(doc.createTextNode(seg.text))
+      charCursor += seg.text.length
+      continue
+    }
+    const start = charCursor
+    const end = start + seg.text.length
+    const wrap = doc.createElement("span")
+    wrap.setAttribute("data-word-index", String(seg.wordIndex))
+    for (const piece of buildPieces(start, end)) wrap.appendChild(piece)
+    fragment.appendChild(wrap)
+    charCursor = end
+  }
+  element.replaceChildren(fragment)
+  return true
+}
+
+/**
  * Wrap each whitespace-separated word inside `element` with a span carrying
  * `data-word-index="N"`. Idempotent — calling twice on the same element
- * is a no-op.
+ * is a no-op. When the element carries `data-segments` (fixed-layout
+ * paragraphs with per-run styling), the styled-span structure is preserved
+ * inside each word wrap.
  */
 export function wrapWordsForElement(element: HTMLElement, text: string): void {
   if (element.hasAttribute(ORIGINAL_HTML_ATTR)) return
   element.setAttribute(ORIGINAL_HTML_ATTR, element.innerHTML)
+
+  const segments = parseSegments(element.getAttribute("data-segments"))
+  if (segments && segments.length > 0) {
+    if (wrapWordsFromSegments(element, segments)) return
+    element.removeAttribute(ORIGINAL_HTML_ATTR)
+    return
+  }
 
   const plan = buildWordRenderPlan(text)
   // No matchable words (pure punctuation, empty string) — leave the element
--- a/apps/adt-runtime/src/features/language/runtime/i18n.ts
+++ b/apps/adt-runtime/src/features/language/runtime/i18n.ts
@@ -14,6 +14,7 @@ import {
   translationsAtom,
   videoFilesAtom,
 } from "@/features/language/state/language.atoms"
+import { rebuildSegmentedInnerHtml } from "@/shared/lib/fl-segments"
 
 async function safeJsonFetch<T = unknown>(
   url: string,
@@ -147,15 +148,20 @@ export function applyTranslationsToDOM(
     elements.forEach((el) => {
       if (el.tagName === "IMG") {
         el.setAttribute("alt", text)
-      } else {
-        if ((el as HTMLElement).hasAttribute("data-tts-original-html")) {
-          ;(el as HTMLElement).setAttribute(
-            "data-tts-original-html",
-            renderedHtml,
-          )
-        }
-        ;(el as HTMLElement).innerHTML = renderedHtml
+        return
       }
+      // Fixed-layout paragraphs carry per-run styling on `data-segments`.
+      // Rebuild the styled-span tree from that JSON so font-family, color,
+      // size, weight, and stroke survive the text swap — straight innerHTML
+      // assignment would flatten them.
+      const segmentsAttr = (el as HTMLElement).getAttribute("data-segments")
+      const html = segmentsAttr
+        ? rebuildSegmentedInnerHtml(segmentsAttr, renderedHtml)
+        : renderedHtml
+      if ((el as HTMLElement).hasAttribute("data-tts-original-html")) {
+        ;(el as HTMLElement).setAttribute("data-tts-original-html", html)
+      }
+      ;(el as HTMLElement).innerHTML = html
     })
 
     const placeholders = document.querySelectorAll(
@@ -173,6 +179,17 @@ export function applyTranslationsToDOM(
   if (titleMeta) {
     const id = titleMeta.getAttribute("content")
     if (id && translations[id]) document.title = translations[id]
+  }
+
+  // Re-run fixed-layout auto-fit after rebuilding segment spans — the new
+  // elements have no `data-adt-fs` cache and the translated content has a
+  // different length than the source, so an earlier auto-fit pass against
+  // the original spans no longer reflects the rendered DOM. Double-rAF so
+  // we measure after the browser has laid out the new innerHTML.
+  const runAutoFit = (window as Window & { __adtRunAutoFit?: () => void })
+    .__adtRunAutoFit
+  if (typeof runAutoFit === "function") {
+    requestAnimationFrame(() => requestAnimationFrame(() => runAutoFit()))
   }
 }
 
--- a/apps/adt-runtime/src/shared/lib/fl-segments.ts
+++ b/apps/adt-runtime/src/shared/lib/fl-segments.ts
@@ -1,0 +1,90 @@
+/**
+ * Fixed-layout paragraph segment helpers.
+ *
+ * FL `[data-id]` paragraphs carry per-run styling on `data-segments` — JSON
+ * mirroring `SectionTextSegment` in `@adt/types`. Two consumers in this
+ * runtime walk that JSON:
+ *
+ *   - `features/audio/lib/word-highlight.ts` rebuilds nested styled spans
+ *     inside each word wrapper during read-aloud, so word highlighting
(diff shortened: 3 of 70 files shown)


elasticsounds:
This is really great - being able to see the text exact positioned on the images works really well.
I created a PR to address the storyboard not displaying the text - it was an issue with the storyboard injecting css to support the image editing popup. I removed that rule and put in a fix. 

Other things to consider
 - The full height images for fixed layout are appearing larger than the vertical height of storyboard - we should address a max height for these kind of books.
 - Styling of text within a <span> with a data-id will be replaced by the text in the JSON by the translation system in the reader. Need support for styled elements that does not break TTS.
  - The text types are not being assigned to these exact positioned text elements. Consider the addition of a new text type for absolutely positioned text, and a group type that creates the full page overlay.
  - TTS word level timecodes does not flow to final output - need word level highlighter and support for media-overlay.
  - Re-running storyboard with a different render strategy does not work on the book. Only two column template seems to work is this hardwired?

elasticsounds:
https://github.com/unicef/adt-studio/tree/elasticsounds/fix-fixed-layout-click-routing

elasticsounds:
One other comment - the exact positioned elements in this mode, should exact positioning of all elements just be added to the datastructure for all objects in the system. I'm very curious how it will work on more complex books.

elasticsounds:
[LP-4-Volcanoes.pdf](https://github.com/user-attachments/files/26882229/LP-4-Volcanoes.pdf)
This book is a great example of a more challenge fixed layout with exact positioned elements

elasticsounds:
@lest are we going to expose the exact position metadata to the user?

elasticsounds:
How does image meaningfulness work with the fixed layout type - should this be disabled for fixed layout layout strategy?

elasticsounds:
Mid-word style breaks produce multiple word-ids per spoken word. In packages/pipeline/src/fixed-layout-rendering.ts, renderSegmentsToHtml tokenises each data-segments entry on whitespace in isolation, so any word that’s split across style runs gets one <span id="…_wNNN"> per fragment instead of one per spoken word. For example, “lahars” with a stylised white “h” on page 14-15 of volcanoes_single renders as three ids (_w011=la, _w012=h, _w013=ars). This silently degrades EPUB 3 Media Overlays — Whisper produces one timestamp for “lahars” so the count guard in tts_highlighter.js falls back to block-level highlighting for the whole paragraph (the commit message for ee752ded already flags this), breaks word-by-word navigation and in-reader search across the affected words, and can introduce audible micro-pauses on some EPUB reading systems (notably Android TalkBack) that don’t fully flatten sibling style spans. The fix has to be at render time because SMIL fragment ids must resolve against the static XHTML, and the only EPUB-valid shape is Option A — one word span wrapping the nested style spans — since reusing the same id across fragments would fail epubcheck’s duplicate-id rule. The information needed to detect mid-word splits is already present in the data model: whitespace lives between segments, so a segment that doesn’t start with \s following a previous segment that didn’t end with \s is, by construction, a continuation of the same word. The same change needs to mirror in assets/adt/modules/translations.js, which re-emits word spans from data-segments on language swap.

elasticsounds:
Multi-segment styling is lost on every translation. A fixed-layout paragraph carries its source-language styling on the data-segments JSON attribute, where each segment records both its text and its style (font-family, size, colour, stroke). At runtime, rebuildSegmentedInnerHtml in translations.js has two branches: when the swapped text matches the source-language concatenation it emits one styled <span> per segment and styling is preserved, but when the text is a translation it collapses every segment into a single span carrying only segments[0].style. The result is that any stylistic differentiation inside a paragraph — a coloured letter mid-word, a contrast-stroked emphasis, a font shift between a heading and its tail — survives only in the source language and is lost in every translation. The cleanest fix is to make the translation catalog segment-aware so each segment is translated independently and re-emitted with its own style; an alternative is to move styling fully out of the rendered HTML into the JSON layer and rely on translations.js to apply all styling at runtime, treating the rendered XHTML as a structural shell rather than a styled artefact. Either approach is fine architecturally — the constraint is that translated text must be able to inherit per-run styling without the renderer or the catalog losing the segment boundaries.

Static SMIL ties word ids to one language, so each EPUB must commit to a single export language. The SMIL document referenced by EPUB 3 media-overlays resolves <text src="…#_wNNN"/> against fragment ids that exist in the static XHTML; there is no JavaScript runtime to swap content in a downloaded EPUB consumed by Thorium or Apple Books. Word ids are produced at render time in the source language and the SMIL audio cues are aligned to that XHTML, so attempting to ship multi-language read-aloud in one EPUB would either misalign every word or require re-rendering the XHTML per language inside one package, which the spec doesn’t really support. The simplest and correct resolution is to treat fixed-layout EPUB export as single-language by construction: the export UI must force the user to pick exactly one target language before the export runs, the renderer re-renders the page XHTML in that language with language-local word ids, and the SMIL is generated against that XHTML using audio in the chosen language. The studio’s current export flow likely needs a language picker added to the fixed-layout EPUB export modal, with no “all languages” option.

Word counts diverge across languages and the count-guard fallback hides the problem at runtime but not in the exported EPUB. Whisper produces one timestamp per spoken word in the target audio, while the rendered spans tokenise on whitespace in the target text — counts only line up coincidentally. A German “Betonflüsse.” renders as one span where an English “rivers of concrete.” gives three; whitespace-free scripts (Chinese, Japanese, Thai) render the entire paragraph as a single span. The in-app web viewer handles this gracefully via the count guard in tts_highlighter.js:191-198, which falls back to a paragraph-level block highlight when flSpanCount !== wordTimestamps.length. The exported SMIL EPUB has no such fallback — there is no JavaScript and no guard, only the static SMIL document, so a count mismatch presents as visible word-by-word misalignment. Once the single-language export constraint is in place (issue #2), this should be solved at export time rather than left to runtime: regenerate the audio and the word timestamps in the chosen target language, validate that the resulting word-span count matches per paragraph, and either accept block-level highlighting for paragraphs where they don’t (omit those paragraphs from the SMIL) or surface a warning in the export UI listing the affected paragraphs.

The mid-word styling split issue from the previous review compounds across every language. When a single source word like “lahars” is broken into multiple data-segments entries because of mid-word styling and the renderer emits one word id per fragment instead of one per spoken word, that same problem now recurs in every translated language because renderInnerWithWordSpans in translations.js mirrors the renderer’s tokenisation rule. A translation that happens to render the equivalent stylised word — or even a plain word in a paragraph where the source had stylisation — inherits the same fragmentation pattern, so the word-count mismatch hits every language, not just the source. Worse, the translated-text branch collapses segments anyway (issue #1), so the original stylistic motivation for splitting the word is also lost — you end up with three word ids for no visual reason. Fixing the source-side splitting issue is therefore a prerequisite for any of the other three: unifying mid-word fragments into a single word span (with nested style spans for the styled portions) makes word counts stable across segment shapes, makes translation rebuild trivially preserve word identity, and lets the SMIL alignment have a chance of matching Whisper’s tokenisation in the export language.

elasticsounds:
Embedded PDF fonts don’t survive to the EPUB and translated text has no matching font at all. The PDF embeds the original publisher’s fonts (often commercial faces like MuseoSans, Chokle, etc.) but the current pipeline doesn’t extract them, so the rendered XHTML lists those names in font-family while the actual files aren’t bundled — readers fall back to whatever the system has, which is essentially never the intended font. Even if we did extract the embedded fonts via mupdf (issue raised earlier), most PDFs ship subset fonts containing only the glyphs used on the original pages; that’s enough for the source language but breaks the moment a translation introduces accented or non-Latin characters the subset doesn’t cover. The right approach for fixed-layout multi-language EPUBs is to map each detected PDF font to a visually close Google Font with broad language coverage, bundle the matched faces into the EPUB’s OEBPS/fonts/ directory, and emit @font-face declarations so the typography is preserved across languages instead of collapsing to a system fallback. The mapping can be a curated table (Chokle → Fredoka, MuseoSans → Nunito Sans, etc.) seeded by category — sans-serif rounded, slab, display — possibly with a small UI to let the user override matches in the export modal. Google Fonts is freely redistributable under the OFL/Apache licences, which sidesteps the licensing question that blocks shipping the original embedded fonts, and it gives users a clear, predictable typography pipeline rather than a font lottery dependent on the reader’s system.

elasticsounds:
Test results — epub-fixed-layout branch review

Priority — must be fixed before merge
1) Outline-only rendering applied to text that doesn’t need it. The contrast-stroke logic in fixed-layout-rendering.ts:135-152 (computeStrokeForColor) is firing on black text against white backgrounds, producing white-outline stylisation where the original PDF was just a plain or larger hand-drawn font. The heuristic is purely WCAG contrast-based and has no concept of the source design intent, so it’s randomly converting legible black-on-white text into outlined glyphs. This is a visible regression and needs to be fixed — either by tightening the trigger condition (only apply stroke when the sampled background is not near-white and the text colour isn’t already near-black), or by adding an opt-out attribute we can flag per segment.

2) Storyboard panel doesn’t fill horizontally for fixed-layout pages — boxed view. BookPreviewFrame.tsx:731-734 computes the iframe scale as Math.min(1, availableWidth / fixedLayoutSize.width). The Math.min(1, …) cap prevents the iframe from ever upscaling, so when a PDF’s natural page width is smaller than the panel the page renders at 1× with whitespace on either side. Removing the Math.min cap (or replacing it with a sensible upper bound like 2×) would let fixed-layout pages fill the storyboard at the same density as reflowable content. This is the same panel users see throughout the pipeline, so the boxed appearance is a notable UX downgrade for fixed-layout books.

3) Speech-bubble images are still being clipped during extraction. extract.ts:753-801 (selectImageClipPath) picks the most-restrictive active clip path and embeds it strictly. When the PDF’s clip rect doesn’t include the bubble’s soft shadow or outer stroke, those borders are cropped off in the extracted asset. The fix is to inflate the clip bounds slightly when the underlying image extends past them, or to detect speech-bubble–shaped clip paths and use the image’s native bounds instead. This was flagged in the previous review and isn’t addressed in the recent commits.

Good progress — but one open question
4) Word-by-word styling now survives language switching — great progress on localisation. Commit 839ef45c (“defer word-span wrapping to packaging and live viewer”) meaningfully improves this: the renderer now emits a structural shell of one <span style> per segment with no per-word ids, and word spans are materialised downstream at EPUB packaging time (package-epub.ts:wrapWordSpans) or at audio-play time in the viewer (tts_highlighter.js:wrapTextInSpans). The packaging path uses positional overlap with style runs and explicitly keeps cross-run words as a single id, which resolves the “lahars → three word-ids” mid-word fragmentation we flagged previously.

The remaining gap is multi-segment styling on translated text: rebuildSegmentedInnerHtml in assets/adt/modules/translations.js still has the branch where, if the swapped text doesn’t match the source-language concatenation, all segments collapse into segments[0].style. So a paragraph with a coloured letter or mixed fonts in the source renders correctly in the source language but loses that styling in every translation. Can this be tightened in this branch, or are you planning to address it in a follow-up after this one merges? Either is fine — just want to know the intent so we can plan around it.

Out of scope — defer to a follow-up
5) CMYK→RGB conversion on extracted images. extract.ts:1276-1288 already detects CMYK JPEGs (DeviceCMYK and ICCBased N=4) and converts via mupdf’s built-in conversion, but the result is colour-flat for print-origin PDFs with embedded ICC profiles, because the conversion is naive and doesn’t honour the profile. Proper handling needs ICC profile parsing and rendering-intent support, which is a non-trivial change to the extraction pipeline. Agreed to defer to a later iteration.

6) Sans-serif font fallback / closest-match font support. Only Merriweather (a serif face) ships in assets/adt/fonts/, and withBundledFallback() appends it to every font-family chain regardless of the original PDF font’s category. So PDFs using sans-serif faces (MuseoSans, Avenir, etc.) fall back to Merriweather instead of a sans-serif equivalent. Out of scope for this PR — moving to a separate change that adds a font-mapping table (likely Google Fonts) so each detected PDF font is matched to a visually close, redistributable face.

elasticsounds:
Thanks Sergey will review

PR #286, opened 2026-04-16
Tree-based sectioning, structured content editor, and section thumbnails

elasticsounds:
## Summary

- Extend `page-sectioning` to consume the recursive content tree from `page-structuring` via a new `SectionContentNodePart` variant; stage-runner detects tree-vs-flat data and routes accordingly, and `web-rendering`'s `expandParts` flattens subtrees.
- Replace the extract stage's legacy detail view with the shared `ContentNodeBlock` tree UI (drag/drop restructure, inline text editing, role/structure pills) and unify storyboard overview + per-section editing into one all-pages view.
- Capture fixed-viewport PNG section thumbnails during every render/re-render using a shared Playwright `ScreenshotRenderer`; serve via new `GET /books/:label/thumbnails/:filename` and swap the scaled-iframe previews in `SectioningOverview` and `ContentEditor` for `` tags.
- Misc: storage gains `thumbnails/` dir + helpers, new Lingui strings extracted and translated to es / pt-BR, prompt updates for tree node summaries.

## Test plan

- [ ] `pnpm typecheck` passes
- [ ] Run storyboard on a book with tree-based structuring and verify sections render with `content_node` parts
- [ ] Verify thumbnails appear in the overview and content editor after render, and refresh after re-render
- [ ] Drag/drop content across sections, save, and confirm re-render reflects the new structure
- [ ] Confirm older flat text-classification books still section + render (backward compat)

### DIFF
--- a/apps/api/src/routes/books.ts
+++ b/apps/api/src/routes/books.ts
@@ -3,7 +3,7 @@ import path from "node:path"
 import { Hono } from "hono"
 import { HTTPException } from "hono/http-exception"
 import { parseBookLabel, PIPELINE } from "@adt/types"
-import { openBookDb } from "@adt/storage"
+import { openBookDb, resolveBookPaths } from "@adt/storage"
 import {
   listBooks,
   getBook,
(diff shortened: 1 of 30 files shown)


nicpottier:
Replaced by #313

PR #295, opened 2026-04-18
Fix dev server port conflicts across worktrees

### LINKED ISSUE #290: Dev server port conflicts across multiple worktrees (Conductor) — studios proxy to the wrong API
## Summary

When running multiple branches in parallel via Conductor worktrees, all studio instances proxy `/api/*` to a single API process, which silently serves stale code from whichever branch booted first. Requests for endpoints that don't exist on that stale branch return `404` in what looks like the current branch's UI — making it easy to misread as an application bug.

**Status:** still investigating. Filing this so the behavior is tracked while I narrow down the exact failure surface and the cleanest fix.

## Observed behavior

Running three worktrees at once:

- API binds to `:3001` (only the first worktree succeeds)
- Studio A binds to `:5173`
- Studio B falls back to `:5174`
- Studio C falls back to `:5175`
- All three studios still proxy `/api/*` → `http://localhost:3001` (hardcoded in `apps/studio/vite.config.ts`)

Result: Studio B and Studio C talk to whichever API started first, typically an older branch missing newly-added routes.

## Reproduction

1. Check out three different branches in three Conductor worktrees
2. Run `pnpm dev` in each
3. In the newest branch's studio window, hit a route that only exists on that branch (e.g. `POST /api/books/preview-import`)
4. Observe `404` from the stale API

Evidence on my machine:
```
lsof -iTCP -sTCP:LISTEN -P | grep -E ":(3001|5173|5174)"
node  32511 ... TCP *:3001 (LISTEN)            # stale API
node  52431 ... TCP localhost:5173 (LISTEN)
node  98098 ... TCP localhost:5174 (LISTEN)    # my current worktree

curl -X POST http://localhost:3001/api/books/preview-import → 404
curl http://localhost:3001/api/health                       → 200
```

## Likely fix (pending confirmation)

Make both API port and studio proxy target configurable so each worktree can claim its own pair:

- `apps/studio/vite.config.ts` — read proxy target from `process.env.VITE_API_PROXY`, fall back to `http://localhost:3001`
- API already reads `PORT` env var
- Per-worktree `.env.local` (gitignored) assigns `PORT` + `VITE_API_PROXY` pair
- Document the pattern in `docs/DEVELOPER-GUIDE.md` under the Conductor workflow section

Still investigating whether there's a better solution (e.g. auto-assigning ports, or a Conductor-level config) before committing to this approach.

## Impact

- Medium developer-experience issue, no production impact
- Confusing because it surfaces as a seemingly-real 404 on the correct URL
- Affects anyone using Conductor/worktrees or running multiple checkouts in parallel

Eliezir:
Closes #290.

## Summary

Fixes dev server port conflicts when running multiple worktrees in parallel (Conductor or plain `git worktree`). Before: only the first worktree's API binds `:3001`; other studios silently proxy to the stale first instance, so endpoints that only exist on the current branch return 404 in what looks like a UI bug. After: each worktree picks its own free API port, and its studio proxy follows.

## How

- `scripts/dev.mjs` — root dev orchestrator. Finds a free port (sequential scan from `3001`), then spawns the existing `pnpm --parallel` dev with `PORT` and `API_PROXY_TARGET` threaded in as env vars.
- `scripts/find-free-port.mjs` — pure-Node port probe. Logic adapted from Zig's `findFreePort` in the `zig/electron` branch (`apps/electron/src/main/api/port.ts`).
- `apps/studio/vite.config.ts` — proxy target now reads `API_PROXY_TARGET` with `http://localhost:3001` fallback.
- `package.json` — root `dev` script now calls `node scripts/dev.mjs`. `dev:api` / `dev:studio` unchanged.

## Approach

We went with a monorepo-level orchestrator because it's self-contained, tool-agnostic (works with plain `git worktree`, Conductor, multiple terminals), needs zero per-worktree config, and adds no dependencies. A more robust alternative would be a dedicated tool like [Portless](https://community.vercel.com/t/using-portless-with-conductor-git-worktrees/34557), which handles automatic subdomain routing with zero env wiring — at the cost of another dev dependency and per-developer setup. 

## Scope

Web dev only. The electron launch path already picks its own port via its own supervisor logic and is untouched. Production builds are also untouched: the api's `PORT` fallback stays `3001`, and `vite build` ignores `server.proxy`.

## Test plan

- [ ] Run \`pnpm dev\` in one worktree → \`[dev] API port: 3001\`, studio on 5173
- [ ] Run \`pnpm dev\` in a second worktree concurrently → \`[dev] API port: 3002\`, studio on 5174, \`/api/*\` requests from the second studio hit the second worktree's API
- [ ] Temporary \`console.log\` on an api route in worktree 2, triggered from worktree 2's studio, appears in worktree 2's terminal (not worktree 1's)
- [ ] Solo \`pnpm dev:studio\` still proxies to \`:3001\` (backwards compatible)
- [ ] \`pnpm typecheck\` passes


Eliezir:
Hey @elasticsounds, I also put together another change that labels the browser tab with the Conductor workspace name, so it’s easier to tell tabs apart when working with multiple branches in parallel:

image

It’s dev-only and falls back to the current behavior for anyone not using Conductor. Worth shipping, or skip it? Let me know what you think.

PR #313, opened 2026-04-21
Add tree-native section editor and sectioning-stage Save

nicpottier:
## Summary
- Adds a reusable `SectionTreeEditor` that edits the canonical `ContentNodeData` tree directly — leaf text, role, container structure, prune toggle, duplicate, delete, nest, unnest, and drag/drop reorder.
- Mounts the editor on the sectioning-stage page with per-section pending state and a header Save/Discard; writes the full tree via a new `PUT /sectioning` endpoint, so container hierarchy survives the save round-trip.
- Splits the old flat-parts endpoint off as `PUT /sectioning-legacy` and redirects the storyboard's 7 existing save/back-propagate callsites to it, keeping the storyboard panel working while the tree-native migration continues.
- Tree ops (`findNode`, `editLeafText`, `moveNode`, `nestNode`, `unnestNode`, …) live in `@adt/types` so studio can depend on them without violating the layering rule. 24 unit tests cover round-trips and reference-equality.
- i18n: 12 new strings translated to es/fr/pt-BR (0 missing).

## Test plan
- [ ] Sectioning stage: edit a leaf's text, toggle a container's prune, nest a leaf into a new container, save; reload and confirm the container hierarchy survives (not flattened).
- [ ] Storyboard stage: confirm existing flat-parts editing still works via `/sectioning-legacy` (prune, crop, replace image, delete image, back-propagate text edits).
- [ ] `pnpm typecheck`, `pnpm --filter @adt/studio lint`, and the `section-tree-ops` unit tests pass.
- [ ] `pnpm --filter @adt/studio extract` reports 0 missing strings across all locales.

## Follow-ups
- Migrate `StoryboardSectionDetail.tsx` to mount `SectionTreeEditor` and drop the flat-parts handlers.
- Delete `SectionDataPanel.tsx`, `PageDetail.sectioning` (flat), `treeToParts*` shim, and the unused zod schemas once the storyboard is migrated.


nicpottier:
@elasticsounds Take over on the editor on this. Right now it only lives on the section pages but once complete we should be able to wire that same editor into the storyboard side panel. (and hook in live edits in either direction)

The goal is to eventually remove all calls to sectioning-legacy. Other parts that use that are the overview section page that shows all sections inline etc..

apps/api/src/routes/pages.ts
```diff
@@ -622,7 +594,9 @@ export function createPageRoutes(
         throw new HTTPException(404, { message: `Page not found: ${pageId}` })
       }
 
-      const version = storage.putNodeData("text-classification", pageId, parsed.data)
+      const version = storage.putNodeData("page-sectioning", pageId, parsed.data)
```
nicpottier, line 660: This save path now writes `page-sectioning` directly and only clears caption/catalog/TTS state. It leaves the existing `web-rendering` node intact, so any tree edit saved from the new sectioning UI or overview (`SectioningPageDetail` / `SectioningOverview`) can immediately diverge from the HTML we still serve as the current storyboard/ADT output. The code comment says the change cascades downstream, but it no longer invalidates storyboard output.

Refs:
- https://github.com/unicef/adt-studio/blob/792f2ac5828ca68c455aae3d494e503c387b8850/apps/studio/src/components/pipeline/stages/sectioning/SectioningPageDetail.tsx#L101-L118
- https://github.com/unicef/adt-studio/blob/792f2ac5828ca68c455aae3d494e503c387b8850/apps/studio/src/components/pipeline/stages/storyboard/components/SectioningOverview.tsx#L126-L138

apps/api/src/routes/pages.ts
```diff
@@ -2075,5 +2021,417 @@ export function createPageRoutes(
...
+      if (!sectioningParsed.success) {
+        throw new HTTPException(400, { message: "Invalid page-sectioning data" })
+      }
+      const sectioning = sectioningParsed.data
+
+      if (idx >= sectioning.sections.length) {
+        throw new HTTPException(400, { message: `Section index ${idx} out of range (page has ${sectioning.sections.length} sections)` })
+      }
+
+      const clonedSection = structuredClone(sectioning.sections[idx])
```
nicpottier: `cloneSection` duplicates the whole tree with `structuredClone`, so every cloned leaf/container keeps the same `nodeId`. That breaks the new tree-native editor because rendering uses `nodeId` as `data-id`, and the storyboard/editor code targets nodes with `querySelector([data-id=...])`, which will hit the first matching element once two sections share IDs. You already added `cloneNodeWithNewIds` for this exact invariant in `section-tree-ops`, but the route is bypassing it here.

Refs:
- https://github.com/unicef/adt-studio/blob/792f2ac5828ca68c455aae3d494e503c387b8850/templates/_render_node.liquid#L18-L30
- https://github.com/unicef/adt-studio/blob/792f2ac5828ca68c455aae3d494e503c387b8850/apps/studio/src/components/pipeline/stages/storyboard/components/StoryboardSectionDetail.tsx#L1040-L1064
- https://github.com/unicef/adt-studio/blob/792f2ac5828ca68c455aae3d494e503c387b8850/packages/types/src/section-tree-ops.ts#L184-L196

PR #315, opened 2026-04-22
feat: add word highlighting, manual glossary terms, and speech audio uploads

ravi-adt:
## Summary
- add word-level read-aloud highlighting with timestamp support, plus toggles in Speech and Accessibility settings
- add manual glossary term creation so users can add entries not generated by AI
- add per-row speech audio upload/replace and allow uploaded audio to use AI-generated timestamps
- align preview/export/runtime behavior with the new speech highlight and timestamp flow

### Speech
- added word-by-word read-aloud highlighting instead of only sentence/block highlighting
- added a Speech-stage `Word Highlight` toggle for book creators
- added per-text manual audio upload/replace in the Speech list
- made uploaded manual audio compatible with AI timestamp generation

### Accessibility
- added a reader-facing `Word highlight` toggle in Accessibility `Settings`
- persisted the setting in runtime state and wired it into playback behavior

### Glossary
- added a manual `Add Term` flow in the Glossary stage
- persisted manual glossary items so they survive reruns and flow into downstream catalog/package output

### Preview and Packaging
- updated preview/export/runtime behavior so highlight mode and timestamps stay consistent across environments
- fixed punctuation loss, fallback timestamp handling, and old blue-box flicker during rollout

## Testing
- `pnpm exec vitest run apps/api/src/routes/adt-preview.test.ts apps/api/src/routes/glossary.test.ts apps/api/src/routes/tts.test.ts apps/api/src/services/stage-runner.test.ts packages/pipeline/src/__tests__/glossary.test.ts packages/pipeline/src/__tests__/text-catalog.test.ts packages/pipeline/src/__tests__/package-web.test.ts assets/adt/modules/tts_highlighter_utils.test.js`
- `pnpm exec tsc --noEmit -p apps/api/tsconfig.json`
- `pnpm exec tsc --noEmit -p apps/studio/tsconfig.json`

## Notes
- manual uploaded audio now works with the existing timestamp flow
- a full Speech rerun still regenerates stage audio and can overwrite manually uploaded files; that behavior was not changed in this PR


elasticsounds:
Hey Ravi - I've made a lot of changes to make this functional. Here is the list.

- Word level highlighting in adt preview would break since it was using class container to do gather audio elements - switched this to content id which is more reliable.
- Word highlighter option is now in text to speech options area.
- Manual glossary term creation uses the book text for selecting glossary terms to use. AI generates the definition, selects variations and glossary.
- Per row speech upload was not working due to cache, fixed it so it functions. Will show in wavefrom and play.
- Seems to work better.

nicpottier:
Very cool set of features, having this broken out into separate PRs would have been good. Nice work on all of these though!

On highlighting, awesome feature to add, work great!
 * settings such as whether to enable highlighting belong in the settings menu, on the speech tab would be fine. That's also where rebuilds should be triggered. Currently the "word highlighting" toggle I take it controls whether we calculate word boundaries when rebuilding so that belongs in settings.
 * the "calculate word boundaries for all entries" feels like a weird thing (and also not very intuitive based on the icon). When is this used? I take it maybe when someone uploads an audio file? Seems like that should just be automatic based on the settings value? Or an option visible when you upload the item?
 * deleting all speech is maybe a useful thing to have but we should chat about a way to unify that kind of global destructive behavior. (I can imagine the same might be wanted for quizzes for example). The confirmation dialog uses a browser alert which we don't use anywhere else, that should be a normal dialog explaining what's going to happen etc..
 * The UI for shifting word boundaries allows me to set overlapping boundaries, is that by design? Will that work? Feels like maybe a UI that has a sliders could maybe be more user friendly but realize that's a lot of work. At the very least we should validate boundaries aren't overlapping unless that's by design.
  * The replace button is kind of giant, maybe it is just an icon with a tooltip?
  
  On Glossary:
 * The dialog says "Start typing to see words from the book" but nothing happens when I start typing? Is that by design?
  * After I add an item it is at the bottom of the list instead of alphabetical.
  * I like how manual entries are preserved on rebuilds by using a manual source field on the versioned list and merging those back in. That's a pattern we should try elsewhere.

elasticsounds:
List of changes:
- Moved timestamp generation from header to settings. (still able to trigger individual timestamps per item).
- Added step in Speech for calculate timestamps.
- Added timestamp toggle to landing page, disabled by default (it takes a long time to calculate these).
- Left delete all speech - we can explore a global update for this across the application.
- Updated UI for setting manual boundaries, will not allow you to enter a value before the preceding entry or after the following entry. You can have to manually update. Slider UI for sure in future version.
- Made the replace button just an icon with a tooltip.
- Start typing bug fixed for glossary.
- Moved manually added items to be alphabetically sorted.
- Agree on the manual tag - will follow up with another pull request on manually edited fields.

elasticsounds:
- fixed no cache support for regeneration of word highlighting step

elasticsounds:
- after adding manual glossary item, sets translation and speech to be re-run. Added banner at top of translation and speech as shortcut for user.

elasticsounds:
- Timestamps no longer overwritten is highlighting is disabled. Previously being overwritten by empty values.

elasticsounds:
- Auto-builds text-catalog when opening modal to add manual glossary terms - consider in a later update to generate text-catalog progressively after each step. Keeping translate to only translations.

elasticsounds:
@nicpottier ready for your review before merge

elasticsounds:
@copilot resolve the merge conflicts in this pull request

PR #344, opened 2026-04-27
Add Sections tab to stage sidebar with desktop thumbnails

elasticsounds:
## Summary
- Adds a Pages | Sections tab toggle to the stage sidebar; the Sections tab lists every non-pruned section across the book with an indexed row (number on the left, page/section caption on the right — section pill only shown when the page has more than one section), a desktop thumbnail, an activity-puzzle badge for activity sections, and the same loader the page list uses while the storyboard stage is rendering.
- Generates a 1280×800 desktop PNG per section via the existing Playwright renderer at the end of `web-rendering` (and after every `reRenderPage` / single-section regen), stored under `books/{label}/.thumbnails/{sectionId}.png` and served from `GET /api/books/:label/sections/:sectionId/thumbnail.png` with `?v={renderingVersion}` cache-busting; Playwright is still optional — if it can't be loaded, the renderer falls back to placeholders.
- Invalidates stale thumbnails inside `saveStoryboardNode` so any sectioning, rendering, AI edit, image insert, split, merge, or page-move drops the old PNGs and the next render regenerates them; no migration is needed but existing books require a re-render to populate thumbnails.
- New helpers: `Storage.{put,get,clear}SectionThumbnail*` in `@adt/storage`, `generateSectionThumbnails` in `@adt/pipeline`, `getSectionThumbnailUrl` in the studio client; `PageSummaryItem` now exposes `sectionType` per section and a per-page `renderingVersion`.

## Test plan
- [ ] `pnpm typecheck && pnpm --filter @adt/studio lint && pnpm test` (all green locally — 1008/1008)
- [ ] In the storyboard stage, run a full render on a book with mixed content + activity sections; confirm the Sections tab populates with thumbnails and the puzzle badge appears on activities
- [ ] Re-render a single section via AI edit; confirm that section's thumbnail updates immediately (cache-busted by version) without a page reload
- [ ] Trigger a sectioning save / split / merge; confirm stale thumbnails are cleared and regenerate on the next storyboard run
- [ ] Verify Pages tab behavior is unchanged


ksokolovic:
I reviewed the exact current head (`13a3ecfc`) against the current branches. I am closing this without merge because the branch has several blocking behavioral issues:

- Saving through the new Activities UI writes the new `ActivitiesOutput` shape, while ADT Preview still reads the legacy `.quizzes` shape. After an activity is saved, activities can disappear from Preview and `actNNN` pages are not handled there.
- The editor permits up to six multiple-choice options, but packaging silently truncates them to three and clamps the answer index. This can remove the authored correct answer and mark a different option as correct.
- The thumbnail lifecycle does not meet the PR acceptance claims: AI edits clear thumbnails without regenerating them, while storyboard rendering without Playwright can retain and serve stale thumbnails instead of showing placeholders.

The branch has also grown to roughly 4,500 added lines across 50 files, bundling the original Sections-tab work with broad Activities and extraction rewrites. It now conflicts with `main`, and the affected architecture has advanced substantially on `develop`. Typecheck, lint, and all 1,047 existing tests pass, but there are no focused tests covering these new integration paths.

At this point, bringing this branch forward safely would be a disproportionate lift and would duplicate or conflict with newer work. If section-thumbnail navigation is still desired, the better path is a focused PR from current `develop`, scoped to that UX and built on the current activity, preview, and pipeline architecture with dedicated lifecycle tests.

PR #354, opened 2026-04-28
Update README.md

elasticsounds:
Adding visual HTML header


elasticsounds (commented):
Check the header

elasticsounds (commented):
Updated banner header

PR #360, opened 2026-04-28
Revise README with updated image and badges

elasticsounds:
Updated README to include new image and badge links.


elasticsounds (commented):
Align center for pills and supported by

PR #384, opened 2026-05-04
Add persistent Save & Rerun action bar with stage-aware styling

elasticsounds:
## Summary
- Introduce `SettingsActionBar` + `SettingsActionButton` to replace the small portaled Save/Save & Rerun buttons in each pipeline stage settings panel. The bar shows dirty state inline ("Unsaved changes — rerun to apply" / "No unsaved changes") and emphasizes the action button using the current stage's color when there are pending changes.
- Wire all stage settings (extract, sectioning, storyboard, quizzes, captions, glossary, toc, translations, speech prompts, voice mappings) through the new component, computing dirty state from existing `dirty` records plus prompt/template drafts.
- Move the portal target out from beside the close button so the action bar can occupy a dedicated row under the stage header.
- Add a `colorHover` per stage in `stage-config.ts` and use it for the active-state button so hover darkens to the same hue instead of falling back to shadcn's default primary blue.

## Test plan
- [ ] Visit settings for each pipeline stage (extract, sectioning, storyboard, quizzes, captions, glossary, toc, translate, speech) — clean state shows muted bar; editing fields toggles to the stage-colored bar with the matching button hover.
- [ ] Confirm Storyboard `image-generation`, Translations `voices`, and Translations `speech-prompts` Save buttons work and stay disabled until edits exist.

### DIFF
--- a/apps/studio/src/components/pipeline/components/SettingsActionBar.tsx
+++ b/apps/studio/src/components/pipeline/components/SettingsActionBar.tsx
@@ -1,0 +1,96 @@
+import { ReactNode, createContext, useContext, ComponentProps } from "react"
+import { useParams } from "@tanstack/react-router"
+import { Trans } from "@lingui/react/macro"
+import { cn } from "@/lib/utils"
+import { Button } from "@/components/ui/button"
+import { STAGES } from "@/components/pipeline/stage-config"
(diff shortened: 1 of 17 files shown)


Eliezir:
fixed at #519

PR #391, opened 2026-05-06
Fix CMYK and soft-mask transparency in PDF image extraction

elasticsounds:
## Summary
- CMYK JPEGs now decode through mupdf and re-encode as RGB so browsers can render them; raw extraction is reserved for single-filter RGB/Gray DCT streams with no soft mask.
- Images with an `/SMask` are composed with their alpha into an RGBA PNG instead of arriving opaque, with `/Matte` un-matting applied when the SMask carries a matte color (handles RGB-3 and Gray-1; CMYK falls through unchanged).
- Added unit tests covering CMYK JPEG conversion, plain SMask alpha recovery, and matted SMask un-matting; updated three raven.pdf hashes that legitimately changed under the new pipeline.

## Test plan
- [x] `pnpm vitest run packages/pdf/src/__tests__/extract.test.ts` — 32 passing
- [x] `pnpm vitest run packages/pdf/src/__tests__/extract-raven.test.ts` — 12 passing
- [x] `pnpm typecheck`

### DIFF
--- a/packages/pdf/src/__tests__/extract-raven.test.ts
+++ b/packages/pdf/src/__tests__/extract-raven.test.ts
@@ -50,7 +50,7 @@ describe("raven.pdf extraction", () => {
     expect(imgs[0].width).toBe(546);
     expect(imgs[0].height).toBe(546);
     expect(imgs[1].imageId).toBe("pg001_im002");
-    expect(imgs[1].hash).toBe("14b8a1e04e930b83");
+    expect(imgs[1].hash).toBe("c9b639d01b6c22f6");
   });
 
   // -- Page 2 --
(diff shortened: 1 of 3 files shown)


elasticsounds:
Superseded by #442 (`elasticsounds/port-pdf-image-color-fixes`), which ported these fixes onto `develop` and merged on 2026-05-27.

All three changes from this PR are on `develop` today in `packages/pdf/src/extract.ts`:

| This PR | On `develop` |
|---|---|
| CMYK JPEG → decode via mupdf, re-encode as RGB | `normalizeToDisplayableRgb()` |
| `/SMask` alpha → composite into an RGBA PNG | `compositeColorAndMask()` (renamed from `composeWithSoftMask`) |
| `/Matte` un-matting (RGB-3 and Gray-1) | `readSoftMaskMatte()` |

The three tests came across too, with identical builders: `createPdfWithCmykJpegImage`, `createPdfWithSoftMaskImage`, `createPdfWithMattedSoftMaskImage`.

`develop` has since moved past this branch — #453 added the `classifyImageColorSpace()` fast path (`"displayable" | "needs-convert"`), plus CTM raster flip transforms and same-size image disambiguation. The raven baseline hashes here are stale for that reason. Rebasing this branch would regress those, so closing rather than reviving.

PR #393, opened 2026-05-06
[Electron] fix image replacement and AI image generation in desktop

Eliezir:
## Summary

- **Image replacement not updating storyboard in Electron**: `BASE_URL` is `http://localhost:<port>/api` in Electron (absolute with port), but the pipeline writes relative `/api/books/...` URLs into the stored rendering HTML. The regex built from `BASE_URL` never matched, so crop/upload/replace-from-book/AI-generate swaps silently left the old image src in place. Fixed by using a `renderedImageSrc` helper that always writes relative paths, matching what the stored HTML contains.

- **AI image generation fails with "Unknown parameter: 'response_format'"**: `@ai-sdk/openai` injects `response_format: "b64_json"` for any model not in its hardcoded `hasDefaultResponseFormat` set (currently only `"gpt-image-1"`), and the OpenAI API rejects that parameter for `gpt-image-2` and other newer variants. Fixed by bypassing the SDK entirely and using raw `fetch` to `/v1/images/generations`.

- **Refactor**: Extracted `callOpenAiImageApi` as a shared helper used by both `generateFreshImage` and `editImage`, removing the duplicated fetch/auth/error-parsing logic.

## Test plan

- [ ] In Electron desktop: crop an image in the storyboard — confirm the storyboard preview updates to show the cropped version
- [ ] In Electron desktop: upload a replacement image — confirm storyboard preview reflects the new image
- [ ] In Electron desktop: use "Replace from book" — confirm storyboard preview updates
- [ ] In Electron desktop: generate an AI image with `gpt-image-2` — confirm it no longer throws "Unknown parameter: 'response_format'"
- [ ] Run `pnpm test --filter @adt/llm` — all 3 image tests pass


Eliezir:
Closing in favor of a clean branch off main.

PR #407, opened 2026-05-08
Fix: desktop startup Issues

zignaggo:
- Remove manually port search and give responsability to SO
- Wait api sends by IPC the new port
- Add to splashscreen a debug screen
- Reorganize folder structure


Eliezir:
Known limitation (not addressed here): In dev mode the main window hardcodes http://localhost:5173, so pnpm dev:desktop will render whatever is currently served on that port

PR #386, opened 2026-05-05
Add custom quiz groups with explicit page selection and placement

elasticsounds:
## Summary
- Adds a `quiz_groups` config option: each group bundles user-picked source pages into one quiz with explicit placement (after the last source page or at the end of the book), as an alternative to the existing every-N-pages auto-batching.
- New landing-card UI on the quizzes step with radio buttons (auto vs. custom), a page picker dialog per quiz, a placement selector, and an "Add another quiz" button. Pipeline `buildQuizBatches` resolves groups into `{pages, afterPageId}` batches that flow into `generateQuiz` unchanged downstream.

## Test plan
- [ ] Pick 7 pages in one custom group → one quiz is generated covering all 7 pages, placed after the last selected page
- [ ] Add a second quiz group with placement "end of book" → second quiz is appended at the very end of the bundled output
- [ ] Switch back to "Every N pages" → auto-batching still works as before
- [ ] Verify all four locales (en, es, fr, pt-BR) render the new UI strings without missing translations

PR #425, opened 2026-05-19
Add custom quiz groups with explicit page selection and placement

elasticsounds:
## Summary
- Adds a `quiz_groups` config option: each group bundles user-picked source pages into one quiz with explicit placement (after the last source page or at the end of the book), as an alternative to the existing every-N-pages auto-batching.
- New landing-card UI on the quizzes step with radio buttons (auto vs. custom), a page picker dialog per quiz, a placement selector, and an "Add another quiz" button. Pipeline `buildQuizBatches` resolves groups into `{pages, afterPageId}` batches that flow into `generateQuiz` unchanged downstream.

## Test plan
- [ ] Pick 7 pages in one custom group → one quiz is generated covering all 7 pages, placed after the last selected page
- [ ] Add a second quiz group with placement "end of book" → second quiz is appended at the very end of the bundled output
- [ ] Switch back to "Every N pages" → auto-batching still works as before
- [ ] Verify all four locales (en, es, fr, pt-BR) render the new UI strings without missing translations


ksokolovic:
Closing without merge after final review. This PR is an exact duplicate of #386: both point to the same head commit, 548d767, and contain the same 70-file diff. It therefore inherits the same merge conflicts and failing focused Studio tests, while the advertised quiz workflow was already delivered by #455. No cherry-pick is recommended.

PR #470, opened 2026-06-02
Add inline translation review with configurable LLM judge

simba-mudz:
**Description**
This PR adds v1 of inline translation review for the Language page. Users can run an LLM-based review on the currently visible translations, see item-level results directly beside each translation, apply suggested fixes, or mark flagged translations as accepted anyway.

**What Changed**
- Added translation evaluation schemas and config support in `@adt/types`.
- Added API routes and services for running, storing, and retrieving translation review results.
- Added inline review UI to the Language page:
  - Review button in the page header.
  - Review summary and filters.
  - Needs attention / acceptable / accepted anyway states.
  - Rationale display for review decisions.
  - Suggested translation display and apply-suggestion flow.
- Added translation review judge settings under Language settings:
  - General settings for non-technical users.
  - Advanced settings hidden behind an expandable section.
  - Default judge model set to `openai:/gpt-5.4`.
- Added validation for suggested fixes so the judge only shows higher-confidence replacement translations.
- Added local terminology evidence to make reruns more consistent after applying suggestions.
- Added Studio-only persistence for review results and stale-review detection.
- Updated Lingui catalogs for all supported locales.


ksokolovic:
Nice feature, and the fundamentals are solid: entity versioning is respected, server state is TanStack-Query-driven, the apply-suggestion flow persists correctly without clobbering sibling edits, evaluations use deterministic caching + version/config-hash **stale detection** (rather than silently overwriting), the two-pass validate/repair suggestion loop is a thoughtful touch, and test coverage (routes/runner/service with an injectable model + task service) is genuinely good.

A few things I'd want addressed before merge, then some polish.

### 🔴 Blocking

**1. Runner mutates global `process.env.OPENAI_API_KEY` — real concurrency bug + deviates from the repo's credentials pattern.**
`apps/api/src/services/translation-evaluation-runner.ts:512-513` sets the process-global env key and `:616-620` restores it in `finally`, wrapping all the async judge calls. But `createLLMModel` already takes a per-request `credentials.openaiApiKey` (`packages/llm/src/client.ts:33`, forwarded to `resolveModel` at `:326`), and that's exactly how `stage-runner.ts:611-700` does it (`createLLMModel({ …, credentials: llmCredentials })`). Nothing else in the app routes keys through `process.env`.
*Failure scenario (the common desktop / bring-your-own-key case, where the server has no `OPENAI_API_KEY`):* two reviews run concurrently (two languages, or a re-run). Both capture `previousOpenAIKey = undefined`; the first to finish runs `delete process.env.OPENAI_API_KEY` while the second is still mid-flight, and the second's remaining pages all fail with "Translation judge failed: … API key".
*Fix:* pass `credentials: { openaiApiKey: apiKey }` into `createLLMModel` and delete the entire env save/restore block.

**2. `sq.po` (Albanian) not updated — CI will fail.**
The PR updates `en/es/fr/pt-BR.po` but not `sq.po`, which is a supported locale. The `i18n` CI job + `lingui/no-unlocalized-strings` lint will fail, and Albanian users get untranslated review UI. Run `pnpm --filter @adt/studio extract` and translate the new keys.

**3. Malformed default judge model `"openai:/gpt-5.4"` (extra slash).**
Declared twice — `packages/types/src/translation-evaluation.ts:26` and re-declared in `apps/studio/src/components/pipeline/stages/languages/LanguageSettings.tsx:50` (used as the default form state and persisted at `:371`, shown as the picker placeholder at `:676`). Everywhere else the format is `provider:model` with no slash. `resolveModel` (`client.ts:320`) parses this to provider `openai`, model `/gpt-5.4`; it only works today because `normalizeJudgeModel` (`runner.ts:111-113`) strips the slash with a bespoke regex — a tell that it's a papered-over typo. The raw slashed value is also what gets hashed into `eval_config_hash`, while the model actually used is the normalized form.
*Fix:* set both constants to `"openai:gpt-5.4"`, have the frontend **import** the `@adt/types` constant instead of re-declaring it, and drop the normalize regex (keep only as defense if you like).

### 🟠 Should fix

**4. "Configurable judge model" only actually works for OpenAI.** `judge_model` is any string and the UI exposes it, but the run route/runner only ever wire an OpenAI key (`X-OpenAI-Key`, `OPENAI_API_KEY`, error text "OpenAI API key required"). Set `judge_model: "anthropic:…"`/`"google:…"` and `resolveModel` needs `anthropicApiKey`/`googleApiKey` that are never supplied → runtime failure. Either constrain the model to OpenAI for v1, or thread real per-provider credentials (folds into fix #1; `llmCredentials` in `stage-runner.ts` already carries all providers). Relevant given the multi-provider direction in #557.

**5. Async review status isn't announced to screen readers.** In `LanguageView.tsx` the running banner, results summary/filter counts, and stale/failed banners have no `role="status"`/`role="alert"`/`aria-live` (grep for those attributes in the file returns nothing). A screen-reader user clicks Review and never hears "3 need attention". Wrap the running/complete banner in a polite live region; confirm the `ui/alert` components set `role="alert"`.

**6. Applying a suggestion doesn't invalidate downstream speech.** `saveTranslation` (`LanguageView.tsx:409-420`) — the persistence path for apply-suggestion (`applyReviewSuggestion:438 → updateEntry:425 → pendingEntries → saveTranslation`) — invalidates only `["books",label,"text-catalog"]` and the evaluations key (`:416-417`), not `["books",label,"tts"]` or `step-status`. (Contrast `handleDeleteTTS:240-243`, which does invalidate both.) After applying + saving, previously-generated audio for that entry silently corresponds to the old text and nothing flags it stale. Pre-existing to manual translation editing, but "apply suggestion" makes it far easier to hit — at least invalidate tts / surface a stale-audio signal.

### 🟡 Polish

- **Redundant enable flags** — `TranslationEvaluationConfig` has both `enable_translation_evaluation` and `enabled` (`translation-evaluation.ts:72-73`; resolver `:115` reads `enable_translation_evaluation ?? enabled ?? true`). Pick one.
- **Untranslated enum badges** — `LanguageView.tsx:136` (`{issueType}`) and `:141` (`{item.severity}`) render raw model enums ("meaning", "high") while the same concepts are localized in settings (`LanguageSettings.tsx:728`). Map to localized labels.
- **Repeated DB open/close** — `translation-evaluation-service.ts` opens+closes the book DB 3× per status read (`parseLatestRows` + `getLatestNodeVersion`×2), and the run route re-opens again in `buildTranslationEvaluationRunRequest`. One connection per request would be cleaner (WASM-SQLite open + `.lock` isn't free). Not a correctness issue.
- **No force-rerun** — the `run` "current" short-circuit (`routes.ts:428-444`) returns the cached result on identical scope/config; there's no way to force a fresh evaluation.

### ⚠️ Merge mechanics
Branch is behind current `main` by the `develop→main` promotion (#550), which added the **easy-read** stage and reworked `text-catalog` + `LanguageView` — exactly the files this PR touches. It needs a real re-merge, not a trivial one; re-verify the `text-catalog.ts` and `LanguageView.tsx` changes land correctly on top.

**Verdict:** request changes — #1 and #2 block, #3 is a one-line correctness fix. The rest is straightforward. Core design is good.

ksokolovic (dismissed):
Full review in my comment above — 1 (global process.env OpenAI-key mutation) and 2 (missing sq.po → i18n CI failure) are blocking; 3 (malformed "openai:/gpt-5.4" default) is a one-line correctness fix. The rest is should-fix/polish.

ksokolovic:
@simba-mudz — heads up: I've pushed six commits to this branch and retargeted the PR to `develop`. Apologies for doing that without checking with you first; happy to revert any of it if you'd rather take it yourself.

The reason was the merge. This branch was cut on 2 June and `main` has moved a long way since — most importantly #550, which brought the easy-read stage and reworked `text-catalog` and `LanguageView`, exactly the files this PR touches. That left the PR conflicting, and the conflicts weren't the kind you'd want to resolve blind.

Everything is append-only — your `1b9617ca` is still in the history and nothing was rewritten.

### The merge

The Languages settings and view were the substantial part:

- `develop` moved the settings tab definitions out to `settings-tabs.ts`, so the Translation Review tab moved with them.
- It also replaced the hand-rolled `createPortal` header buttons with `useStageSettingsBar`. Your "Save Review Settings" button is gone as a result — the review tab now registers with that shared bar instead, and the generic `save()` already persists the same overrides your `saveReviewSettings` did.
- `LanguageView` needed the review UI re-applied on top of develop's version. The entry list now runs category filter → review filter → missing-audio filter, so your review filter composes with both rather than replacing either.

One thing worth knowing for future rebases: **#556 reformatted `LanguageView.tsx` with semicolons** — 521 semicolon-terminated lines, against 0 in `LanguageSettings.tsx` on the same branch. There's no prettier config and no format check in CI, so nothing caught it. That reformat is most of why develop's diff on that file looks like a rewrite, and it's what made this merge expensive. I deliberately left it alone rather than fold a whole-file reformat into this PR.

### Review findings

All ten from the review are addressed:

**Blocking** — the `process.env.OPENAI_API_KEY` mutation is gone, replaced by per-request credentials into `createLLMModel`. Since the judge model is configurable, I threaded every provider's credentials rather than constraining to OpenAI, and the required key is now resolved from the model's provider. The default judge model is `openai:gpt-5.4` (the stray slash meant `resolveModel` was parsing the model as `/gpt-5.4`), and the frontend imports the constant instead of keeping its own copy — that duplication is how the two drifted apart. `normalizeJudgeModel` stays, since books configured while the bad default shipped still have it saved.

`sq.po` turned out not to be an oversight on your part at all: Albanian landed in #467 *after* you cut this branch, so the file simply didn't exist here. It's added and translated now.

**Should-fix** — the review banners are polite live regions so the async result is announced, and `saveTranslation` now invalidates `tts`/`step-status` so audio generated from the old text doesn't silently go stale.

**Polish** — dropped the unused `enabled` config flag, localized the issue-type and severity badges, and collapsed the evaluation status read to one database connection.

I skipped the force-rerun idea. Judge calls are cached on disk keyed by model, prompt, schema and temperature, so a forced re-run with unchanged inputs returns the identical cached answer — and the cache short-circuit already doesn't apply to failed runs. Making it meaningful means busting the LLM cache from the runner, which is a change to shared infrastructure and belongs on its own.

### Verification

1790 tests pass, typecheck and lint are clean, and all five locales are complete so the i18n job should be green.

I also drove the review UI in a browser against a seeded book, since `LanguageView` has no test coverage and I didn't want to merge a large re-application on a typecheck alone. The filters, apply-suggestion, and accept-anyway round-trip all behave correctly, and accept-anyway persists as a new version.

Two things I could not verify: no judge call has been made against a live Anthropic, Google, or custom provider — that wiring mirrors `stage-runner` and is unit-tested, but never exercised. And I checked the live regions exist in the accessibility tree without testing them with an actual screen reader.

Worth a careful look at the `LanguageView` merge in particular — that's where I'd expect a mistake if there is one.

simba-mudz:
@ksokolovic.... Thanks for the detailed follow-up and for keeping the history append-only. I’ve reviewed the six commits and it looks great!

I found two remaining points:

1. `enable_translation_evaluation` is persisted by the settings UI but is not currently enforced by `LanguageView` or the API, so disabling translation review has no effect. I can address this.
2. The route currently groups all selected entries into one synthetic `visible` page. Since `batch_size` operates on pages, it does not limit the size of a normal UI request. Before changing this, could you confirm whether the intended batching unit is pages, entries, or concurrent judge calls?

Once we agree on the batching behavior, I’m happy to implement the remaining change and run the relevant checks.

ksokolovic:
Thanks for going through all six — and both of these are good catches.

**On (1):** confirmed. Worth noting it's pre-existing rather than something these commits changed: `enable_translation_evaluation` is written by the settings form and returned by the resolver, but the only reader is `LanguageSettings.tsx` loading it back into the toggle — neither `LanguageView`'s Review button nor the run route consults it, so disabling review currently has no effect. (The flag I dropped in these commits was the separate redundant `enabled` duplicate; this one was simply never wired up.) Enforcing it at the Review button and the run route is the natural home, so please do take it.

**On (2):** you've read it exactly right. The run route folds all selected entries into a single synthetic `visible` page, the runner issues one judge call per page and walks the pages sequentially, and `batch_size` only groups pages for the progress messages — so for a normal UI request it chunks a one-element array and has no effect.

My read of the intended unit, since you asked: the judge system prompt is explicit that it evaluates a whole page's entries together — "Use all entries on the page as context when judging each individual entry," and "Treat existing translations on the page as local terminology memory" for consistent term choices. So **the page is the natural judge-call boundary**, and what `batch_size` controls there is really "pages per progress step," not "entries per call." The gap the UI path exposes is a different axis the current knobs don't cover: there's no ceiling on *entries per call*, so a large "review visible" selection becomes one very large prompt.

An entries-per-call cap would bound that, but it isn't free — splitting a selection into multiple calls breaks the shared-page context the prompt leans on, so cross-entry terminology consistency weakens at the split points. Bounded cost/context vs. that consistency, plus whether sequential-vs-concurrent is worth revisiting, is a product-and-cost decision I don't want to presume on — so I'd defer to you and a maintainer on the intended shape rather than pick one. Happy to implement whichever direction you land on, and to pair on the batching change if that's useful.

simba-mudz:
Thanks, agreed on both points. I’ll implement enforcement of `enable_translation_evaluation` in both the UI and API.

For batching, my proposed v1 behavior is to group the selected entries by their real page and evaluate one page per judge call, sequentially. That preserves the page-level context while preventing a whole-book selection from becoming one unbounded synthetic prompt. I would avoid splitting individual pages for now.

Since `batch_size` does not currently affect execution, I propose hiding/removing that user-facing setting until we define whether it should control concurrency. I’ll also update the cached-scope comparison for multi-page requests and add route tests covering real page grouping.

I’ll proceed with the changes and verification.

ksokolovic (approved):
These two look right to me — verified locally on `9758f95b`.

**Page-scoped reviews (`94486f31`):** grouping by real page with a `book-level` catch-all is exactly the v1 we landed on — it keeps the per-page context the judge prompt relies on while bounding what used to be one unbounded synthetic prompt. The `buildScopeHash` round-trip is coherent end to end (built in the request → persisted into metadata → compared in the "current" short-circuit), and dropping `equalEntryIds` left no dead code behind.

**Settings enforcement (`9758f95b`):** `enable_translation_evaluation` is now enforced in both places it was missing — the Review button and the run route's 409 — which closes the persisted-but-inert gap. Removing the `batch_size` control until it has a defined meaning is the right call.

Typecheck is clean and the 40 tests across the four touched files pass locally, including the new route tests for real page grouping.

Approving — nice work on the batching, the page-boundary choice is the right one.

PR #477, opened 2026-06-03
Storyboard thumbnail loading: skeletons, empty states, and cache warming

Eliezir:
## Summary

Improves the storyboard thumbnail experience on two fronts: **perceived speed** (skeletons + clean empty states) and **actual cold-load speed** (pre-warming the section-screenshot cache during the render stage).

### Studio — skeleton & empty states (`feat(studio)`)
- Add a shadcn `Skeleton` component with a glow shimmer sweep (`skeleton-shimmer` keyframe, respects `prefers-reduced-motion`).
- Show a glowing skeleton list in the storyboard sidebar while pages load (previously fell through to the "No pages extracted yet" text).
- Per-thumbnail skeleton while a section screenshot loads, with a guard against already-cached images that never fire `onLoad` (which would leave the thumb blank).
- Comparison popover: icon-based empty states for **not-rendered** and **pruned** sections; shows image-or-empty-state only (no loader — the popover only appears once we already have the image or know its empty state).
- Serve section screenshots with an immutable `Cache-Control` so re-opens are instant from browser cache (the URL is already version- and content-hash-keyed).

### API — cold-load cache warming (`perf(api)`)
- Cold opens previously generated every section screenshot lazily via Playwright on first view (~1–3s each, largely serialized) → slow first load.
- Extract the section-screenshot HTML build + hash + cache-path into a shared helper (`section-screenshot.ts`) used by **both** the on-demand endpoint and the warmer, so their cache keys can't drift.
- After the storyboard render step completes, pre-render each section's **desktop** screenshot into `.section-renders/` (bounded concurrency, best-effort: failures are logged and fall back to lazy generation).
- Refactor the screenshot endpoint onto the shared helper (no behavior change).

**Result:** the storyboard opens against a warm disk cache and the endpoint just serves files, independent of platform or HTTP connection limits.

## Verification
- `tsc` clean for `@adt/studio` and `@adt/api`.
- Tests pass: stage-runner (13), debug (17), screenshot-html (3).
- No new user-visible strings (reused existing `<Trans>`), so no i18n re-extract needed.

## Follow-up (not in this PR)
- Thumbnail-size variant: the sidebar displays at ~96px but the endpoint renders/serves a full desktop-width PNG; a `?size=thumb` variant would cut payload/decode for the many sidebar thumbs.

### DIFF
--- a/apps/api/src/routes/pages.ts
+++ b/apps/api/src/routes/pages.ts
@@ -952,7 +952,7 @@ export function createPageRoutes(
       }
 
       const buffer = fs.readFileSync(cachePath)
-      c.header("Cache-Control", "private, max-age=300")
+      c.header("Cache-Control", "private, max-age=31536000, immutable")
       c.header("Content-Type", "image/png")
       return c.body(buffer)
     }
--- a/apps/api/src/routes/pages.ts
+++ b/apps/api/src/routes/pages.ts
@@ -18,12 +18,12 @@ import {
   applyCrop,
   generateStyleguide,
   buildStyleguideGenerationConfig,
-  buildScreenshotHtml,
   createScreenshotRenderer,
   SCREENSHOT_VIEWPORTS,
   isFixedLayoutBook,
   type ScreenshotRenderer,
 } from "@adt/pipeline"
+import { prepareSectionScreenshot, writeSectionScreenshot } from "../services/section-screenshot.js"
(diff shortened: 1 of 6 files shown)


Eliezir:
Closing this, Zig is working on a new interface for the pipepline that will replace the changes made on this branch

PR #498, opened 2026-06-10
fix(pipeline): enforce semantic heading tags in LLM-rendered HTML

### LINKED ISSUE #494: [Storyboard] Headings detected during processing are not rendered as semantic HTML headings
### Problem
Some content is correctly identified as headings during sectioning, but in the storyboard/output it is rendered as styled text (e.g., larger font) inside generic elements (e.g., `div`), rather than semantic heading tags (`h1`, `h2`, etc.).

### Expected behavior
- Headings should be rendered using semantic HTML tags (`h1`, `h2`, etc.).
- The document should preserve a valid heading hierarchy.

### Actual behavior
- Headings are visually styled but not semantically marked.
- Rendered as non-heading elements (e.g., `div` with larger font).

### Impact
- Screen readers cannot navigate the document structure.
- Loss of accessibility and document hierarchy.

### Suggested direction
Ensure that elements identified as headings during processing are:
- Rendered using appropriate semantic HTML heading tags
- Maintaining correct hierarchical structure

@eyadwin do provide more context.

gbergengruen:
## Summary

Fixes #494 — content correctly identified as headings during page-sectioning was rendered in the storyboard as visually styled generic elements (e.g. `<div class="text-2xl font-bold">`) instead of semantic `<h1>`–`<h6>` tags, so screen readers could not navigate the document structure.

## Root cause

The `role: "heading"` metadata flows intact from sectioning all the way into the render stage, but the LLM render path never used it and nothing enforced it:

- The **template** render path already emits `<h2 data-id=...>` for heading leaves (`templates/_render_node.liquid`) — correct.
- The **LLM** render path (`llm`, `llm-overlay` strategies and all 8 `activity_*` section types) had no rule about heading tags in its prompts — rule 15 of `web_generation_html.liquid` even pushed text leaves toward inline `<span>`s — and `validateSectionHtml()` never checked heading semantics, even though its caller already knows exactly which data-ids are headings via `leaf_texts.text_type`.

The LLM had to guess the tag from the page image alone and frequently emitted styled `div`s.

## Changes

**Validator enforcement** (follows the existing `WRITABLE_SECTION_TYPES` fail-loudly precedent):
- `packages/pipeline/src/validate-html.ts`: new `headingTextIds` option. An element whose `data-id` belongs to a heading leaf must be an `<h1>`–`<h6>`, sit inside one (split headings render as `<span data-id>` fragments inside a single heading element), or use the ARIA equivalent (`role="heading"` + numeric `aria-level`). Violations produce an actionable error that feeds the existing LLM retry loop.
- `packages/pipeline/src/render-llm.ts`: new `collectHeadingTextIds()` wired into `validateWebRendering` — the single chokepoint covering LLM generation, activity rendering, and the visual-review loop.
- The template path is intentionally **not** wired (no retry loop there; it already renders headings correctly for the common shapes).

**Prompts** (cache invalidation on re-render is by design; existing storyboards are untouched until re-rendered):
- `web_generation_html.liquid`: new rule 16 requiring semantic heading tags (default `<h2>`, `<h1>` only for the page's main title, `<h3>` for subordinate subheadings) + a carve-out in rule 15's inline-span rule. Also fixes a pre-existing typo in rule 15 ("use render each").
- `web_generation_html_overlay.liquid`: same rule as Fundamental Rule 13.
- The 8 `activity_*.liquid` render prompts: heading-role leaves must always use a heading tag.
- `html_edit.liquid`: AI edits must preserve existing heading tags (the edit flow has no leaf-role data, so this is prompt-only).

**Tests**: 13 new Vitest cases across `validate-html.test.ts` and `render-llm.test.ts`, including pass cases (h1–h6, split heading spans inside one `<h2>`, ARIA equivalent), fail cases (styled `div`, bare inline `span`, `role="heading"` without a valid `aria-level`), a regression guard that omitting the option changes nothing, and an integration-style test proving a `<div>`-rendered heading is rejected and `<h2>` passes through the real validation wiring.

## Out of scope (follow-up)

Heading **level** hierarchy: sectioning carries no level data, so adding it would touch the sectioning schema/prompts and invalidate every book's sectioning cache. All headings default to `<h2>` (matching the template path), which already restores screen-reader heading navigation. Level detection in sectioning + strict hierarchy validation should be a separate issue.

## Verification

- Full repo suite: **1497 tests / 115 files pass**; `pnpm typecheck` clean.
- Independent code-review pass (7 review angles + adversarial verification) found no regressions.
- To verify end-to-end: re-render an LLM-strategy section on a test book and inspect the stored HTML for `<h2 data-id=...>` on heading leaves.


elasticsounds:
Fixed in #688

PR #500, opened 2026-06-11
WebPub export: reader-ready content (no embedded UI) + working quizzes

### LINKED ISSUE #488: WebPub export should be reader-ready content, not a self-contained app
### Problem / motivation

When someone exports an ADT as WebPub, the goal is for them to read it in a **dedicated reading app** — the way you'd open an EPUB in Apple Books or Thorium. In that world the reading app owns the experience (navigation, read-aloud, glossary, sign language, easy-read, theming, settings), and the exported file should just be the *book*: clean, standards-based content the reader knows how to present.

Today the two exports behave very differently. **EPUB** already ships as a book — it strips out our embedded viewer and leans on the reading app, following the [Readium WebPub EPUB profile](https://readium.org/webpub-manifest/profiles/epub.html) ideas. **WebPub** still ships the entire ADT viewer app embedded inside the package, so it behaves like a standalone website rather than a book — which means a reader app opening it gets a whole UI baked into the content, causing duplicated or conflicting interfaces and a worse reading experience.

There's one nuance: even with the viewer stripped, the content pages still have *in-content* interactivity (e.g. the quizzes UI). Some of those interactions need to drive the reader itself — for example, completing a quiz should advance to the next page. So the content needs a way to **emit events to the host reader** (via `postMessage` or a similar contract) so the reader can react and update its own page state.

### EPUB vs. WebPub today

| Behaviour | EPUB export | WebPub export |
|---|---|---|
| Embedded ADT viewer (`base.bundle.*` JS) | **Stripped** — reader app provides the UI | **Bundled** — content ships as a full app |
| Navigation / TOC | Real `toc.xhtml` (`nav epub:type="toc"`) + `toc.ncx`, from the LLM TOC | Only `readingOrder` in `manifest.json`; nav still served by the embedded viewer |
| Read-aloud | **As data** — SMIL media overlays + `media:*` metadata | Played by the embedded viewer's TTS engine |
| Cover | Declared (`cover-image`) | Declared (`rel: "cover"`) |
| Page list | not emitted | not emitted |
| Landmarks | not emitted | not emitted |
| Accessibility metadata (schema.org) | not emitted | not emitted |
| Content→reader events (e.g. quiz → next page) | none | none — navigation handled internally by the viewer |
| ADT-specific features (sign language, glossary, easy-read) | Files carried along, but unused | Driven by the embedded viewer |

The short version: **EPUB already treats the export as a book; WebPub still treats it as an app.** This issue is about bringing WebPub up to the same "reader-ready content" bar and finishing the few pieces neither export has yet.

### Already solved in the EPUB export (reusable)

These are done for EPUB and are the model to follow / port for WebPub:

- **Stripping the embedded viewer** — `stripRuntimeBundle()` (`packages/pipeline/src/package-epub.ts:581`) deletes `base.bundle.*` and removes its `<script>` tags. Near copy-paste for WebPub (run it on `webpub/`).
- **TOC as data** — `buildNavDocument()` produces a real table of contents from the LLM-generated TOC.
- **Read-aloud as data, not a built-in player** — `generateSmilFiles()` + `media:duration`/`media:active-class` in the OPF. The word-timing/sync data is already computed; for WebPub this just needs re-serializing into [Guided Navigation](https://readium.org/guided-navigation/) JSON instead of SMIL — not new logic.

### Missing in both exports (net-new work)

- **Page list** — `<nav epub:type="page-list">` for EPUB / `pageList` collection for WebPub (the NCX currently hardcodes `dtb:totalPageCount = 0`).
- **Landmarks** — bodymatter/cover/toc landmarks so readers can jump to key sections.
- **Accessibility metadata** — schema.org `accessMode`, `accessModeSufficient`, `feature`, `conformsTo`, `summary` in the manifest/OPF.
- **Content→reader event channel** — a defined `postMessage` (or equivalent) contract so in-content UI like the quizzes can notify the host reader of events (e.g. "navigate to next page", progress/score updates) and the reader can update its page state accordingly.
- **A convention for ADT-specific features** — sign language video, glossary, easy-read have no community standard. Today these `content/i18n/*.json` files are carried along but unused by readers. We need a documented sidecar (and/or Guided Navigation alternates for audio/video) that a reader can consume.

### Proposed solution

Make WebPub export produce **reader-ready content + data, with no embedded viewer UI**, conforming to the [Readium Web Publication Manifest](https://readium.org/webpub-manifest/) and its [EPUB profile](https://readium.org/webpub-manifest/profiles/epub.html):

1. **Drop the embedded ADT viewer** from `webpub/` (reuse the EPUB approach) so the package is the book, not an app.
2. **Put navigation/structure in the manifest** — `toc`, `pageList`, `landmarks`, cover, and schema.org `accessibility` metadata — so the reader builds its own UI.
3. **Express read-aloud as Guided Navigation** (the WebPub-standard equivalent of EPUB Media Overlays), ported from our existing word-timing data.
4. **Define a content→reader event contract** (`postMessage` or similar) for the interactivity that stays in the content — e.g. the quizzes UI requesting next-page navigation or reporting completion — so the host reader can update its page/state.
5. **Define a documented convention for ADT-specific features** so readers can render sign language, glossary and easy-read

Where it makes sense, the net-new pieces (page list, landmarks, accessibility metadata) should be backported into the EPUB export too.

### Open decisions

- Clean break (WebPub = pure standards-based content) vs. staying openable in generic Readium readers like Thorium?
- Keep `content/i18n/*.json` as the reader's feature source, or migrate fully into manifest-standard structures?
- What's the event contract — raw `postMessage` with a small typed schema, or an existing standard (e.g. a Readium/EPUB scripting interaction convention)?

### LINKED ISSUE #497: TOC generation produces an empty table of contents for fixed-layout books
### Summary

Table-of-contents generation always returns an **empty TOC** (`{"entries":[]}`) for **fixed-layout books**, in both `dynamic` and `extract` modes. Re-running the TOC step never helps — the result is deterministic.

### Steps to reproduce

1. Open a fixed-layout book (all pages sectioned as `sectionType: "fixed-layout-page"`).
2. Run the **Table of Contents** step (either `dynamic` or `extract` mode).
3. Observe the `toc-generation` node: `{"entries":[],"pageCount":N,"generatedAt":"…"}` — no entries.
4. Deleting the node and re-running reproduces the same empty result.

Observed on a real 27-page fixed-layout book (`LP-4-Volcanoes`): every TOC run yields `entries: []`.

### Root cause

`generateToc` builds the TOC only from section leaves whose role is `heading`:

- `packages/pipeline/src/toc-generation.ts:20` — `const HEADING_ROLE_TYPES = new Set(["heading"])`
- `collectHeadingsAndToc()` walks each rendered, non-pruned section for a `heading`-role leaf.
- `packages/pipeline/src/toc-generation.ts:148-154` — if `headings.length === 0`, it **returns an empty TOC early, without ever calling the LLM**.

**Fixed-layout page-sectioning never emits a `heading` role** — fixed-layout sections only contain `text` and `image` roles. So `collectHeadingsAndToc` finds zero headings on every page, and the early return fires every time.

Evidence from the affected book: all 27 pages are `fixed-layout-page`; the latest page-sectioning data contains only `"role":"text"` / `"role":"image"` and **zero** `"role":"heading"`.

A secondary consequence: because the early return happens before the LLM call, **`extract` mode can't even fall back to the printed TOC text** (`originalTocText` is collected but never used when there are no heading roles).

### Proposed fixes (combination recommended)

1. **Relax the early return** so the LLM still runs when an original printed TOC (`originalTocText`) is available, even with no heading-role leaves — fixes `extract` mode for fixed-layout books that have a TOC page.
2. **Infer headings for fixed-layout** — derive candidate section titles from `text`-role nodes (e.g. largest / topmost text per page, or a per-page title) so `dynamic` mode has something to work from.
3. (Larger) **Emit `heading` roles in fixed-layout sectioning** so the problem is fixed at the source.

### Impact

Fixed-layout books ship with no navigable table of contents — degrades navigation and accessibility (the manifest/reader TOC is empty). Affects both the in-app reader and exported EPUB/WebPub (which derive their TOC from this node).

### Notes

Pre-existing limitation, unrelated to the WebPub export reader work — surfaced while exporting a fixed-layout book.

Eliezir:
## Summary

Makes the **WebPub export reader-ready content instead of a self-contained app**, so an ADT opens cleanly in a dedicated reading app (the reader owns the UI; the package is just the book + data). Addresses #488. Closes #497 (empty TOC for fixed-layout books).

Previously a `.webpub` shipped the entire ADT React runtime, so it behaved like a standalone website and fought a host reader's own UI. Now it ships content + standards-based manifest + the feature-data contract, with interactive activities still working via a small dedicated bundle.

## What changed

- **Strip the embedded runtime** from WebPub — removes the React bundle, offline preloader, and SCORM adapter (files + `<script>` tags). The shared `stripRuntimeBundle` is now used by both the WebPub and EPUB packagers.
- **Navigation as manifest data** — emit Readium `toc` (nested by level), `pageList`, and `landmarks` collections so the reader builds its own navigation UI.
- **Accessibility metadata** — schema.org `accessMode` / `accessModeSufficient` / `feature` / `hazard` / `summary` derived from the book's feature flags (no unverified conformance claims).
- **Fixed-layout books render correctly in readers** — declare fixed layout in the manifest (`metadata.layout: "fixed"` + `presentation.fit`/`spread`) so a reader doesn't treat the absolutely-positioned page as reflowable and collapse it to 0 height. Pages scale to the viewport via the shared web fit script (`fixedLayoutWebFit`, from #576), which is neutralized for EPUB where the reader sizes pre-paginated pages itself.
- **Table of contents for fixed-layout books** — TOC generation now falls back to the semantic `page-sectioning` tree (which carries `heading` roles) when a page renders fixed-layout, instead of the positioned render tree that has none. Fixes the always-empty TOC (#497). (Books packaged before this fix need the TOC step re-run to populate `toc.json` / the manifest `toc`.)
- **Quizzes in WebPub** — a standalone `activities.bundle.local.js` (built from `apps/adt-runtime`, ~309 KB vs the 692 KB full runtime) that reuses the existing activity validators but loads only config/pages/translations and renders a minimal Submit/Next control. Injected only into pages that contain an activity. Mounts `<Toaster>` and `<ActivityConfetti>` so quiz feedback (toasts + correct-answer confetti) works.
- **Data-contract doc** — `docs/WEBPUB_EXPORT.md` documents the package layout, manifest shape, `config.json` feature flags, and the `content/i18n/**` sidecar files for reader authors.
- **Manifest/package hygiene** — drop SCORM/non-reader files (`imsmanifest.xml`, `AGENTS.md`) from the export instead of shipping and listing them; label resources with a complete extension→MIME map (fonts, video, images no longer fall back to `application/octet-stream`); and exclude reading-order documents from the `resources` collection (Readium holds only non-reading-order items there).
- **Packaging refactor** — moved the packagers into `packages/pipeline/src/packaging/` (one file per format: `web`, `webpub`, `epub`, `strip-runtime-bundle`) and extracted format-agnostic helpers out into `html-escape.ts`, `tailwind.ts`, and `html-semantics.ts` so nothing outside `packaging/` reaches into a format file for utilities.

## Key decisions

- **Guided Navigation dropped.** For pre-recorded read-aloud in generic readers (Thorium), EPUB is the right tool (it already works via SMIL). The WebPub's own reader uses the shipped `content/i18n/**` timecode data directly.
- **No `postMessage` channel needed.** On success the activities bundle navigates the iframe (`window.location.href`) to the next reading-order page; the host reader tracks the change via the manifest `readingOrder`. Verified working in a reader.
- **PNLD export split out.** The PNLD (`.zip`) export format that was previously on this branch now lives in its own PR (#577), stacked on this one.

## Out of scope / follow-ups

- WebPub fixed-layout pages reserve the reader's ~96px dock band via `fixedLayoutWebFit`; the stripped WebPub has no dock, so that leaves a small empty band at the page bottom. Cosmetic; can be zeroed for WebPub in a follow-up.

## Testing

- `pnpm typecheck` and the pipeline test suite pass (incl. tests for `nestTocEntries`, the runtime-strip, `pageList`/`landmarks`, accessibility metadata, `injectActivitiesBundle`, the fixed-layout manifest hints, the fit-neutralization gate, and fixed-layout TOC heading fallback).
- Exported a real fixed-layout book to WebPub and confirmed in a reader: no embedded viewer; pages scale to fit; manifest carries `toc`/`pageList`/`landmarks` + `accessibility`; quizzes render + validate + advance pages.


apps/adt-runtime/src/activities-entry.tsx
zignaggo, line 1: it's missing <Toaster/> component and <ActivityConfetti /> should solve the problem with conffetti and toaster in webpub export.

packages/pipeline/src/packaging/webpub.ts
```diff
@@ -0,0 +1,356 @@
...
+    title,
+    language: availableLanguages.length > 0 ? availableLanguages : [options.language],
+    modified: new Date().toISOString(),
+    // Fixed-layout books must be declared fixed (like the EPUB OPF's
+    // rendition:layout). Otherwise a reader treats the absolutely-positioned
+    // page as reflowable and it collapses to 0 height.
+    presentation: options.fixedLayout
+      ? { layout: "fixed", fit: "contain", spread: "none" }
+      : { overflow: "scrolled", spread: "none" },
+  }
```
zignaggo: presentation.layout is deprecated the new one is metadata.layout:
https://readium.org/webpub-manifest/contexts/default/#layout-and-reading-progression

zignaggo (changes requested):

PR #506, opened 2026-06-12
Feature: custom fonts

zignaggo:


elasticsounds:
Findings

#1 Uploaded body font emits a broken Google Fonts <link> on every page (functional bug)
packages/pipeline/src/package-web.ts:1239

This PR’s new resolveReflowableFontChain path returns bookFontFamilyChain(bodyFont) for any body-role book font, including uploaded non-Google fonts, and that flows into opts.bodyFontFamily. The block here then pushes the primary family into googleFamilies, which googleFontsCss2Url turns into a CDN URL with no catalog validation (confirmed — it blindly builds family=... from whatever it’s given). Result: assigning an uploaded font the body role on a reflowable book makes every page request a Google font that 404s, and re-adds a network dependency to the offline bundle for a font already bundled locally. Fix: gate the push to real Google families.

#2 Dead query invalidation — config cache stays stale after applying a font (functional bug)
apps/studio/src/hooks/use-book-fonts.ts:80

Invalidates ["books", label, "config"], but no query uses that key. The real keys are ["book-config", label] and ["debug", "config", label]. Since /fonts/apply writes reflowable_font into the book config, applying/resetting a font leaves the current-font card and rendering settings showing the old value until an unrelated refetch.

#3 UTF-8 BOM at the start of fonts.ts (minor)
apps/api/src/routes/fonts.ts:1 — confirmed EF BB BF prefix; only file in the diff with one. Doesn’t break the build, but an unintended artifact worth stripping.

#4 Unused import GOOGLE_FONTS (minor)
packages/types/src/book-fonts.ts:2 — imported but never referenced; CI ESLint no-unused-vars will flag it.

The rest of the feature is solid: path traversal is guarded in every filesystem-touching route, font uploads are validated by magic-byte parsing with a size cap, the new pages SQL is parameterized, font-name CSS injection is escaped, and all new user-visible strings are wrapped in Lingui macros.

PR #514, opened 2026-06-22
feat: split a book into page-range parts and merge them back (#503)

gbergengruen:
## Summary

Implements the split-into-parts / merge workflow from discussion #503: a **coordinator** can split a large book into page-range **parts**, hand each part to a contributor (or process them sequentially to cap RAM/compute), and **merge** the completed parts back into the full book — without ever overwriting data.

Closes the design questions raised in **https://github.com/unicef/adt-studio/discussions/503**.

## The contract

Page ids are deterministic and absolute (`pg051`, or `pg051052` in spread mode), so per-page `node_data` is portable across books with **no id remapping** — *as long as the PDF bytes and extraction config match*. That guarantee is encoded as a **pipeline fingerprint** with two tiers:

| Tier | Inputs | Merge behavior |
|------|--------|----------------|
| **identity** (`identityHash`) | PDF sha-256, DB schema version, extraction-affecting config (`spread_mode`, `vector_text_grouping`, `layout_type`) | **hard-fail** on mismatch |
| **semantics** (`semanticsHash`) | per-page prompt/model subset + editing language | **warn**, coordinator can override |

`start_page` / `end_page` are deliberately excluded — that exclusion is the whole point.

## Workflow

1. **Export a part** → lightweight archive: full PDF + windowed `config.yaml` + `part.json` manifest (no DB/images — the contributor re-extracts locally, which reproduces identical page ids).
2. **Import a part** → comes in as a normal windowed book (the existing import auto-detects part archives); run the per-page stages.
3. **Return trip** → contributor exports their book as a normal **project** zip (no new code needed).
4. **Merge** → preview shows pages added/replaced + node coverage + the identity/semantics verdict; applying copies per-page `node_data` as **new versions** (rollback-safe), upserts pages/images, carries `.cache` + `audio` so the book-level re-run is cheap, and clears book-level `step_runs` to mark them stale.

## Book-level stages

Rather than bespoke stitching, book-level stages (glossary, quiz, ToC, summary, easy-read, TTS, packaging) are **regenerated on the assembled book**, but the merge carries each part's content-addressed cache + audio so re-running is nearly free. This is both simpler and more correct — it picks up glossary-derived catalog entries that pure stitching would miss, and produces a true whole-book summary.

## What's in here

- **`@adt/types`** — `computeFingerprint` / `canonicalJson` / `PartManifest`.
- **api** — `part-service` (export / import / preview-merge / merge / part-info), shared `zip-util` (refactored out of `export-service`), and routes.
- **studio** — a "Split & merge" panel on the book overview (range picker + download; upload → preview → merge with acknowledge), an "imported part" badge, client methods, and hooks. Fully localized (en/es/fr/pt-BR/sq).

## Testing

- Unit: fingerprint tiers (range-exclusion, spread-flip, prompt-only drift, key-order).
- Service: merge round-trip, overlap → new version, identity hard-fail, semantics acknowledgement, lightweight part import.
- `pnpm typecheck` clean; studio i18n lint clean; all locales fully translated.

## Open item for review

After merge, book-level `step_runs` are cleared but their `node_data` versions are kept (for rollback). This should make those stages show as "needs run" and re-run via the normal cache-fast path. If reviewers know the runner skips a step when prior `node_data` exists despite a cleared `step_run`, I'll add a `clearNodesByType` for the stale book-level outputs.


gbergengruen:
Pushed a fix for a spread-mode bug found in review: `computeEqualWindows` assumed spreads pair as (1,2),(3,4)… but the extractor treats page 1 as a standalone cover and pairs the rest as (2,3),(4,5)… (matching the backend's `snapSpreadRange`). The mismatch caused suggested equal-part windows to be snapped into **overlapping** ranges at export, the displayed range to differ from what was exported, and the plan badges to never check off in spread mode.

The split is now built from indivisible spread units so suggestions are already spread-aligned and `snapSpreadRange` is a no-op on them. Marking ready for review.

PR #542, opened 2026-07-03
Revert "feat: add activity generation + layout mirror agents (#428)"

Eliezir:
#428 was merged directly to `main` before `develop` became the default branch. This reverts it from `main`.

The feature is being re-landed through the normal flow via #541 (`elasticsounds/port-layout-codex-agents` → `develop`).

No "revert-the-revert" trap: #541 delivers the feature to `develop` as fresh commits, so a later `develop` → `main` release brings it back cleanly, unrelated to this revert.


Eliezir:
Superseded — removing #428 from main via history rewrite (reset to release point b2e084f3a) instead of a revert commit. Feature re-lands via #541.

PR #544, opened 2026-07-04
Consistent web rendering: type scale, layout, and book-styled quizzes

elasticsounds:
Text size in web rendering is now driven by a book-level, editable typography map instead of per-page LLM guesses or raw PDF sizes: semantic `.adt-*` classes resolve to accessible, fluid (mobile→desktop `clamp()`) sizes injected as CSS appended unlayered after Tailwind, so they override any `text-*` the model emits and headings/body stay consistent across every page. A new **Fonts-tab "Text sizes" editor** — backed by `GET`/`PUT /books/:label/typography` — lets users tune each role using the style editor's Tailwind-token control (dropdown + typeable px), with defaults aligned to Tailwind's scale (body 24/16). The generation and styleguide prompts now instruct the model to apply the classes and never set sizes; the font size detected from the PDF is kept only as a seed/hint. Includes API + pipeline unit tests and full i18n across en/es/fr/pt-BR/sq.


apps/studio/src/components/pipeline/stages/storyboard/TypographySettings.tsx
```diff
@@ -0,0 +1,174 @@
...
+        </div>
+        <div className="text-[10px] font-semibold uppercase tracking-wide text-muted-foreground">
+          <Trans>Mobile</Trans>
+        </div>
+        <div className="text-[10px] font-semibold uppercase tracking-wide text-muted-foreground">
+          <Trans>Desktop</Trans>
+        </div>
+        <div className="text-[10px] font-semibold uppercase tracking-wide text-muted-foreground">
+          <Trans>Preview</Trans>
+        </div>
```
zignaggo, line 114: Missing tablet breakpoint on fonts sizes:

Image

zignaggo (changes requested):
The new font-size feature broken current font sizes.

https://github.com/user-attachments/assets/f7e700e2-df34-4a49-bc22-f0a8139956e4

ksokolovic:
Builds on @zignaggo's existing "broke font sizes" review — see above.

### 🔴 Blocking

**1. Cross-book state leak → silent data corruption in the Text-sizes editor.**
`TypographySettings.tsx:48-50`:
```ts
useEffect(() => {
  if (data?.data.styles && styles.length === 0) setStyles(data.data.styles)
}, [data, styles.length])
```
`styles` starts `[]` and is assigned **only while `styles.length === 0`**, so it populates once per component lifetime. The component isn't keyed by book (`FontSettings.tsx:240` renders `<TypographySettings bookLabel={bookLabel} />` with no `key`), and TanStack Router doesn't remount on a `$label` change.
*Failure scenario:* open Fonts → Text sizes for book A (grid populates with A's sizes) → navigate to book B without a full reload. `useTypography(B)` refetches B, but the guard no longer fires, so the grid still shows **A's** values; `dirty` (`:52-55`) compares A's `styles` to B's `data` → true → Save enables → `handleSave` → `update.mutate` targets `useUpdateTypography(B)` → **A's sizes are persisted onto book B**. Fix: reload when the book identity changes (depend on `bookLabel` / reset on `data` identity change) or `key={bookLabel}` the subtree.

**2. Every unedited book jumps to the large default type scale — the regression @zignaggo flagged.**
`DEFAULT_TYPOGRAPHY` (`packages/types/src/typography.ts`) sets body 24px desktop / 16px mobile (h1 48, etc.), and `readTypography()` returns it whenever a book has no saved typography node. The **detected** per-book scale (`type-scale.ts`, body ≈ the PDF's modal size, usually ~16px) is deliberately demoted to a "hint/seed" and never seeds `readTypography` or the editor's empty state; "Reset to defaults" (`TypographySettings.tsx:~155`) also resets to the large values. So every already-processed book immediately renders body at fluid 24px (unlayered `.adt-body` wins over prior `text-*`), which is exactly the "broke current font sizes" regression. The large accessible default is a reasonable *goal*, but there's no path that preserves a book's prior/detected sizing. Recommend seeding `readTypography`/the editor empty-state from the detected `TypeScale` (or offering "use detected sizes"), and reconcile against @zignaggo's video before merge.

### 🟠 Should-fix

**3. Typography CSS pins `font-size` but not `line-height`, and the preservation guard doesn't catch `text-*` utilities.**
`buildTypographyCss` (`typography.ts:48-53`) emits only `font-size` (+ optional weight). `typographyPreservationErrors` (`typography.ts:81-100`) rejects *removed* `adt-*` classes and *added inline* `font-size:`, but **not** added Tailwind `text-*` size utilities. The generation/visual-review prompts tell the model not to emit `text-*` sizes, but LLMs disobey — that's what the guard is for. When a residual `text-*` survives (e.g. `class="adt-body text-base"`), the unlayered rule bumps font-size to 24px while `line-height` stays the utility's ~24px → ratio ~1.0 → cramped/overlapping text. This compounds [2]. Fix: (a) also reject added `text-(xs…9xl)`/`text-[..]` in `typographyPreservationErrors`, and (b) pin a `line-height` (ratio) on the `adt-*` classes so leading tracks the pinned size.

**4. CSS-injection invariant lives only at the PUT route, not in the schema or the CSS generator.**
`className`/px/`fontWeight` are unconstrained in `packages/types/src/typography.ts` (`z.string()`/`z.number()`); only `typography.ts` route's `CLASS_RE`/`MIN_PX`/`MAX_PX` guard them, and `buildTypographyCss` interpolates `className` straight into a selector. Any *other* writer of the `typography/book` node (import/seed/migration/second endpoint) bypasses the guard and emits attacker-controlled CSS into every packaged/preview stylesheet — and packaged bundles are shareable. Move the `className` pattern (`^adt-[a-z0-9-]+$`) and px/weight ranges into the `TypographyStyle` Zod schema so validation is centralized; the route just `safeParse`s.

**5. Quiz option/selected text isn't contrast-checked (matters for an a11y product).**
`quiz-palette.ts` contrast-picks `headerText`/`submitText` via WCAG luminance (good), but option text uses the book's `ink` on a pale-accent `optionFill`, and selected-option text is `accent` on `accentSoft` — neither is verified for AA. On some book palettes these can fall below 4.5:1. Consider running the same `readableText`/contrast check for option and selected states.

**6. Malformed/empty PUT body returns 500, not 400.** `typography.ts` route does `await c.req.json()` unguarded; bad JSON throws a non-`HTTPException` → global 500. Use `.catch(() => null)` + null-check → 400, per the `fonts.ts` convention.

**7. Tests miss the versioning core.** `typography.test.ts` covers GET defaults/404 and PUT happy-path + px/className rejection, but never asserts append-only versioning (no `version === 1` then second PUT `=== 2`, no latest-after-resave), and doesn't cover the empty-styles or Zod-shape-failure branches. An accidental switch to overwrite semantics would pass all tests.

### 🟡 Polish
- `fontWeight` is written into CSS (`font-weight: ${...}`) with no range check, unlike the px fields right beside it.
- GET can return `isDefault:false` + saved `version` while serving `DEFAULT_TYPOGRAPHY` if a persisted row fails `safeParse`/is empty (three fields from two reads can disagree). Latent.
- PUT is last-write-wins; the editor ignores the returned `version` (no concurrency check).
- No duplicate `className`/`key` detection (last CSS rule wins).

ksokolovic (dismissed):
Full review in my comment above — #1 (cross-book state leak in the Text-sizes editor: saving on book B can persist book A's sizes) and #2 (unedited books jump to the large default type scale — the regression @zignaggo flagged) are blocking; the line-height / `text-*` guard gap (#3) should ship with them. The rest is should-fix/polish, and the branch needs a rebase on develop.

apps/studio/src/components/pipeline/stages/storyboard/TypographySettings.tsx
elasticsounds: The font-size regression this built on is resolved in `398df11b`: a book with no saved typography now defaults to its **detected per-book sizes** (floored at 16px), not the large scale — so pages stay close to the original. The large scale is now an opt-in **"Accessible sizes"** preset in the Text-sizes editor.

On the tablet breakpoint specifically: the scale renders as a fluid `clamp(mobile, …, desktop)`, so tablet width gets a smoothly-interpolated size (no hard jump, no missing stop) rather than a separate discrete value. If you'd rather have an explicit **Tablet** control (a third editable size), I can add `tabletPx` to the schema + a 3-stop clamp — happy to wire it if that's preferred.

elasticsounds:
Thanks for the very thorough pass — all blocking + should-fix items are addressed across `00516aa4` and `398df11b`. Verified with typecheck, 776 pipeline + 372 API tests, lint, and 0 missing i18n.

### 🔴 Blocking
1. **Cross-book state leak** — Fixed. `TypographySettings` is now `key={bookLabel}` in `FontSettings.tsx`, so it remounts on book change and can no longer carry book A's `styles` into a Save targeting book B.
2. **Large default = regression** — Fixed by changing the default itself. A book with no saved typography now **seeds from its detected per-book scale** (the PDF's own sizes) rather than the large `DEFAULT_TYPOGRAPHY`, so already-processed books render close to their original. Seeded sizes are **floored at 16px** (a11y minimum) and flat (mobile == desktop, since the PDF isn't responsive). The large 24/16 scale is now an opt-in **"Accessible sizes"** preset in the editor, alongside **"Reset to detected"** (GET now also returns the detected scale). This keeps the "consistent, globally-set" goal while defaulting conservatively.

### 🟠 Should-fix
3. **`text-*` guard + `line-height`** — Both done. `typographyPreservationErrors` now also rejects newly-added `text-*`/`text-[..]` size utilities (gated so it only fires on sections that actually use `adt-*`), and `buildTypographyCss` pins a unitless `line-height` per role (1.2 headings / 1.55 body / 1.4 caption) so leading tracks the pinned size.
4. **Validation centralized in the schema** — `className` (`^adt-[a-z0-9-]+$`), px (8–200), `fontWeight` (100–900), and non-empty `styles` now live on the `TypographyStyle`/`BookTypography` Zod schema; the route just `safeParse`s. Any writer of the node is guarded, not just the PUT route.
5. **Quiz option/selected contrast** — Now WCAG-contrast-checked with a readable fallback, same approach as `headerText`/`submitText`.
6. **Malformed PUT → 500** — Fixed: `c.req.json().catch(() => null)` → 400.
7. **Versioning tests** — Added append-only versioning (v1 → v2 → latest-wins), plus empty-styles, malformed-JSON, and `text-*`-guard branches.

### 🟡 Polish
- `fontWeight` range: now schema-validated (100–900) as part of [4].
- Not changed yet (happy to follow up if you'd like): the GET `isDefault`/`version` edge when a persisted row fails `safeParse`, last-write-wins concurrency (editor ignores returned `version`), and duplicate-`className` detection. The editor emits a fixed, valid set so these are latent; call it and I'll harden them.

zignaggo (approved):

PR #553, opened 2026-07-09
Feature: Add model-specific prompt management and global prompt settings

zignaggo:
This PR adds model-specific prompt support across the pipeline and introduces a global prompt settings screen for managing fallback and model-specific prompt files.

### Changes include:

- Model-specific prompt resolution with openai:gpt-5.4 as the default fallback model.
- Global prompt settings page with a model-grouped prompt tree.
- Monaco-based prompt editing with Liquid guide support.
- Floating save/discard/reset behavior for prompt editing screens.
- Prompt fallback indicators and default model badges.
- Tools to create model prompt files from templates and manage non-default model prompts.
- Prompt version retention limit to avoid unbounded version file growth.
- Skeleton loading states for prompt file/editor screens.
- Updated i18n catalogs for all supported locales.

Validation:

- Confirm global prompt files can be edited from the Prompt Settings screen, saved as global versions, reset back to shipped fallback files, and reused by books through the fallback prompt resolution path.
- Confirm book-level prompts can be edited directly from pipeline settings screens, saved for the current book, and used independently from the global prompt defaults.


elasticsounds:
Hi Zig! 

This is looking good! For the global prompts section - adding all the prompt files greatly increases the size of the application test surface area - especially if we need to fundamentally change a part of prompt, tune the prompts for other performance related issues. 

I really like the feature where you can generate model specific prompts based on the model selected. 

My thoughts are it might make more sense to not include these prompts in the repo, but instead ship with the ability to just generate the prompts if that makes sense? What do you think @Eliezir

Eliezir:
> Hi Zig!
> 
> This is looking good! For the global prompts section - adding all the prompt files greatly increases the size of the application test surface area - especially if we need to fundamentally change a part of prompt, tune the prompts for other performance related issues.
> 
> I really like the feature where you can generate model specific prompts based on the model selected.
> 
> My thoughts are it might make more sense to not include these prompts in the repo, but instead ship with the ability to just generate the prompts if that makes sense? What do you think @Eliezir

Agreed, as new models drop we’d keep accumulating prompt files, so bundling them all into the repo doesn’t scale. I really like the idea of generating the prompts on demand.

we would ship just the default prompts and for the future, we could work on some sort of registry/store where we publish prompts for different models/scenarios so the user can still have access to tested prompts without cluttering the repo

PR #557, opened 2026-07-10
Multi-Provider Selection & Model Resolution

Mominadar:
## Problem

When multiple LLM providers are configured (e.g., OpenAI + Anthropic), the system always silently picks the first one in a hardcoded priority order (OpenAI → Anthropic → Google → Custom). Users have no way to control which provider is used for pipeline runs, no way to change the default model per provider, and no visibility into which provider was actually used.

Worse, the app previously required an OpenAI key to function — even if the user only wants to use Anthropic or Google, the pipeline would fail without a valid OpenAI key configured. There was no workaround short of providing an OpenAI key.

Additionally, when the configured model's provider has no key available, the pipeline fails silently rather than falling back to an available provider.

## Solution

**User-facing: Default provider selection in Settings**
- Added a "Mark as default" button next to each provider's API key label in the Settings dialog
- Shows a green "default" badge and ✦ tab indicator for the active default
- Choice persists across sessions (localStorage)
- Only visible when 2+ providers are configured — single-provider setups are unchanged
- If no provider is explicitly selected, the first available configured provider is used automatically instead of failing

**User-facing: Per-provider default model configuration**
- Each provider tab in Settings now includes a "Default AI model" dropdown
- Users can change the model used for each provider (e.g., switch OpenAI from gpt-5.4 to gpt-4.1, or Anthropic from claude-sonnet-4-6 to claude-opus-4-6)
- The selected model is persisted per provider and used whenever that provider is active
- System provides sensible fallbacks if no model is explicitly chosen (e.g., openai:gpt-5.4, anthropic:claude-sonnet-4-6)

**Backend: Model resolution engine**
- New `resolveEffectiveModelId()` function in `packages/llm` that routes models to available providers
- Respects per-step model overrides (`X-Step-Model-Overrides` header)
- Respects the user's default model preference (`X-Default-LLM-Model` header)
- Falls back gracefully when a provider's key is missing — automatically resolves to the first available provider instead of failing
- Handles Anthropic's lack of JSON-mode by automatically switching to tool-mode
- Removed the hard dependency on OpenAI — any single configured provider is now sufficient to run the full pipeline

**Frontend: Unified model source of truth**
- `useApiKey().defaultModel` now respects the active provider choice everywhere (workflow pages, step configs, stage runs)
- `useActiveProvider` hook encapsulates resolution, persistence, and reactivity
- `StageRunCard` shows which provider was used for each run
- Removed the OpenAI-specific key validation that previously blocked non-OpenAI-only setups

**Speech page: Better Anthropic messaging**
- Added visible info banner explaining that Anthropic doesn't offer TTS when no speech-capable provider is configured

### Key Points

- **No breaking changes** — single-provider setups behave identically to before
- **OpenAI is no longer required** — the app now works with any single configured provider (Anthropic, Google, or Custom alone)
- **Graceful fallback** — if no default provider is explicitly selected, the first available configured provider is used instead of failing
- **Per-provider model customization** — users can now choose which specific model each provider uses without editing config files
- **Frontend-only provider selection** — the backend already accepted all provider keys; this just controls which model ID is sent as the default
- **Backend model resolution is new** — enables automatic fallback when a step's configured model isn't available
- **Anthropic JSON→tool mode fix** — the AI SDK's Anthropic provider doesn't support `mode: "json"`, so we automatically switch to `mode: "tool"` which provides equivalent schema-constrained output
- **i18n complete** — all new strings translated to pt-BR, es, fr, sq

### Testing

- All 1661 tests pass (2 pre-existing timeouts in `part-service.test.ts` unrelated to this PR)
- TypeScript strict mode passes clean
- New test files: `model-resolution.test.ts`, `use-step-config.test.tsx`, `StageSidebar.test.tsx`
- Manual testing: verified provider switching, persistence across sessions, model display on workflow pages, stage run execution with non-OpenAI providers

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@ Thumbs.db
 
 # AI agent local config
 .claude/
+.kiro/
 
 # AI agent rule files
 CLAUDE.md
--- a/apps/api/src/routes/glossary.ts
+++ b/apps/api/src/routes/glossary.ts
@@ -122,9 +122,13 @@ export function createGlossaryRoutes(
     const { label } = c.req.param()
     const safeLabel = safeParseLabel(label)
 
-    const apiKey = c.req.header("X-OpenAI-Key")
-    if (!apiKey) {
-      throw new HTTPException(400, { message: "Missing X-OpenAI-Key header" })
+    const apiKey = c.req.header("X-OpenAI-Key") || undefined
+    const anthropicApiKey = c.req.header("X-Anthropic-API-Key") || undefined
+    const googleApiKey = c.req.header("X-Google-API-Key") || undefined
+    const customBaseUrl = c.req.header("X-Custom-Base-URL") || undefined
+    const customApiKey = c.req.header("X-Custom-API-Key") || undefined
+    if (!apiKey && !anthropicApiKey && !googleApiKey && !customBaseUrl) {
+      throw new HTTPException(400, { message: "At least one LLM provider API key is required." })
     }
 
     const body = await c.req.json()
@@ -157,6 +161,10 @@ export function createGlossaryRoutes(
         onLog: (entry) => storage.appendLlmLog(entry),
         credentials: {
           openaiApiKey: apiKey,
+          anthropicApiKey,
+          googleApiKey,
+          customBaseUrl,
+          customApiKey,
         },
       })
 
--- a/apps/api/src/routes/pages.ts
+++ b/apps/api/src/routes/pages.ts
@@ -1151,10 +1151,13 @@ export function createPageRoutes(
     }
     const { sectionIndex } = queryParsed.data
 
-    const apiKey = c.req.header("X-OpenAI-Key")
-    if (!apiKey) {
+    const apiKey = c.req.header("X-OpenAI-Key") || undefined
+    const anthropicApiKey = c.req.header("X-Anthropic-API-Key") || undefined
+    const googleApiKey = c.req.header("X-Google-API-Key") || undefined
+    const customBaseUrl = c.req.header("X-Custom-Base-URL") || undefined
+    if (!apiKey && !anthropicApiKey && !googleApiKey && !customBaseUrl) {
       throw new HTTPException(400, {
-        message: "Missing X-OpenAI-Key header",
+        message: "At least one LLM provider API key is required.",
       })
     }
 
@@ -1218,7 +1221,11 @@ export function createPageRoutes(
             promptsDir,
             webAssetsDir,
             configPath,
-            apiKey,
+            apiKey: apiKey ?? "",
+            anthropicApiKey,
+            googleApiKey,
+            customBaseUrl,
+            defaultModelId: c.req.header("X-Default-LLM-Model") || undefined,
           })
         },
         { pageId, url: `/books/${safeLabel}/storyboard/${pageId}` }
@@ -1236,7 +1243,11 @@ export function createPageRoutes(
       promptsDir,
       webAssetsDir,
       configPath,
-      apiKey,
+      apiKey: apiKey ?? "",
+      anthropicApiKey,
+      googleApiKey,
+      customBaseUrl,
+      defaultModelId: c.req.header("X-Default-LLM-Model") || undefined,
     })
 
     return c.json(result)
@@ -1252,9 +1263,12 @@ export function createPageRoutes(
       throw new HTTPException(400, { message: "Invalid section index" })
     }
 
-    const apiKey = c.req.header("X-OpenAI-Key")
-    if (!apiKey) {
-      throw new HTTPException(400, { message: "Missing X-OpenAI-Key header" })
+    const apiKey = c.req.header("X-OpenAI-Key") || undefined
+    const anthropicApiKey = c.req.header("X-Anthropic-API-Key") || undefined
+    const googleApiKey = c.req.header("X-Google-API-Key") || undefined
+    const customBaseUrl = c.req.header("X-Custom-Base-URL") || undefined
+    if (!apiKey && !anthropicApiKey && !googleApiKey && !customBaseUrl) {
+      throw new HTTPException(400, { message: "At least one LLM provider API key is required." })
     }
 
     const body = await c.req.json()
@@ -1281,7 +1295,11 @@ export function createPageRoutes(
             promptsDir,
             webAssetsDir,
             configPath,
-            apiKey,
+            apiKey: apiKey ?? "",
+            anthropicApiKey,
+            googleApiKey,
+            customBaseUrl,
+            defaultModelId: c.req.header("X-Default-LLM-Model") || undefined,
           })
 
           // Save the edited HTML as a new rendering version
@@ -1319,7 +1337,11 @@ export function createPageRoutes(
       promptsDir,
       webAssetsDir,
       configPath,
-      apiKey,
+      apiKey: apiKey ?? "",
+      anthropicApiKey,
+      googleApiKey,
+      customBaseUrl,
+      defaultModelId: c.req.header("X-Default-LLM-Model") || undefined,
     })
 
     return c.json(result)
@@ -1947,9 +1969,12 @@ export function createPageRoutes(
         return c.json({ error: `Book not found: ${safeLabel}` }, 404)
       }
 
-      const apiKey = c.req.header("X-OpenAI-Key")
-      if (!apiKey) {
-        return c.json({ error: "Missing X-OpenAI-Key header" }, 400)
+      const apiKey = c.req.header("X-OpenAI-Key") || undefined
+      const anthropicApiKey = c.req.header("X-Anthropic-API-Key") || undefined
+      const googleApiKey = c.req.header("X-Google-API-Key") || undefined
+      const customBaseUrl = c.req.header("X-Custom-Base-URL") || undefined
+      if (!apiKey && !anthropicApiKey && !googleApiKey && !customBaseUrl) {
+        return c.json({ error: "At least one LLM provider API key is required." }, 400)
       }
 
       const pageId = c.req.query("pageId")
@@ -2005,7 +2030,7 @@ export function createPageRoutes(
           desc,
           async () => {
             return await executeAiImageGeneration({
-              safeLabel, bookDir, dbPath, apiKey, pageId,
+              safeLabel, bookDir, dbPath, apiKey: apiKey ?? "", pageId,
               prompt, referenceImageId, targetImageId,
               style, imageType, styleImageId, promptsDir,
               sectionIndex, mode, booksDir,
@@ -2018,7 +2043,7 @@ export function createPageRoutes(
 
       // Fallback: run synchronously
       const result = await executeAiImageGeneration({
-        safeLabel, bookDir, dbPath, apiKey, pageId,
+        safeLabel, bookDir, dbPath, apiKey: apiKey ?? "", pageId,
         prompt, referenceImageId, targetImageId,
         style, imageType, styleImageId, promptsDir,
         sectionIndex, mode, booksDir,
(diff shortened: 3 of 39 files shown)


ksokolovic (changes requested):
Requesting changes per my detailed review in [this comment](https://github.com/unicef/adt-studio/pull/557#issuecomment-5022677519) — the three blocking items are all in the model-resolution contract, and each fails silently rather than erroring.

ksokolovic:
This is an ambitious PR and the shape of it is right — a single `resolveEffectiveModelId` in `packages/llm`, provider credentials already flowing through `buildLLMCredentials`, and the fallback logic living behind one function rather than scattered through callers. A few things stand out as genuinely well done: **all five locales are updated with real translations** (`sq` included — `Shëno si parazgjedhje`, not copied English), there are new unit tests for the resolution logic, there are no dynamic Tailwind class names, and `LandingPageShell.tsx:155-161` correctly pairs `opacity-0` with both `pointer-events-none` and `inert`.

The blocking issues below are all in the same area: the model-resolution contract between the frontend and `resolveEffectiveModelId`.

### 🔴 Blocking

**1. `explicitModelIds` is matched by model id rather than by step, so one override leaks onto every step sharing that model.**

`client.ts:47-49` decides whether a step is "explicitly overridden" with `!credentials.explicitModelIds?.includes(modelId)`, and `use-book-run.ts:26-44` builds that list from `record.model` values in the **book config** — model ids, not step names. A model id is not unique to a step.

That matters here because all 14 `model:` entries in `config.yaml` are the same string, `openai:gpt-5.4`, and a default OpenAI user's `defaultModel` is also `openai:gpt-5.4` (`use-api-key.ts:103`). So the colliding value is the common one, not an exotic case. Note too that `onModelChange` (`use-step-config.ts:70-75`) sets `model_override: true` with no equality check — re-picking the same model in a dropdown is enough to create the override.

*Failure scenario:* a user has one step overridden with the value `openai:gpt-5.4` — either because they re-selected it in a dropdown, or automatically via issue 2 below. `explicitModelIds` becomes `["openai:gpt-5.4"]`. They then set their OpenAI default model to `gpt-4.1` in Settings. Every step still on `openai:gpt-5.4` matches the list, is treated as explicitly overridden, and keeps using `gpt-5.4`. The default-model setting silently does nothing, with no error.

Issues 1 and 2 compound: issue 2 manufactures exactly the `{ model: "openai:gpt-5.4", model_override: true }` entry that triggers this.

*Fix:* key the exclusion by step rather than by model value — e.g. send `{ stepName: modelId }` and resolve per step — so two steps that happen to share a model id are not conflated.

**2. A saved per-step override is silently rewritten to the default model and then persisted.**

`use-step-config.ts:43` sets `modelOverride` from config, but `:53-57` sets `model` to `defaultModel` whenever the configured model's provider isn't currently available. The two states then disagree, and `configOverrides` (`:79-84`) emits `{ model: <defaultModel>, model_override: true }`.

*Failure scenario:* `config.yaml` has `glossary: { model: "anthropic:claude-opus-4-6", model_override: true }`. The user removes their Anthropic key. They then save any unrelated glossary setting. Their Anthropic override is overwritten with the OpenAI default — still flagged as an intentional override. Re-adding the Anthropic key does not bring it back. The user never touched the model field.

That fabricated override then feeds `collectExplicitModelIds`, so it is also shipped to the backend as a real explicit override.

*Fix:* don't rewrite `model` when the provider is unavailable — keep the configured value and surface unavailability in the UI instead. The new test only covers the available-provider case.

**3. Custom-only setups always send a placeholder model id.**

`use-api-key.ts:106` and `:135` fall back to the literal `"custom:your-model-name"`, but `ApiKeyDialog.tsx:499` gates the model dropdown behind `tab !== "azure" && tab !== "custom"` — so `customDefaultModel` can never be set through the UI.

*Failure scenario:* a user configures only a custom base URL. `defaultModel` resolves to `"custom:your-model-name"`, which is sent as `X-Default-LLM-Model` and used as the fallback in `resolveEffectiveModelId` (`FALLBACK_MODELS.custom`). Every run fails on an unknown model, and the same placeholder string is displayed in `ProviderSelector`.

*Fix:* either expose the model input for the custom tab, or require the model as part of custom-provider configuration.

### 🟠 Should fix

**4. "OpenAI is no longer required" holds for stage runs, but not for the on-demand AI actions.** `stages.ts` was relaxed; these four still hard-require `X-OpenAI-Key` and were not touched:

- `POST /books/:label/pages/:pageId/re-render` (`pages.ts:1154-1157`)
- `POST /books/:label/pages/:pageId/sections/:sectionIndex/ai-edit` (`pages.ts:1255-1257`)
- `POST /books/:label/quizzes/generate-one` (`quizzes.ts:135-137`)
- `POST /books/:label/glossary/generate-one` (`glossary.ts:125-127`)

All four are text-LLM operations that any provider could serve. So an Anthropic-only user can run the whole pipeline but cannot re-render a page, use AI edit, or regenerate a single quiz or glossary item. Not a regression — they behaved this way before — but it does mean the "works with any single configured provider" claim needs either extending or narrowing.

*Correctly left alone:* `tts.ts:1111,1216` (Whisper) and `POST /books/:label/images/ai-generate` (`pages.ts:1950`) are genuinely OpenAI-bound — `IMAGE_MODEL_GROUPS` (`ModelSelect.tsx:266-274`) offers only OpenAI image models.

**5. Question rather than a defect: was the Anthropic tool-mode switch verified against a recursive schema?** `resolveObjectMode` (`client.ts:370-380`) rewrites `mode: "json"` → `"tool"` for Anthropic, and `client.test.ts:145` does test that rewrite. But that test mocks `generateObject` and uses a flat schema (`z.object({ ok: z.boolean() })`), so it verifies the mode string is forwarded, not that tool mode accepts the payload. The only caller requesting json mode is `page-sectioning.ts:137,179`, which uses the **recursive** `ContentNodeData` schema (`z.lazy`) — precisely what the code comment claims tool mode handles. If that assumption is wrong, sectioning fails for Anthropic users at stage two. Was this exercised against a live Anthropic call?

**6. `handleRun` swallows failures and can double-fire.** `SpeechLandingPage.tsx:103-120` is `async`, awaits `updateConfig.mutateAsync` with no `try`/`catch`, and is passed straight to `onRun`. If the config write rejects it's an unhandled rejection — the run never starts and the user gets no feedback at all. The guard checks `status.isRunning` but not `updateConfig.isPending`, and `isRunning` is still false during the await, so a double-click fires two config writes and two `queueRun` calls.

**7. Speech provider selection reverts under refetch.** The new `else { setProvider(firstAvailableSpeechProvider) }` branch at `SpeechLandingPage.tsx:80-82` runs whenever `speech.default_provider` isn't set. Since the persist is fire-and-forget, any `activeConfigData` refetch landing before the write is reflected snaps the selection back to OpenAI under the user's cursor. Before this PR the `else` branch didn't exist, so the stale read was harmless.

**8. The provider selector is invisible but still focusable when the rail is collapsed.** `StageSidebar.tsx:181-184` uses `opacity-0` with no `pointer-events-none` and no `inert`, so a keyboard user tabs into an invisible combobox and can open an invisible listbox. `LandingPageShell.tsx:155-161` in this same PR already does this correctly — same treatment here would fix it.

**9. An unsaved `max_retries` edit is wiped.** `use-step-config.ts:37` calls `setRetries("")` with no `touchedRef` guard, while `model` and `modelOverride` are both guarded. The PR added provider/model values to the effect's dependency array (`:66`), so it now re-runs on provider changes too: type a retry value, change the default provider, and the field resets while the section is still marked dirty — saving then persists the wrong value.

**10. The user's provider choice is destroyed rather than left inert.** `use-active-provider.ts:154-159` writes `setPersisted(resolvedType ?? "")` when the persisted provider is no longer configured. Rotating an Anthropic key (remove, save, re-add) silently discards the selection. `resolveActiveProvider` already ignores an unconfigured persisted value, so the write isn't needed for correctness — it only destroys recoverable state.

### 🟡 Accessibility

The new controls are reachable but not described. Worth a pass:

- **Provider tabs aren't tabs** — `ApiKeyDialog.tsx:197` is a plain `<div>` of `<button>`s; the file contains zero `role="tab"`, `role="tablist"` or `aria-selected`. Selection is conveyed by border/text colour only, and arrow-key navigation doesn't work.
- **"Mark as default" has no pressed state and destroys focus** — no `aria-pressed`, and the button is swapped for a static `<span>` badge on activation (`ApiKeyDialog.tsx:239-251`), so a keyboard user's focus lands back on `<body>`. One persistent button with `aria-pressed` would fix both.
- **The `✦` default indicator and the green configured-check have no text alternative** (`ApiKeyDialog.tsx:222,224`) — colour and glyph only.
- **`ProviderSelector`'s listbox has no keyboard navigation** (`ProviderSelector.tsx:134-174`) — only `Escape` is handled; no arrow keys, no `aria-activedescendant`, no focus move on open or restore on close.

### 🟡 Polish

- **`ModelSelect.tsx:184`** — `{open && filtered.length > 0 && ...}` unmounts the whole popup on a no-match query. No empty state. With zero providers configured every group is disabled at 40% opacity with only a 9px "Not configured" chip explaining it.
- **`PromptViewer.tsx:74-80`** — the disabled-provider mapping has no `gemini` case, so Gemini TTS models are never marked unconfigured when the Google key is absent.
- **`SpeechLandingPage.tsx:283-295`** — the Anthropic TTS banner is gated on "no speech provider key", not on Anthropic being configured, so a brand-new user with no keys is told about a limitation of a provider they haven't chosen.
- **`StageSidebar.tsx:193`** — `<Trans>This step uses</Trans> <span>{model}</span>` splits a sentence, so translators get a dangling verb phrase they can't reorder. Wrap the whole thing in one `<Trans>`.
- **`LandingPageShell.tsx:153`** — the +27/−25 on this file is a re-indent that wraps the run button in `<div className="flex items-center gap-3">` containing exactly one child. A single-child flex row with a gap does nothing; this reads like a provider selector was meant to sit beside the Run button. Either add it or revert the file to cut ~52 lines of diff noise.
- **`modelTouchedRef` is never reset after a save** (`use-step-config.ts:32`) — it only clears on `configKey` change, which never changes for a mounted settings page, so the field ignores server state for the component's lifetime after one edit.
- **Provider resolution is duplicated** — `use-api-key.ts:98-123` reimplements `resolveActiveProvider` (`use-active-provider.ts:69-88`) including a second hardcoded copy of the priority order at `:119`, and skips the validation the original does. Currently equivalent; will drift.
- **Unmemoized identities defeat the `useMemo` chain** — `use-api-key.ts:131-136` returns a fresh `providerDefaultModels` object each render, which is a dep of `use-active-provider.ts:170`, so `allProviders` → `configuredProviders` → `activeProvider` all get new identities every render, and `queueRun`'s `useCallback` never memoizes. Churn rather than a correctness bug, but it makes those memos dead code.
- **Bare model ids are assumed to be OpenAI** in both `client.ts:39` and `use-step-config.ts:46-47`. Fine today; worth a shared helper.

### ⚠️ Merge mechanics

Conflicting against `develop` in 10 files, including `packages/llm/src/client.ts`, `use-book-run.ts`, `StageSidebar.tsx`, `StageRunCard.tsx` and all five `.po` catalogs. For the catalogs, take develop's side and re-run `pnpm --filter @adt/studio extract` rather than hand-merging.

### Note on overlap with #470

Full disclosure since I pushed it: #470 now resolves the required credential from the judge model's provider in `apps/api/src/routes/translation-evaluations.ts`, which is conceptually the same job `resolveEffectiveModelId` does here. The two don't conflict and neither blocks the other, but whichever lands second is worth a consolidation pass — two independent notions of "which provider does this model id belong to" will drift apart.

elasticsounds:
@ksokolovic Is this something we will be merging eventually or superceded by other work?

ksokolovic:
After reviewing the branch against current `develop`, I do not think this PR should be merged or rebased in its present form. Much of its default-model/provider-selection architecture has been superseded by #567, which is now merged.

The remaining blocking issue is that per-step overrides still are not wired through production: `resolveEffectiveModelId` accepts `stepName`, but production callers do not pass it, and config keys such as `quiz_generation` do not match pipeline step names such as `quiz-generation`. The update also relaxes validation on some OpenAI-only operations while still invoking OpenAI with an empty key.

My recommendation is to close this PR as superseded while retaining it as implementation reference. If provider-neutral text operations should work without an OpenAI key, that valuable part should be submitted as a focused follow-up from current `develop`, using #567’s model configuration and selecting credentials based on the resolved model provider.

I am leaving the existing “Changes requested” review in place. Thanks again for the work here—the core multi-provider routing goal is still worth pursuing, but it will be safer and considerably smaller on the current architecture.

PR #567, opened 2026-07-14
Feature: change default model

zignaggo:
## Summary

Adds centralized application settings and support for changing the default pipeline model.

## Changes

- Adds a new `/settings` page with navigation for:
  - Default model
  - API keys
  - Global prompts
- Updates the header settings button to open the new page.
- Moves default-model configuration out of Global Prompts.
- Embeds API-key configuration in the settings page.
- Preserves model-specific prompt resolution when changing the default model.
- Redirects the previous prompt-settings route.
- Adds responsive, accessible navigation.
- Updates all supported translations.

## Validation

- Typecheck passed
- Studio production build passed
- 10 targeted tests passed
- Lint passed with no errors
- Translation catalogs have no missing entries


elasticsounds:
I love it - just a question as to why the default model is missing from the gear icon within the actual pipeline screen - can it be there also?
image

elasticsounds:
Default model is a different concept depending on the different tasks, so in reality gpt 5.4 is not the default model across all. For instance, gpt generation is gpt-image-1.5. For tts it's gpt-4o-min. In the "default" screen - can you list in text the tasks that take the default model 5.4, and then separate the tasks where the default model is something else - for instance speech generation and, image generation.

elasticsounds:
Hey Zig - this is really amazing work. It makes perfect sense to me and is a very powerful addition. One note is if you change the default model to say gemini 2.5 or to gpt 5.5, I think we should at that point warn the user that the current prompts are optimized for 5.4, we recommend they should generate new global prompt files for the model they have switched to. You can then direct them to the screen with the global prompts. 

Or they can just continue and try with the existing prompts.

elasticsounds:
For the options for images - if you could also add gpt-images-1.5. I will put a separate issue for you to wire in the nano banana model series.

For TTS please add the azure models and the gemini models as are already available on the tts screen. (ie. 2.5 flash and pro and 3.1)

PR #568, opened 2026-07-14
feat(kids): Kids Mode — buddy-guided reading + Studio authoring [draft]

Eliezir:
## Summary

Adds a complete Kids Mode reading experience and the Studio authoring tools needed to configure, preview, generate, and export it. The feature is designed for the UNICEF/Ceibal accessibility workflow and keeps the book language, Kids Mode interface, and generated voice assets aligned.

### Kid-facing runtime

- Buddy-guided reading chrome with five characters and seven expressions per character.
- Read-aloud controls, reading speed, sign language, easy read, word helper, story map, language selection, accessibility controls, resume flow, and end-of-book experience.
- Per-buddy OpenAI TTS with offline voice assets, silent text fallback, and a global book-independent cache.
- Full onboarding with buddy selection, player naming, localized prompts, and replay support.
- Kid-sized navigation with reduced-motion support, keyboard focus states, semantic status updates, and touch-friendly targets.

### Responsive interface

- A grouped list interface on tablet and desktop.
- A single bottom-sheet interface on mobile, with no layout selector or legacy menu variants.
- Full-width reading and reset actions, a fluid three-column action grid, centered mobile switches, and a compact speed control.
- Compact page arrows at the vertical midpoint and bottom reading controls sized to minimize content obstruction.
- Buddy feedback anchored above the buddy and playback controls instead of floating at the top of the page.
- The buddy panel is modal on mobile and hides unrelated floating controls while open.

### Language parity gate

- Kids Mode can only run when every language available in the book also has a complete localized Kids Mode interface.
- Preview and export use the same readiness rules, preventing a book-language and Kids-UI mismatch.
- The Studio preview now shows a themed, actionable warning when translations are incomplete, including a direct link to the Kids Mode configuration section.
- All supported Studio locales are updated: English, Portuguese (Brazil), Spanish, French, and Albanian (Kosovo).

### Studio authoring

- Kids Mode canvas and inspector with character art, localized lines, voice preset and prompt editing, save/reset controls, expression gallery, and generation/audition actions.
- Per-buddy, per-language, and global voice generation.
- Server-side activation through `features.kidsMode`, applied consistently to preview and export.

## Architecture

- Shared Kids Mode definitions in `@adt/types/kids`.
- Generation in `packages/pipeline`.
- API routes in `apps/api/routes/kids-voice.ts`.
- Runtime UI in `apps/adt-runtime/src/features/kids`.
- Authoring UI in `apps/studio`.
- Full implementation notes in `docs/KIDS_MODE.md`.

## Verification

- `pnpm typecheck` passes.
- Runtime suite: 39 files, 292 tests passing.
- Studio lint: 0 errors. Existing warnings are unchanged.
- Lingui extraction completes with no missing translations across `en`, `pt-BR`, `es`, `fr`, and `sq`.
- Mobile bottom sheet and reading chrome visually verified at 320x700 and 390x844, including active/inactive switches, reading controls, page navigation, and buddy feedback.

## Commits added in the final review

- `1c1103d54` Fix audit findings.
- `7a9e468b4` Enforce language parity and responsive menu behavior.
- `5d3599939` Polish the preview gate and mobile controls.


caroline-ssantoss (changes requested):
### Tested from three angles:

- Child-reader field test: 3-year-old son on a tablet, observing his behavior with the device. Used the eliezir.github.io/adt-kids-mode-demo 
- Reader test: self-tested the actual reading experience against the current PR branch, checked out and run locally.
- Author/creator test: self-tested the Studio authoring flow for setting up Kids Mode on a book, same local build.

### Issues
**user flow: Book reading with companion**
- End-of-book celebration message says the buddy's name, instead of the child's name. Reproduced on current PR build.

- "All done" button at end of book navigates one page back, instead of to a "next step". My suggestion is to have one button only, labeled "all done" that leads to the beginning of the book

- In the bottom area, there is a X button close to play, pause, etc. Closing this container leaves no way to reopen it. A dead end for the child mid-book.

- Confirmation voice (when a setting is changed, e.g. a new reading speed) overlaps the book's own narration. Book's narration should be paused while the confirmation plays.

**user flow: Enhancing a book with Kids mode**
- the left menu doesn't show the option "kids mode" under enhancements

- "Check plan" button at same visual weight as "Generate all buddy voices".
"check plan" provides info about what voices were already created, so credit consumption is under control. "generate all buddy voices" button is very close by, risking the user ato trigger a real API spend. I would not put both close in the screen. The info that "check plan" loads in the screen could be permanently visible. This is not a real "bug", but flagging as such because of the risks associated.

- It wasn't obvious that the buddy reading along the book text requires running the Language and Speech stages first — some inline guidance on the Kids Mode screen would help.


### Suggestions

**user flow: Enhancing a book with Kids mode**
- Clarify voice-generation scope. All voices need to be generated or can the user select characters? It was not clear to me.. I'd wish for the user to have control.

- Clarify what "Translate interface" does. I could not figure that out myself.

- Use the child's profile character as the settings entry point, instead of the buddy. Let the child tap their own profile face to access settings (speed, language, etc.), and let the buddy be purely the encouraging reading companion, not a menu system. The buddy could "jump" word-to-word as the book is read aloud, reinforcing its role as a playful reading companion rather than a settings hub. clicking/tapping the buddy could trigger a small animation too.

- Add an end-of-book reward/collectible: The child could earn a medal, points, a special hat, or some other collectible. Gamification hook to encourage finishing books. Not sure if we have the proper way to store such badges.

Issue #227, opened 2026-04-01
[Storyboard] Simple re-ordering support for text and images


elasticsounds:
Allow for re-ordering text and image elements in the storyboard. ie. put text after an image or before, swapping an image column with a text column. 

In reality, the llm will do a better job of re-rendering and re-ordering pages with instructions.

PR #573, opened 2026-07-16
fix(storyboard): preserve intentional reading order

### LINKED ISSUE #250: [Storyboard] Add controls to the storyboard to move objects back or next.
For objects on the storyboard, they may not be in the order you want. You should be able to move an object and swap its position with the object directly after it, within the existing div it resides in.
For divs you can also swap them with the next div within the div they reside in.
Same for images.

gbergengruen:
## Summary

Storyboard users can reorder paragraphs, images, and containers in the edit pane, but the subsequent web-rendering pass could silently restore the source PDF order. The overlay prompt was especially restrictive: it described the tree as the original top-to-bottom page order and instructed the model to recover exact positions from the PDF.

This PR makes an intentional storyboard reorder authoritative throughout rendering and review.

## What changed

- Marks drag-and-drop tree changes as an intentional reading-order edit.
- Carries that intent through the Studio API client, page re-render route, page-edit service, and pipeline render inputs.
- Provides the intent to both the initial web-generation prompt and the visual-review prompt.
- Updates standard web generation so sibling DOM order and visual sequence follow the edited content tree.
- Updates overlay generation so reordered groups receive positions consistent with the user-selected sequence instead of being restored to their previous PDF positions.
- Keeps the original PDF as a reference for styling, sizing, and general composition when an intentional reorder exists.
- Tells visual review not to reject or rewrite the user-selected order while continuing to review overlap, clipping, responsiveness, readability, broken images, and other visual defects.
- Leaves normal rendering unchanged when no intentional reorder was made.

## Resulting behavior

When a user reorders storyboard objects and saves:

1. The saved section tree remains the source of truth.
2. Web rendering emits elements in that tree order.
3. Overlay rendering visually positions groups in that same sequence.
4. Visual review accepts the order difference as intentional but still evaluates all other quality rules.

## Validation

- Focused Vitest suite passed: 5 test files and 66 tests.
- Added prompt regression coverage for standard generation, overlay generation, and intentional-order visual review.
- The overlay prompt regression specifically verifies that original-position restoration instructions are absent when user-selected order is active.
- Diff whitespace validation passes.

Fixes #250

### DIFF
--- a/apps/api/src/routes/pages.ts
+++ b/apps/api/src/routes/pages.ts
@@ -1160,11 +1160,16 @@ export function createPageRoutes(
 
     // Optional prompt for LLM guidance during re-render
     let prompt: string | undefined
+    let intentionalReadingOrder = false
     try {
       const body = await c.req.json()
-      const parsed = z.object({ prompt: z.string().optional() }).safeParse(body)
+      const parsed = z.object({
+        prompt: z.string().optional(),
+        intentionalReadingOrder: z.boolean().optional(),
+      }).safeParse(body)
       if (parsed.success) {
         prompt = parsed.data.prompt
+        intentionalReadingOrder = parsed.data.intentionalReadingOrder ?? false
       }
     } catch {
       // No body or not JSON — that's fine, prompt is optional
@@ -1214,6 +1219,7 @@ export function createPageRoutes(
             pageId,
             sectionIndex,
             prompt,
+            intentionalReadingOrder,
             booksDir,
             promptsDir,
             webAssetsDir,
@@ -1232,6 +1238,7 @@ export function createPageRoutes(
       pageId,
       sectionIndex,
       prompt,
+      intentionalReadingOrder,
       booksDir,
       promptsDir,
       webAssetsDir,
--- a/apps/api/src/services/page-edit-service.test.ts
+++ b/apps/api/src/services/page-edit-service.test.ts
@@ -219,6 +219,7 @@ describe("page-edit-service", () => {
         label,
         pageId,
         sectionIndex: 1,
+        intentionalReadingOrder: true,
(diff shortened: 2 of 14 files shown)


gbergengruen:
This will fix #227 as well.

ksokolovic:
@gbergengruen Thanks for the work on this. I investigated the behavior further and implemented an alternative based on this PR’s original description and the requirements in #227 and #250.

The replacement is now available in #658.

The current PR treats reordering as a special, transient state: Studio records an `intentionalReadingOrder` flag and carries it through the API and pipeline so rendering behaves differently only while that flag is preserved.

#658 takes a simpler and more durable approach:

- Once the section tree is saved, its leaf order is always authoritative.
- Standard and overlay rendering must preserve that order.
- Visual review must not restore the source PDF’s order.
- A deterministic validator rejects generated HTML when its text or image `data-id` order differs from the saved tree.
- No transient intent flag has to survive save/reload, rerender, cache reuse, restart, or another entry point.
- Editor regressions cover dragging individual images and complete image-and-caption containers.

I verified the replacement through:

- save and Studio reload;
- fresh and cached rerenders;
- API-process restart;
- desktop, tablet, and mobile previews;
- DOM and screen-reader order;
- packaged ADT output;
- two representative PDFs;
- both `openai:gpt-5.4` and `openai:gpt-5.4-mini`.

In every tested case, the generated DOM matched the saved content-tree order exactly.

I propose treating #658 as the replacement for this PR and closing #573 without merging. The replacement includes:

- `Fixes #227`
- `Fixes #250`
- `Supersedes #573`

Could you please verify that this interpretation matches the intended behavior, particularly whether anything from the conditional intent-flag design still needs to be preserved? Once confirmed, I will close this PR as superseded.

PR #574, opened 2026-07-16
feat(studio): select & style a blank group in the storyboard (#341)

### LINKED ISSUE #341: [Storyboard] Insert a blank text, div, element in storyboard
Allow the user to choose to add a blank text field or div. This text field should be added to the JSON as a new key so it is supported across the application.
For the div, allow the user to insert a blank div that the user can add classes to.

gbergengruen:
## Summary

Closes #341.

Lets users add a blank **group** (or blank text) in the storyboard and give it a deterministic look by styling it — no LLM required.

Previously, the tree editor's *Add group* / add-node actions only flagged an LLM re-render and injected nothing into the preview, so the new element had no `data-id` in the iframe and couldn't be selected or styled. This PR:

- **Injects the element into the rendered section HTML immediately** on add (reusing the existing `duplicateElementsInRendering` DOM-mirroring pattern) — deterministic, no LLM call.
- **Adds a selection path from both the tree and the preview**: a per-node *Edit styles* (palette) button opens the style pane for that element (and it auto-opens right after adding); empty elements get a clickable dashed placeholder box in the preview so they aren't zero-size.
- **Guards inline editing**: blank `<div>` containers are selected but not made `contentEditable`.
- **Preserves blanks across a later LLM re-render** via a `buildEditInstructions` "added element" branch.
- **Persists added sectioning nodes on save**: when a rendering edit also adds a tree node, `saveRendering` now saves the sectioning too, so sectioning and rendering version together (fixes a v1/v2 mismatch where the group vanished from the tree after reload).

No new "blank div/text" UI or terminology — reuses the existing *Add group* / add-node affordances. No `@adt/types` schema changes.

## Test plan

- [x] Add a group → dashed placeholder appears and style pane auto-opens; adding Tailwind classes applies live.
- [x] Deselect, click the group's tree node's *Edit styles* button → style pane re-opens.
- [x] Save → sectioning and rendering both bump to v2; group persists in the tree after full reload; no `data-adt-*` leak into saved HTML.
- [x] `pnpm --filter @adt/studio` typecheck + lint clean; i18n catalogs extracted and fully translated (en/pt-BR/es/fr/sq).

## Verification

Driven end-to-end with Playwright against a real book (storyboard stage): add-group injects `<div data-id>` (data-id count 18→19), auto-selects it, opens the **Container** style pane; save cycle confirmed `sectioning 1→2` / `rendering 1→2` with the node present in both and persisting across reload.

### DIFF
--- a/apps/studio/src/components/pipeline/stages/storyboard/components/BookPreviewFrame.tsx
+++ b/apps/studio/src/components/pipeline/stages/storyboard/components/BookPreviewFrame.tsx
@@ -72,6 +72,14 @@ export interface BookPreviewFrameHandle {
    *  place the selection toolbar when selection is driven from outside the
    *  preview (e.g. clicking a node in the section tree). */
   getElementRect: (dataId: string) => DOMRect | null
+  /** Locate a container element by the nearest common ancestor of its descendant
+   *  leaves (which carry data-ids), assigning it a transient data-id if needed.
+   *  Groups aren't tagged with their tree nodeId in the rendered HTML, so this is
+   *  how the section tree selects a pre-existing group for styling. Returns the
+   *  resolved data-id, rect and tag, or null when no descendant leaf is present. */
+  selectContainerByDescendants: (
+    leafDataIds: string[]
+  ) => { dataId: string; rect: DOMRect; tagName: string } | null
   /** Regenerate Tailwind CSS including the given extra HTML, then inject into iframe.
    *  Use after AI edits introduce new Tailwind classes not yet in the DB. */
   refreshCss: (extraHtml: string) => Promise<void>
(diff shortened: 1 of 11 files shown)


ksokolovic:
I reviewed the exact current head (`a80baf82`) against the current `develop` branch (`44da69ff`). My final verdict is to close this PR without merging and keep #341 open for a fresh implementation from current `develop`.

The main blockers are:

- **Tree/HTML structure can diverge.** Adding a child passes its parent ID as an insertion anchor, but the rendering path inserts the new element after that parent, or at the section root when the parent has no rendered ID. The saved sectioning JSON can therefore say “child” while the saved HTML contains a sibling or root element.
- **Container selection can style the wrong group.** The nearest-common-ancestor calculation is order-dependent for three or more descendant leaves and can resolve to a nested subgroup instead of the selected group.
- **Rendering and sectioning are not persisted atomically.** Rendering is saved first and sectioning second. If the second request fails, the book retains HTML for an element that is absent from the sectioning JSON, despite the intended version-together behavior.
- **Later LLM re-render preservation is not deterministic.** The generated instructions preserve tag, classes, and text, but omit the new element's parent and position.
- The branch is now 100 commits behind `develop` and conflicts in four source files plus all five locale catalogs.

I verified the head with Node 24: root typecheck, Studio production build, Studio lint, and the 24 section-tree operation tests pass. Those checks do not cover the failure paths above: this PR adds no automated tests for DOM insertion, container selection, cross-entity save consistency, or re-render preservation, and GitHub has no CI result attached to it.

There are useful UX ideas here, but fixing the structural mapping, persistence model, tests, and conflicts would amount to redesigning the feature against the current codebase rather than safely repairing this branch.

PR #587, opened 2026-07-22
Feature(Workflow): New beta tests workflow

zignaggo:
### Summary

This branch overhauls how beta builds are staged, released, and consumed. It introduces an on-demand **PR staging** pipeline, reworks the **release/versioning** scripts, and ships a new **in-app beta version browser** that lets testers install or downgrade to any beta release directly from the desktop app.

### Motivation

Previously, betas were cut only from `develop` and there was no safe way to build signed installers from an open PR before merging, nor a way for QA to move between beta versions inside the app. This branch makes staging explicit, safe, and reproducible, and gives testers full visibility into available beta builds.

### What changed

**Staging pipeline (CI)**
- New `staging.yml` workflow, dispatched manually **from `develop`** by PR number. It checks out `develop`, merges the PR head (`refs/pull/<n>/head`), and fails loudly on conflicts before any signing happens.
- Guardrails: the PR must be open, target `develop`, and not come from a fork; only users with write access can dispatch it.
- Produces signed desktop artifacts and a pre-release versioned `X.Y.Z-beta-pr-<n>` for QA.
- New `staging-cleanup.yml` removes staging branches/pre-releases via `pull_request_target`.
- `protect-release-tags.json` ruleset locks the `v*` tag namespace so only release automation can create tags.

**Release & versioning scripts**
- New `calculate-release-version.mjs`, `release-version.mjs`, `release-source-notes.mjs`, and `compose-release-notes.mjs` (all with tests) unify semver/pre-release parsing and generate release notes and beta provenance.
- `release.yml` updated for the new beta flow (skip-notarize path for macOS, pre-release publishing step).
- `docs/RELEASING.md` rewritten to document the branch-driven staging + release model.

**Desktop / Studio — beta version browser**
- New `release-catalog.ts` service (with tests) plus `auto-updater.ts` and IPC/preload changes to support browsing and installing arbitrary beta releases, including **downgrades**.
- New UI under `components/updates/beta/*` (version library, install button, release cover art, source/provenance cards, direction badge, search/empty states) and a refactored `UpdateDialog` split into `UpdateStateSurface`, `UpdateProgress`, and `UpdateWhatsNew`.
- New `use-beta-update-versions` hook and `scroll-area` UI primitive.

**i18n**
- All new user-visible strings translated across `en`, `es`, `fr`, `pt-BR`, and `sq`.


ksokolovic:
This is strong, security-aware release engineering — genuinely careful in the places that matter most. Before the findings, the things done right, because they're the hard parts:

- **`pull_request_target` (staging-cleanup) is used correctly** — no `actions/checkout` of PR code, and the PR number is validated `^[0-9]+$` before it reaches the shell. The classic "pwn request" holes are both closed.
- **Staging refuses to sign fork code with production certs** (`staging.yml:67-70`) — exactly the right call, since the desktop job holds the Apple/Azure/CSC secrets. Signing is also scoped to `environment: desktop`, and every build runs `--publish never`, so staging never touches the production update feed.
- **Channel separation is enforced in `release.yml`** — `develop` ships only `beta*`, `main` only stable (`:81-88`), `:latest` docker tag is gated to stable (`:101`), and a tag-collision check refuses to overwrite an existing release (`:96`).
- **No command injection** anywhere in the release scripts — all subprocess calls use `execFileSync(file, argsArray)` (no shell), and `repo`/`prNumber` are regex-validated before reaching a URL/API path. The version comparator is semver-correct and rejects non-safe-integer components, so a malformed tag can't produce a downgrade.
- **Release-notes composition is defended** — free-text fields (PR titles, commit subjects, branch names) are single-lined and backtick-rejected before interpolation, and URLs are host-allowlisted, so a crafted PR title can't forge a `- Branch:` line or inject a `javascript:` link. The catalog response is validated field-by-field and downloads are host-pinned to `github.com/unicef/adt-studio`.
- **Notes rendering is XSS-safe** (`ReleaseNotesMarkdown` emits React-escaped text, no `dangerouslySetInnerHTML`, links restricted to `https?:` opened with `noopener`), **i18n is complete** across all five locales with real translations, and the −452-line `UpdateDialog` change is a faithful extraction, not lost behaviour.

A few things to flag — mostly for conscious sign-off rather than defects.

### 🟠 Should fix / confirm intent

**1. Staging per-PR builds leak into the end-user beta catalog.** `staging.yml:297` publishes each staging build as a non-draft `--prerelease` tagged `vX.Y.Z-beta-pr-<N>`. `isBetaVersion` treats `beta-pr-*` as beta, and the catalog filter at `release-catalog.ts:84` excludes only **drafts** — so staging builds pass it. The result: an ordinary beta tester's "Beta versions" browser lists ephemeral per-PR test builds alongside real betas, and because `beta-pr-<slug>` sorts *above* a numbered `beta.N` of the same core, the auto-updater's `find(direction === "upgrade")` can pick one. Those builds are deleted on PR close, so a user who installs one lands on a version that then vanishes from the catalog. It's signed and origin-pinned, so not an integrity break — but it's almost certainly not intended for general beta users. Simplest fix: exclude the `beta-pr-*` slug pattern in the catalog filter (keep only `beta.N`), or publish staging as `draft`, or give staging its own tag namespace.

### 🟡 Security-relevant, defensible — worth explicit acknowledgement

**2. `allowDowngrade` is relaxed for beta users** — `auto-updater.ts` changes `allowDowngrade = false` to `allowDowngrade = isBeta`. The earlier security audit noted the hardcoded `false` as a deliberate anti-rollback protection. Relaxing it is defensible (the beta feature lets testers move between beta versions, which needs downgrade), and it's bounded — stable users keep the protection, and electron-updater still signature-verifies downloads on Windows/macOS. The residual is Linux, where there's no signature check (a pre-existing gap), so a beta Linux user is the most exposed to being steered to an older beta. Fine to ship, but it's a conscious loosening a maintainer should sign off on rather than absorb silently.

**3. Runtime `setFeedURL` is introduced** — `checkBetaRelease` calls `autoUpdater.setFeedURL({ provider: "generic", url: betaReleaseDownloadUrl(target) })`. The security audit had recorded "no runtime `setFeedURL`" as a hardening property (the update origin couldn't be redirected at runtime); this regresses it. It's bounded today because `betaReleaseDownloadUrl` pins the host to a hardcoded `https://github.com/unicef/adt-studio/releases/download` constant and only the URL-encoded tag varies — so a catalog entry can't redirect the feed off the official repo. That host-pinning is load-bearing: worth a short comment at the `setFeedURL` call so a future change doesn't let a catalog-supplied field into the host.

### 🟡 Low

**4. Release-notes images have no host allowlist.** `safeAssetUrl` in `ReleaseNotesMarkdown.tsx` accepts any `https://` image host, while `ReleaseCover.tsx` correctly allowlists (`TRUSTED_COVER_PREFIXES`). Since notes are attacker-influenceable (arbitrary beta tags), `` renders an `` that leaks the viewer's IP/UA on view — a tracking-pixel vector. Apply the cover allowlist here too.

**5. `isPending` should be `isLoading` for the loading flag (defensive).** `BetaVersionsView:50` uses `versionsQuery.isPending`; in TanStack v5 a query that's `enabled: false` and never fetched stays `pending`, so `loading` never clears. Verified this only bites when `window.api?.updates` is absent while on a beta channel — i.e. a degraded/non-Electron env, which shouldn't happen in the shipped desktop app, so it's effectively unreachable in practice. Worth fixing anyway: it's a one-word change to `isLoading`, and the queryFn already has a graceful `return []` path that `enabled: false` currently preempts, so the fix also makes the empty state reachable as the author evidently intended.

**6. Defense-in-depth:** CI-set `TRIGGER_SHA`/`TAG` reach `git` as bare arguments (e.g. `compose-release-notes.mjs:186,242`) with no `--` terminator; a value starting with `-` would be parsed as an option. Not attacker-controlled (set by the workflow) and no shell involved, so at most an error — but a `--` / `^[0-9a-f]{7,40}$` check is cheap.

### ⚠️ Process — please add a PR description (and let's make it the norm)

This PR ships 5,666 lines across 60 files of release infrastructure — new CI workflows, a version scheme, staging↔release↔beta-catalog relationships, code-signing paths — with an **empty description**. That's a lot to reverse-engineer from the diff, both for reviewing it now and for whoever debugs the release pipeline in six months.

Could you add one to this PR before merge, covering at least: what the beta/staging workflow does end to end, how the `beta.N` vs `beta-pr-N` version scheme works and which is user-facing, and what the `allowDowngrade`/`setFeedURL` beta-update behaviour is intended to be. A few of the findings above (1, 2, 3) are exactly the kind of thing a description would have pre-empted a reviewer's questions on.

More broadly — worth adopting as a team convention that non-trivial PRs carry a short description of intent and approach. It doesn't need to be long; even a few sentences on *why* and *what changed structurally* saves every future reviewer the reconstruction cost. A repo PR template would make it automatic (there isn't one today).

ksokolovic (changes requested):
Requesting changes on the strength of finding 1 in [my detailed review above](https://github.com/unicef/adt-studio/pull/587#issuecomment-5069896628) — staging per-PR builds leak into the end-user beta catalog and the auto-updater can pick one as an "upgrade," landing beta testers on a build that's deleted when the PR closes. That's the one item I'd want resolved or explicitly deferred before merge.

Everything else is either a conscious sign-off (2, 3 — the `allowDowngrade` / `setFeedURL` beta-update behaviour, worth a maintainer's nod) or minor. The core is genuinely well-built — this is a scoped ask, not a structural one.

zignaggo:
Fixes applied from review

- **1 — Staging builds leaking into the beta catalog:** the catalog filter now excludes the `beta-pr-*` pattern, so only numbered `beta.N` releases appear to testers and can be picked by the auto-updater.
- **2 — `allowDowngrade = isBeta`:** intentional. Downgrade is required for testers to move between betas;
- **3 — Runtime `setFeedURL`:** added a comment at the call site noting that the host is hardcoded to the official repo and must stay pinned (no catalog-supplied field may reach the host).
- **4 — Release-notes image host allowlist**: Fixed
- **5 — Loading flag**: Fixed
- **6 — CI git args**: Fixed

PR #593, opened 2026-07-23
fix: improve image matching with content-digest for flip detection

### LINKED ISSUE #590: Fix: PDF raster images with horizontal flip transformations extracted inverted
### Preliminary Checks

- [x] I have read and understood the important section above.
- [x] I am filing a bug report, not a feature request.

## Problem

When extracting raster images from PDFs, images placed with a horizontal flip transformation in the content stream are extracted inverted (mirrored left-right).

### Example
- **Page 15, Image 3** (the tobogán illustration):
  - Full page render (`pg015_page.png`): The slide ladder is on the **right** side, ramp descends **left**
  - Extracted image (`pg015_im003.jpg`): Ladder is on the **left**, ramp descends **right**
  - Both are the same image, but one is horizontally flipped

## Root Cause

In PDFs, image positioning and transformation is controlled by the **CTM (Current Transformation Matrix)** in the content stream. When an image is placed with a negative X scale (e.g., `[-W 0 0 H tx+W ty]`), it renders flipped horizontally.

- **During full-page rendering** (`page.toPixmap()`): mupdf applies the CTM → flip is rendered correctly
- **During raster extraction** (`extractRasterImagesFromPdf`): We read the raw XObject bytes directly from the PDF, without access to or application of the placement CTM → image is extracted in its original orientation (unflipped)

## Why Only Some Images?

Only images placed with horizontal flip transformations in the PDF are affected. Most illustrations are placed without transformation, so their raw bytes match their visual appearance on the page. The tobogán was the only image on page 15 flipped—likely intentional design (e.g., composed to face text direction).

## Solution

Extract the CTM from the content stream for each image, detect flip/rotation transforms, and apply them to the pixel data before saving the extracted image. This requires cross-referencing:
1. Image XObject reference in content stream op
2. The CTM at that point in the stream
3. Apply any scale/rotation transforms to the extracted pixels

## Related

- Code: `packages/pdf/src/extract.ts` — `extractRasterImagesFromPdf()` function
- Stream recorder: `packages/pdf/src/page-stream-recorder.ts` — has access to CTM

<html>
<body>
<h3 style="line-height: normal; font-size: 13px; font-weight: 600; margin: 1.5em 0px 0.875em; font-family: &quot;Segoe WPC&quot;, &quot;Segoe UI&quot;, sans-serif; color: rgb(191, 191, 191); font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><strong style="font-weight: 600;">Impact</strong></h3><div class="monaco-scrollable-element rendered-markdown-table-scroll-wrapper" role="presentation" style="width: 918px; margin-bottom: 16px; color: rgb(191, 191, 191); font-family: &quot;Segoe WPC&quot;, &quot;Segoe UI&quot;, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial; position: relative; overflow: hidden;"><div style="overflow: hidden;">
Component | Consequence
-- | --
LLM Processing | Receives image in wrong orientation; may make incorrect decisions about content/layout
Book Export | Illustrations appear flipped in final output
Storyboard Preview | Discrepancy between preview and actual rendering
User Experience | Confusion when exported book doesn't match what was shown during editing

</div></div>
</body>
</html>

### Expected behavior

When a PDF contains a raster image placed with a horizontal flip transformation in the content stream:  ✅ The extracted image (pg015_im003.jpg) should match the visual appearance on the page render (pg015_page.png) ✅ Both the full-page PNG and the cropped image should show the tobogán with: Ladder on the right side Ramp descending to the left ✅ When the extracted image is sent to the LLM for processing (image-cropping, segmentation, meaningfulness detection), it should be in the correct orientation ✅ Visual consistency across the entire pipeline: extract → storage → LLM → output

### Actual behavior

Actual Behavior Currently, the extracted raster image is horizontally flipped compared to the page render:  ❌ Page render (pg015_page.png):  Shows tobogán with ladder on the right, ramp descending left ✓ Correct ❌ Extracted image (pg015_im003.jpg):  Shows tobogán with ladder on the left, ramp descending right ✗ Inverted/Mirrored ❌ LLM receives:  The inverted image without any indication that it's been transformed Image-cropping model sees a flipped version of what's in the book Segmentation and other steps process incorrect orientation ❌ Visual inconsistency:  Preview (using page render) shows correct orientation Actual exported book (using extracted image) shows flipped orientation Storyboard displays inconsistent image

### Steps to reproduce

Prerequisites
Have ADT Studio running locally with any PDF that contains raster images placed with horizontal flip transformations in the content stream. Most PDFs won't have this, but any professionally designed book may have intentionally flipped illustrations for layout purposes.

Step-by-Step Reproduction
1. Open the Book

Launch ADT Studio. Navigate to Books and select any book from your collection. Click into the book to open it.

2. Run Extract Step

Navigate to the Pipeline and locate the Extract step. Select a page range that you know contains images (or run the full book). Click "Run Extract" and wait for completion.

3. Inspect Page Render

After extraction completes, navigate to the Image Filtering step. Select a page from the sidebar. View the full page render. Locate any illustration or diagram that has recognizable directional elements (a person facing a direction, an arrow, text orientation, etc.). Observe and mentally note the orientation of these elements.

4. Inspect Extracted Image

Still in the Image Filtering view, scroll down to see the extracted images for that page. Find the image that corresponds to the illustration you just observed in the page render. View or download the extracted image.

5. Compare Orientations

Compare the full page render with the extracted image. Check if any recognizable element that has a clear left/right or top/bottom orientation appears differently between the two images. Specifically, look for any illustration that appears horizontally flipped (left-right mirrored) in the extracted image compared to how it appears in the page render.

6. Confirm the Issue

If the extracted image is horizontally flipped compared to the page render, the bug is confirmed. The full page rendering applies the CTM correctly and shows the image in the intended orientation. The extracted image, however, shows the raw XObject bytes without the flip transform applied, resulting in a mirrored appearance.

When This Occurs
This issue occurs only with images that have been placed in the PDF with a negative X scale in their transformation matrix. Most images in PDFs are placed without transformation, so they are unaffected. However, designers may intentionally flip images for layout purposes, and those images will exhibit this bug during extraction.

### Screenshots and videos

_No response_

### Version

v0.7.4-beta.4

### What platform are you seeing the problem on?

Windows - x64

### What area is this issue related to?

Pipeline — Extract

### Relevant log output

```shell

```

gaguerre-iugo:
﻿## Issue
Fixes #590 - PDF raster images extracted with correct orientation

## Changes
- **Improved image matching**: Replaced dimension-only indexing with content-digest (SHA-256) 
  matching for precise 1:1 image identification
- **Fixed distortion bug**: Multiple same-sized images (girl + phones in pg012) now receive 
  their own specific flip based on their individual CTM, not another image's CTM
- **Fallback strategy**: Dimension matching only used if content digest unavailable

## Files Modified
- `packages/pdf/src/flip-utils.ts` - New `findMatchingStreamOp()` using content-digest
- `packages/pdf/src/extract.ts` - Refactored to use new matching strategy

## Testing
- ✅ pg015 image extraction orientation verified correct
- ✅ pg012 all 6 images extract with correct orientation
- ✅ Storyboard rendering verified (proportions now accurate)
- ✅ No regression on existing images

### Reviewer feedback resolution

- **Blocking — content-digest matching is a dead branch (wrong field):** ✅ Fixed by removing flip-stage re-matching and relying on placement-time consuming assignment.
- **Blocking — matching is re-solved instead of reused:** ✅ Fixed by stamping `flipTransform` during `stampRasterPlacementsFromOps` and applying flips from stamped data only.
- **Regression risk — exact-hash snapshot tests:** ✅ Verified by running `pnpm vitest run packages/pdf/src/__tests__`; added raven raster-hash stability guard.
- **No test coverage:** ✅ Added focused tests for CTM sign detection, PNG flip behavior + round-trip, and collision disambiguation path.
- **Lossy JPEG re-encode:** 🟡 Documented tradeoff (current JPEG flip path is lossy due to re-encode at quality 90).
- **Unused imports:** ✅ No longer applicable after refactor (imports are now used/clean).
- **Fragile JPEG vertical flip:** ✅ Fixed by replacing `Buffer.from(decoded.data.buffer)` row copy with offset-safe `decoded.data.subarray(...)` + `set(...)`.
- **Only axis-aligned flips handled:** ✅ Scope explicitly documented in CTM flip detection comments.
- **New dependency `jpeg-js` vs MuPDF path:** 🟡 Kept current pure-JS implementation; MuPDF-CTM render path noted as follow-up improvement.

### DIFF
--- a/packages/pdf/package.json
+++ b/packages/pdf/package.json
@@ -9,11 +9,11 @@
   "dependencies": {
     "@adt/types": "workspace:*",
     "@resvg/resvg-wasm": "^2.6.2",
+    "jpeg-js": "^0.4.4",
     "mupdf": "^1.27.0",
     "pngjs": "^7.0.0"
   },
   "devDependencies": {
-    "jpeg-js": "^0.4.4",
     "@types/pngjs": "^6.0.5"
   }
 }
--- a/packages/pdf/package.json
+++ b/packages/pdf/package.json
@@ -13,7 +13,7 @@
     "pngjs": "^7.0.0"
   },
   "devDependencies": {
-    "@types/pngjs": "^6.0.5",
-    "jpeg-js": "^0.4.4"
+    "jpeg-js": "^0.4.4",
+    "@types/pngjs": "^6.0.5"
   }
 }
--- a/packages/pdf/package.json
+++ b/packages/pdf/package.json
@@ -13,6 +13,7 @@
     "pngjs": "^7.0.0"
   },
   "devDependencies": {
-    "@types/pngjs": "^6.0.5"
+    "@types/pngjs": "^6.0.5",
+    "jpeg-js": "^0.4.4"
   }
 }
--- a/packages/pdf/src/__tests__/create-test-pdf.ts
+++ b/packages/pdf/src/__tests__/create-test-pdf.ts
@@ -206,6 +206,61 @@ Q
 }
 
 /**
+ * Create a 1-page PDF with a single raster image placed via a `cm` with a
+ * negative scale on one axis, simulating a genuinely mirrored image as it
+ * would appear in a real (possibly malformed) source PDF.
+ *
+ * The pixmap is asymmetric (left half red, right half blue for a horizontal
+ * flip; top half red, bottom half blue for a vertical flip) so the mirror is
+ * detectable by inspecting the *extracted* pixel buffer, not just its hash.
+ * This exercises the full `extractPdf` pipeline end-to-end (stream parsing →
+ * CTM-based flip detection → buffer flip), unlike unit tests that feed
+ * hand-authored CTM arrays directly to internal functions.
+ */
+export function createMirroredRasterTestPdf(axis: "horizontal" | "vertical"): Buffer {
+  const doc = new mupdf.PDFDocument();
+
+  const imgW = 20;
+  const imgH = 20;
+  const pixmap = new mupdf.Pixmap(mupdf.ColorSpace.DeviceRGB, [0, 0, imgW, imgH], false);
+  pixmap.clear(255);
+  const samples = pixmap.getPixels();
+  for (let y = 0; y < imgH; y++) {
+    for (let x = 0; x < imgW; x++) {
+      const i = (y * imgW + x) * 3;
+      const isFirstHalf = axis === "horizontal" ? x < imgW / 2 : y < imgH / 2;
+      if (isFirstHalf) {
+        samples[i] = 255; samples[i + 1] = 0; samples[i + 2] = 0; // red
+      } else {
+        samples[i] = 0; samples[i + 1] = 0; samples[i + 2] = 255; // blue
+      }
+    }
+  }
+
+  const image = new mupdf.Image(pixmap);
+  const imgObj = doc.addImage(image);
+
+  const xobjects = doc.newDictionary();
+  xobjects.put("Im1", imgObj);
+  const resourcesDict = doc.newDictionary();
+  resourcesDict.put("XObject", xobjects);
+  const resources = doc.addObject(resourcesDict);
+
+  // Negative scale on the flipped axis mirrors the image via the CTM.
+  const cm = axis === "horizontal" ? "-200 0 0 200 300 500" : "200 0 0 -200 300 500";
+  const stream = `
+q
+${cm} cm
+/Im1 Do
+Q
+`;
+  const buf = new mupdf.Buffer();
+  buf.writeLine(stream);
+  doc.insertPage(-1, doc.addPage([0, 0, 612, 792], 0, resources, buf));
+  return Buffer.from(doc.saveToBuffer("").asUint8Array());
+}
+
+/**
  * Create a 1-page PDF with a raster image, overlapping vector, AND nearby text label.
  * The text "Figure 1" is placed just below the image+vector figure.
  * Used to verify text label absorption expands the figure group bbox.
--- a/packages/pdf/src/__tests__/extract-raster-flip-e2e.test.ts
+++ b/packages/pdf/src/__tests__/extract-raster-flip-e2e.test.ts
@@ -1,0 +1,68 @@
+import { describe, expect, it } from "vitest";
+import { PNG } from "pngjs";
+import { extractPdf } from "../extract.js";
+import { createMirroredRasterTestPdf } from "./create-test-pdf.js";
+
+/**
+ * End-to-end regression test for the raster flip pipeline.
+ *
+ * Unlike the unit tests in `flip-utils.test.ts` and
+ * `extract-raster-placement.test.ts` (which feed hand-authored CTM arrays
+ * directly to internal functions) and the `raven.pdf` snapshot test in
+ * `extract-raven.test.ts` (whose fixture contains no mirrored images, making
+ * the flip pass a no-op there), this test parses a real PDF content stream
+ * containing a `cm` with a negative scale and asserts the *extracted pixel
+ * data* has the mirror correctly baked in (matching what the PDF actually
+ * renders). This is the scenario from #590 that no other test exercises
+ * end-to-end.
+ */
+describe("extractPdf — mirrored raster image (end-to-end)", () => {
+  it("bakes a horizontal mirror into the extracted pixels", async () => {
+    const pdfBuffer = createMirroredRasterTestPdf("horizontal");
+    const result = await extractPdf({ pdfBuffer });
+
+    const rasters = result.pages[0].images.filter((i) => i.renderMethod === "raster");
+    expect(rasters).toHaveLength(1);
+
+    const image = rasters[0];
+    // flipTransform is cleared once applied (see applyFlipsToRasterImages),
+    // so by the time extraction finishes it's undefined again — the pixel
+    // buffer itself is the source of truth here.
+    expect(image.flipTransform).toBeUndefined();
+
+    // The embedded XObject bytes are left=red/right=blue, but the page's
+    // negative-X `cm` mirrors it visually when rendered. The flip pass must
+    // bake that mirror into the extracted buffer so it matches what the PDF
+    // actually displays: left=blue/right=red.
+    const png = PNG.sync.read(image.buffer);
+    const leftPixel = pixelAt(png, 0, 0);
+    const rightPixel = pixelAt(png, png.width - 1, 0);
+    expect(leftPixel).toEqual([0, 0, 255]);
+    expect(rightPixel).toEqual([255, 0, 0]);
+  });
+
+  it("bakes a vertical mirror into the extracted pixels", async () => {
+    const pdfBuffer = createMirroredRasterTestPdf("vertical");
+    const result = await extractPdf({ pdfBuffer });
+
+    const rasters = result.pages[0].images.filter((i) => i.renderMethod === "raster");
+    expect(rasters).toHaveLength(1);
+
+    const image = rasters[0];
+    expect(image.flipTransform).toBeUndefined();
+
+    // The embedded XObject bytes are top=red/bottom=blue, but the page's
+    // negative-Y `cm` mirrors it visually when rendered. The flip pass must
+    // bake that mirror into the extracted buffer: top=blue/bottom=red.
+    const png = PNG.sync.read(image.buffer);
+    const topPixel = pixelAt(png, 0, 0);
+    const bottomPixel = pixelAt(png, 0, png.height - 1);
+    expect(topPixel).toEqual([0, 0, 255]);
+    expect(bottomPixel).toEqual([255, 0, 0]);
+  });
+});
+
+function pixelAt(png: PNG, x: number, y: number): [number, number, number] {
+  const idx = (png.width * y + x) << 2;
+  return [png.data[idx], png.data[idx + 1], png.data[idx + 2]];
+}
--- a/packages/pdf/src/__tests__/extract-raster-placement.test.ts
+++ b/packages/pdf/src/__tests__/extract-raster-placement.test.ts
@@ -1,0 +1,75 @@
+import { describe, expect, it } from "vitest";
+import { _testing, type ExtractedImage } from "../extract.js";
+import type { ImageStreamOp } from "../page-stream-recorder.js";
+
+function makeRaster(
+  imageId: string,
+  pixelDigest?: string,
+): ExtractedImage {
+  return {
+    imageId,
+    pageId: "pg001",
+    buffer: Buffer.from([0]),
+    format: "png",
+    width: 100,
+    height: 100,
+    hash: imageId,
+    renderMethod: "raster",
+    pixelDigest,
+  };
+}
+
+function makeImageOp(
+  seqno: number,
+  contentDigest: string | undefined,
+  ctm: number[],
+): ImageStreamOp {
+  return {
+    kind: "image",
+    seqno,
+    bbox: { x0: 10, y0: 20, x1: 110, y1: 120 },
+    nativeWidth: 100,
+    nativeHeight: 100,
+    currentTransformationMatrix: ctm,
+    hasMask: false,
+    activeClipBbox: null,
+    activeClipPaths: [],
+    blendMode: "Normal",
+    alpha: 1,
+    contentDigest,
+  };
+}
+
+describe("stampRasterPlacementsFromOps", () => {
+  it("disambiguates same-dimension collisions by digest and stamps matching flip", () => {
+    const imageA = makeRaster("pg001_im001", "digest-a");
+    const imageB = makeRaster("pg001_im002", "digest-b");
+    const ops: ImageStreamOp[] = [
+      makeImageOp(10, "digest-b", [1, 0, 0, -1, 0, 0]),
+      makeImageOp(11, "digest-a", [-1, 0, 0, 1, 0, 0]),
+    ];
+
+    _testing.stampRasterPlacementsFromOps([imageA, imageB], ops);
+
+    expect(imageA.streamSeqno).toBe(11);
+    expect(imageB.streamSeqno).toBe(10);
+    expect(imageA.flipTransform).toEqual({ flipHorizontal: true, flipVertical: false });
+    expect(imageB.flipTransform).toEqual({ flipHorizontal: false, flipVertical: true });
+  });
+
+  it("falls back to stream order for same-dimension images without digests", () => {
+    const first = makeRaster("pg001_im001");
+    const second = makeRaster("pg001_im002");
+    const ops: ImageStreamOp[] = [
+      makeImageOp(20, undefined, [-1, 0, 0, 1, 0, 0]),
+      makeImageOp(21, undefined, [1, 0, 0, 1, 0, 0]),
+    ];
+
+    _testing.stampRasterPlacementsFromOps([first, second], ops);
+
+    expect(first.streamSeqno).toBe(20);
+    expect(second.streamSeqno).toBe(21);
+    expect(first.flipTransform).toEqual({ flipHorizontal: true, flipVertical: false });
+    expect(second.flipTransform).toEqual({ flipHorizontal: false, flipVertical: false });
+  });
+});
--- a/packages/pdf/src/__tests__/extract-raven.test.ts
+++ b/packages/pdf/src/__tests__/extract-raven.test.ts
@@ -98,4 +98,27 @@ describe("raven.pdf extraction", () => {
     expect(rasters[0].hash).toBe("360de701626b47da");
     expect(rasters[1].hash).toBe("b20ccebb2681ac27");
   });
+
+  it("keeps first-3-page raster hash snapshot stable with flip pass enabled", () => {
+    const rasterSnapshot = result.pages.map((page) =>
+      page.images
+        .filter((image) => image.renderMethod === "raster")
+        .map((image) => ({ imageId: image.imageId, hash: image.hash })),
+    );
+
+    expect(rasterSnapshot).toEqual([
+      [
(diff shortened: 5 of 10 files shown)


gbergengruen:
## Code review

Thanks for tackling the raster-orientation bug 🙏 The overall approach is right — recording the CTM on each `ImageStreamOp` and baking the flip into the pixels (since the placement bbox is axis-aligned and loses the sign) is the correct call. However, there are two blocking issues that mean the fix doesn't actually work for its own headline scenario, plus a few smaller items.

### 🔴 Blocking — content-digest matching is a dead branch (wrong field)

`findMatchingStreamOp` compares:

```ts
(op as ImageStreamOp).contentDigest === image.hash
```

But these are hashes of **different inputs** and are never equal:

- `op.contentDigest` = hash of the **decoded pixel samples** (`hashImagePixels`, `page-stream-recorder.ts`).
- `image.hash` = SHA-256 of the **encoded PNG/JPEG file bytes** (`hashBuffer(buf)`, `extract.ts:1855`).

So the content-digest branch never returns, and every lookup falls through to the dimension-only fallback — which returns the **first** op with matching `WxH` and doesn't consume it. For the exact case in the PR description (pg012's girl + phones as *same-sized* images), both images match the same op and receive the same flip — i.e. the distortion bug this PR aims to fix. Unique-dimension images happen to work, which likely explains why the manual test looked correct.

The field you want is **`image.pixelDigest`** (`extract.ts:143`, set in `stampPixelDigestsOnCollisions`), which the codebase already computes specifically to disambiguate same-dimension collisions and is designed to compare equal to `op.contentDigest`. Note it's only populated when there's a dimension collision — which is exactly the case that needs it.

### 🔴 Blocking — matching is re-solved instead of reused

`stampRasterPlacementsFromOps` (`extract.ts:1068`) already builds a correct **consuming** 1:1 op↔image pairing (via `candidates.splice`) one line before `applyFlipsToRasterImages` is called. The flip pass throws that away and re-matches from scratch with a non-consuming, buggy key. Cleaner and more robust: stamp the matched op's CTM (or the derived `{flipH, flipV}`) onto the image inside `stampRasterPlacementsFromOps`, and drive the flip from that — no second matching pass that can disagree with the first.

### 🟠 Regression risk — exact-hash snapshot tests

`applyFlipsToRasterImages` mutates `image.buffer`/`image.hash` and is now wired unconditionally into the main extract path. `extract-raven.test.ts` asserts exact raster hashes (e.g. `expect(rasters[0].hash).toBe("f7c69061b5fb89ed")`). Any fixture image drawn with a flipped CTM will now produce a different buffer/hash and break those assertions. Please run `pnpm --filter @adt/pdf test` and, if hashes change, verify the new orientation is actually correct before regenerating the snapshots.

### 🟠 No test coverage

234 lines added with no tests. A single unit test on `findMatchingStreamOp` with two same-dimension images and differing CTMs would have caught the first blocking issue. Worth adding tests for flip detection from CTM signs, a horizontal/vertical buffer-flip round-trip on a small known PNG, and the collision-matching path.

### 🟡 Smaller items

- **Lossy JPEG re-encode**: flips decode + re-encode JPEG at quality 90 → generational quality loss on already-compressed images. A flip is losslessly expressible; at minimum, worth documenting the tradeoff.
- **Unused imports**: `extract.ts` imports `detectFlipFromCurrentTransformationMatrix`, `flipImageBufferHorizontal`, `flipImageBufferVertical`, and the `ImageFlipTransform` type, but only `applyFlipsToRasterImages` is used.
- **Fragile JPEG vertical flip**: uses `Buffer.from(decoded.data.buffer)`, which ignores a non-zero typed-array `byteOffset` (works today only because jpeg-js happens to allocate at offset 0) and is inconsistent with the horizontal path — use `decoded.data` directly.
- **Only axis-aligned flips handled**: `b`/`c` (rotation/skew) are ignored. A pure 180° rotation coincidentally maps to flipH+flipV, but arbitrary rotations aren't handled — fine for storybooks, but worth a comment noting the scope.
- **New dependency `jpeg-js`**: it's pure-JS (good, aligns with the "pure JS over native" principle), but mupdf — already a dependency — can render an image through its CTM to a correctly-oriented pixmap, which would avoid both the new dep and the manual pixel loops. Worth considering as the cleaner implementation.

### Positives

- CTM recording in `page-stream-recorder.ts` is clean and well-documented.
- Baking the flip into pixels rather than relying on the renderer is the right approach given the bbox loses the sign.
- Gating the extra pixel-hashing behind `hasDuplicateDimensions` keeps the cost off the common path.

Happy to pair on the `stampRasterPlacementsFromOps`-based approach if useful.

gaguerre-iugo:
Reviewer feedback resolution (point-by-point)
🔴 Blocking — content-digest matching is a dead branch (wrong field)
✅ Fixed.
The incorrect matching path (contentDigest vs image.hash) was removed by eliminating flip-stage re-matching. Matching now relies on the canonical consuming assignment in stampRasterPlacementsFromOps, which already handles digest-based disambiguation for same-dimension collisions, and stamps flipTransform directly onto the matched image.

🔴 Blocking — matching is re-solved instead of reused
✅ Fixed.
applyFlipsToRasterImages no longer performs a second matching pass. It now consumes image.flipTransform stamped during stampRasterPlacementsFromOps, ensuring placement and flip use the same 1:1 assignment.

🟠 Regression risk — exact-hash snapshot tests
✅ Validated and covered.
The effective PDF test suite was run (pnpm vitest run packages/pdf/src/__tests__) and no raven hash regressions were observed.
Additionally, an explicit stability guard was added in extract-raven.test.ts for first-3-page raster {imageId, hash} snapshots.

🟠 No test coverage
✅ Addressed.
Added focused tests for:

CTM sign-based flip detection,
PNG horizontal/vertical flip behavior and round-trip invariants,
same-dimension collision disambiguation in the current placement matcher (stampRasterPlacementsFromOps),
JPEG vertical-copy offset safety.

🟡 Lossy JPEG re-encode
🟡 No functional change (documented).
Documented explicitly that current JPEG flips use decode + re-encode (quality 90), which can introduce generational quality loss.

🟡 Unused imports
✅ Resolved / no longer applicable.
After the refactor, imports in extract.ts are now consistent with actual usage.

🟡 Fragile JPEG vertical flip (Buffer.from(decoded.data.buffer))
✅ Fixed.
Replaced with offset-safe copying via decoded.data.subarray(...) + flipped.set(...), avoiding reliance on zero byteOffset. Covered by a dedicated unit test.

🟡 Only axis-aligned flips handled
✅ Documented.
Added explicit scope note in CTM flip detection: current logic handles axis-aligned flips (a/d) and intentionally ignores b/c (rotation/skew).

🟡 New dependency jpeg-js vs MuPDF-based path
🟡 Kept as-is for now (explicit decision).
jpeg-js remains as the current pure-JS implementation. A MuPDF-CTM render-based path is acknowledged as a potential follow-up refactor for simplification.

gbergengruen:
## Review

### Mergeability: ✅ clean

- `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, base `develop`
- All three CI checks pass (docker, i18n, test)
- The head is ~30 commits behind `develop`, but `develop` hasn't touched `packages/pdf/` since the merge base. I test-merged `origin/develop` into the head and ran the full pdf suite (**17 files / 165 tests pass**) plus `tsc --build packages/pdf` — both clean. No hidden merge risk.

### What I verified about the fix

**The CTM polarity is correct.** I dumped every image op's CTM across all 12 pages of `tests/fixtures/raven.pdf` — all 24 have `a > 0, d > 0`. Deriving the matrix mupdf hands to `fill_image` (PDF placement × page y-flip × mupdf's "PDF images are upside down" pre-fix) reproduces the observed `[388, 0, 0, 388, 0, 0]` exactly, which confirms `d > 0` means *upright* in this space. So `a < 0 ⇒ flipH` / `d < 0 ⇒ flipV` is the right reading.

**It works end-to-end.** I authored a synthetic PDF placing an asymmetric image (left half red, right half blue) with `-200 0 0 200 300 500 cm`. The extracted asset's left pixel goes red → blue and `flipTransform` is `{ flipHorizontal: true }`. The fix does what it claims.

Nice work on the reviewer-feedback round — dropping the re-matching pass in favour of stamping at placement time is the right call.

---

## Findings

### 1. 🔴 High — the new raven snapshot test is tautological

Since zero raven image ops have a negative scale, `keeps first-3-page raster hash snapshot stable with flip pass enabled` passes because the flip pass is a **no-op** on that fixture. It cannot catch a polarity inversion, swapped H/V, or an accidental always-flip.

The `flip-utils.test.ts` cases use hand-authored CTMs, so they only assert the function does what it's coded to do. Net result: **no test exercises a real mirrored PDF**, which is the actual thing #590 is about.

`create-test-pdf.ts` already authors raw content streams, so this is ~20 lines. My probe (which passed) is essentially it — clone `createRasterOnlyTestPdf`, make the pixmap asymmetric, and use a negative-scale placement:

```ts
// left half red, right half blue so a mirror is detectable
for (let y = 0; y < imgH; y++) {
  for (let x = 0; x < imgW; x++) {
    const i = (y * imgW + x) * 3;
    const left = x < imgW / 2;
    samples[i] = left ? 255 : 0;
    samples[i + 1] = 0;
    samples[i + 2] = left ? 0 : 255;
  }
}

const stream = `
q
-200 0 0 200 300 500 cm
/Im1 Do
Q
`;
```

Then assert the extracted raster's leftmost pixel is blue (mirror baked in) and `flipTransform` is `{ flipHorizontal: true, flipVertical: false }`. A `200 0 0 -200 …` variant covers the vertical axis.

### 2. 🔴 High — unguarded `jpeg.decode` can fail an entire extraction

Verified: a malformed JPEG makes `applyFlipsToRasterImages` throw, and neither `extractPage` nor its three call sites wrap it. Compare `extractRasterImagesFromPdf`, which try/catches every image individually.

This matters more than it looks. The `canRawExtract` fast path copies raw `DCTDecode` bytes **without ever decoding them**, so `jpeg-js` is now the first thing to parse those bytes — and it's considerably stricter than mupdf (arithmetic coding, 12-bit, truncated streams, plus its own `maxMemoryUsageInMB` guard, default 512 MB). One odd JPEG in one book fails the whole extraction, where previously it extracted fine.

Suggested fix — per-image guard, degrade to unflipped:

```ts
for (const image of images) {
  const { flipHorizontal, flipVertical } = image.flipTransform ?? { flipHorizontal: false, flipVertical: false };
  if (!flipHorizontal && !flipVertical) continue;
  try {
    // …flip + rehash…
  } catch (err) {
    // Leave the image unflipped rather than failing the page. jpeg-js is
    // stricter than mupdf, and this is the first decode of raw DCTDecode bytes.
    console.warn(`[pdf] flip failed for ${image.imageId}, keeping original orientation:`, err);
  }
}
```

### 3. 🟠 Medium — 180° placements decode + re-encode twice

Verified with spies: when `flipHorizontal && flipVertical` are both set, the code makes **2 `jpeg.decode` + 2 `jpeg.encode` calls** — double generational loss and double the CPU, for exactly the common 180°-rotation case.

Collapsing into a single `flipImageBuffer(buffer, format, { h, v })` that decodes once, applies both axes, and encodes once fixes the loss and also merges the four near-identical pixel loops in `flip-utils.ts` into one.

### 4. 🟠 Medium — `applyFlipsToRasterImages` isn't idempotent

It mutates `image.buffer` but leaves `image.flipTransform` set. A second call double-flips: H+H silently reverts to the original, H+V corrupts. Only called once per array today, so this is latent — but it's a one-line guard (clear `flipTransform` after applying, or track an `flipApplied` marker).

Relatedly, `pixelDigest` is stale after the flip (it hashes pre-flip pixels). Harmless today since it's only read *before* the flip, but worth a comment so a future reader doesn't reuse it downstream.

### 5. 🟡 Low — op→image mismatches are now baked into pixels

When digests are unavailable, `stampRasterPlacementsFromOps` falls back to stream-order pairing. A mispair used to mean wrong `bounds` (a positioning bug); now it *also* bakes the wrong mirror into the stored asset, which is not recoverable without re-extraction.

Same first-op-wins arbitrariness applies to a single XObject placed twice with different CTMs: I confirmed only the first op's `bounds` **and** flip survive. (Later placements being dropped is pre-existing behaviour, not introduced here — flagging only because the flip now inherits that same arbitrariness.) A short comment on `flipTransform` noting it shares `bounds`' first-op-wins semantics would help.

### 6. 🟡 Nits

- The JSDoc for `detectFlipFromCurrentTransformationMatrix` sits above the `ImageFlipTransform` **interface**, not above the function — so it currently documents the wrong symbol.
- `format: string` should be `ImageFormat` (already exported from `extract.ts` as `"png" | "jpeg"`). As typed, an unexpected format makes the flip silently no-op and the image stays mirrored with no signal; narrowing the type makes the fallthrough provably unreachable.
- `hashBuffer` is duplicated with the comment "avoids circular dependency" — but there is no cycle: `flip-utils.ts` imports only `type ExtractedImage`, which erases at compile time. Two definitions of the canonical image hash is a real footgun (change one, hashes diverge silently). Move it to a shared `hash-utils.ts` and import from both.
- Trailing whitespace on `flip-utils.ts:4`.
- `ImageFlipTransform` is part of the public `ExtractedImage` shape but isn't re-exported from `packages/pdf/src/index.ts` next to `ImageFormat`.

### 7. 🟡 On the `jpeg-js` dependency (your own noted tradeoff)

Beyond the generational loss you already documented: quality-90 re-encode also drops ICC/EXIF and expands grayscale JPEGs to 3 channels, so flipped JPEGs typically get *larger*, not smaller.

Two dep-free alternatives for the follow-up, both arguably better aligned with principles 5 and 6:
- Re-render the image through mupdf with the flip folded into the CTM — removes `jpeg-js` entirely.
- Emit flipped JPEGs as PNG — lossless, bigger files, no new dep.

Fine to defer, but worth an actual tracking issue rather than leaving it only in the PR description.

---

## Recommendation

**Merge after #1 and #2.** Both are small, and #2 is a genuine availability risk on books we haven't seen. #3–#7 are good follow-ups and shouldn't block.

gbergengruen:
I have manually tested this, and the code fixes the unintentional flip of the images. There are some recommended code fixes to be done, but then, I would merge it.

gaguerre-iugo:
Pushed `47b8145f` addressing the High/Medium findings from the review. Details per finding:

---

### 🔴 High — the raven snapshot test is tautological
**Followed the reviewer's exact recommendation.** Added `createMirroredRasterTestPdf()` (cloned from `createRasterOnlyTestPdf`, per the suggestion) with an asymmetric pixmap (red/blue halves) placed via a real negative-scale `cm` (`-200 0 0 200 …` for horizontal, `200 0 0 -200 …` for vertical) — exactly the two variants the review proposed. New test file `extract-raster-flip-e2e.test.ts` runs the *full* `extractPdf` pipeline (not internal functions with hand-authored CTMs) and asserts the extracted pixel buffer has the mirror actually baked in (leftmost/topmost pixel color), not just a stable hash. This now fails if the flip polarity is inverted, the H/V axes are swapped, or the flip pass is always-on/always-off — none of which the old `raven.pdf` snapshot test could ever catch, since that fixture has zero mirrored images.

### 🔴 High — unguarded `jpeg.decode` can fail an entire extraction
**Followed the reviewer's exact recommendation.** Wrapped the flip+rehash of each image in `applyFlipsToRasterImages` in a per-image `try/catch`, matching the existing pattern in `extractRasterImagesFromPdf`. On failure, the image is left unflipped and a `console.warn` is logged with the `imageId`, instead of the exception propagating up through `extractPage` and aborting the whole book. Added a regression test with a malformed JPEG buffer confirming: (1) no throw, (2) the bad image is left untouched, (3) the *next* image in the same batch is still processed normally.

### 🟡 Medium — 180° placements decode + re-encode twice
**Followed the reviewer's exact recommendation.** Collapsed the two sequential single-axis flips into one `flipImageBuffer(buffer, format, { flipHorizontal, flipVertical })` that decodes once, remaps pixels for both axes in a single pass, and encodes once — merging the four near-identical pixel loops into two (JPEG/PNG). `flipImageBufferHorizontal`/`flipImageBufferVertical` are kept as thin wrappers for API/test compatibility. Added a spy-based test (same technique the reviewer used to verify the bug) confirming exactly 1 `jpeg.decode` + 1 `jpeg.encode` call for a `flipHorizontal: true, flipVertical: true` image.

### 🟡 Medium — `applyFlipsToRasterImages` isn't idempotent
**Followed the reviewer's exact recommendation** ("clear `flipTransform` after applying"). `image.flipTransform` is now set to `undefined` once the flip is applied, so a hypothetical second call over the same array is a no-op instead of silently double-flipping (H+H reverting to the original, or any other combination corrupting the image). Also added the suggested comment on `pixelDigest` in `extract.ts` noting it's stale after the flip pass runs and must not be reused downstream. Added an idempotency regression test.

---

**Validation**: full `packages/*` vitest suite — 83 files / 1175 tests, all passing. `pnpm typecheck` (tsc --build, full monorepo) clean. `pnpm build` (tsc --build + `@adt/runtime` bundle) clean.

gbergengruen (approved):
This PR fixes the issue. This is ok to merge.

PR #601, opened 2026-07-24
Fix: Issues template tag list

zignaggo:
### Problem

Generating a stable release on `main` was not updating the version dropdown shown in GitHub issue forms. `v0.7.4` is a clear example: after its release, the "Version" dropdown users see still listed `v0.7.3 / v0.7.2 / v0.6.0` and was missing `v0.7.4`.

**Root cause:** GitHub renders issue-form templates from the repository's **default branch (`develop`)**, but stable releases run on `main`. The release workflow's `finalize` job commits the refreshed dropdown to `github.ref_name` (`release.yml`), i.e. only to `main`. There is no `main → develop` sync, so stable versions never reach the templates users actually see. (Beta releases don't hit this because they already run on `develop`.)

Evidence:
- `origin/main` @ v0.7.4 release commit: dropdown correctly includes `v0.7.4`.
- `develop` (default branch): dropdown still shows the stale beta-era list (`v0.6.0`, no `v0.7.4`).

### Fix

1. **Workflow** (`.github/workflows/release.yml`): new `finalize` step, gated on `github.ref_name == 'main'`, that after publishing the release checks out `develop`, re-runs `scripts/update-issue-template-versions.mjs`, and pushes the refreshed templates directly to `develop` (only when there's a diff; `[skip ci]`). The dropdown is derived purely from `git tag --list`, so re-running on `develop` after the tag is pushed reproduces the correct list — idempotent and safe.
2. **Immediate catch-up**: regenerated the three templates with version dropdowns so `develop` reflects the current tags right away (`+ v0.7.4`, `- v0.6.0`).

### Changes
- `.github/workflows/release.yml` — add "Sync issue template versions to default branch" step.
- `.github/ISSUE_TEMPLATE/{bug_report,accessibility,question}.yml` — regenerated dropdowns.


ksokolovic:
I updated this PR onto current `develop` and resolved the stale generated-template conflicts in favor of the current version lists.

During review, I reproduced a shallow-checkout failure that could remove historical versions from the issue forms. I added an explicit full-tag fetch and a fail-closed guard, with regression coverage for both incomplete and complete tag histories.

Verification completed:
- focused tests: 20/20 passed
- build and lint passed
- full CI: test, Docker, and i18n passed
- PR is clean and mergeable at `ac5832f4`

This is ready to merge.

PR #615, opened 2026-07-27
Cut pipeline LLM failures: connection ramp, render fidelity, issue analyzer

elasticsounds:
Reduces the repeated LLM-call failures seen when running full books through extract → sectioning → storyboard (they only recover after wasteful retries). Prefers deterministic auto-heals over prompt nudging wherever the correct action is unambiguous.

- **Connectivity burst** (`api`): the page-level concurrency pool fired all ~32 requests at once (a cold-connection thundering herd → `write EPIPE`/`ECONNRESET` at step start); it now ramps only the opening wave so keep-alive sockets warm up, leaving steady-state throughput unchanged (with tests).
- **Render text fidelity** (`prompts`): a shared `_render_fidelity` partial, included in all 10 web/activity render prompts, tells the model to keep each leaf's text whole inside its `data-id`, reproduce it exactly, emit every `data-id`, preserve standalone marker/label leaves (`"1."`, `"(a)"`), and not invent bare text; plus a marker-label example in `activity_matching`.
- **Deterministic validation auto-heals** (`pipeline`): alias the invented sectioning structure `boxed_text` → `panel`; auto-add the `fitb-sentence` class when blank markers are present instead of failing; and reinforce single-running-counter item numbering in the fill-in-the-blank prompt.
- **Triage tooling** (`scripts`): `analyze-book-issues.mjs` reads a book's `llm_log`/`step_runs` and itemizes failures by task type, pipeline stage, and error category (Markdown + CSV/JSON).

On follow-up runs, connectivity errors dropped to 0 and render/sectioning validation retries fell sharply.


elasticsounds:
Reviewed the diff (20 files: concurrency ramp, HTML-validator relaxations/auto-heal, sectioning alias, prompt fidelity rules, new analysis script). pnpm typecheck and the touched test suites pass (pnpm lint fails on a missing root eslint binary — pre-existing).

Four findings posted as inline comments:

#1 Launch ramp orphans rejected promises — unhandled rejection / silently swallowed step failure
await sleep(rampMs, …) is a macrotask, and p.finally() deletes p from executing as soon as it settles. If a task rejects during the ramp, nothing ever attached a rejection handler. Reproduced with the diff’s exact logic (4 items / concurrency 8 / item 0 rejects): rampMs=0 rejects correctly, rampMs=100 produces an unhandled rejection and resolves successfully, so the step is marked done. No unhandledRejection handler exists in the app, so Node’s default exits the API process. Reachable from the catalog-translation worker (no internal try/catch) — and it’s triggered by exactly the EPIPE/ECONNRESET failures the ramp was added to reduce.

apps/api/src/services/stage-runner.ts:456

#2 Enumeration-marker exemption ships a duplicated visible marker instead of failing
Allowing bare 1. / (a) removes the only check on a leaf’s leading marker being peeled out of its data-id element. The text-mismatch branch still substitutes the full expected text back in, so the marker renders twice. Verified: <p>1. <span data-id="tx001">Andika majibu…</span></p> with expected "1. Andika majibu…" returns valid: true, errors: [] and output <p>1. <span data-id="tx001">1. Andika majibu…</span></p>. This is the same failure mode the new _render_fidelity.liquid rule 1 was written to prevent.

packages/pipeline/src/validate-html.ts:343

#3 Roman-numeral branch matches ordinary words
[ivxlcdm]{1,7} is “any letters from that set”, not a roman numeral — did., mild., civil., mix., Mimi., (mimi) all match. Such a label leaf becomes optional in collectOptionalTextIds, so the renderer can drop it and validation passes, losing the text. Suggested [ivx] — passes all new tests, excludes the words.

packages/pipeline/src/validate-html.ts:52

#4 Recovery detection ignores row ordering
succeeded is built from all rows, so recovered is true for a success from an earlier run, or from the issue’s own row when success=1, error_count>0. The report labels it “no later success” and prints ”✅ Every issue was recovered on a later attempt”, which can hide a rerun that actually blocked output.

scripts/analyze-book-issues.mjs:251

Verified clean: STRUCTURE_ALIASES (panel is a real key; repair runs pre-validation on both the initial and refinement paths), the fitb-sentence auto-heal reaching stored HTML for both callers and working with the runtime hydrator when the class lands on the data-id element itself, all 10 _render_fidelity includes resolving, and the new activity_matching marker/term span pattern being compatible with the matching runtime (it keys off data-activity-item).

PR #616, opened 2026-07-27
fix(studio): show run feedback immediately when a stage starts

elasticsounds:
Starting Extract left the UI looking idle for ~5s: measured on a 5.7MB PDF, the book page painted an idle "Run Extract" button ~600-900ms after navigation and the violet task indicator did not appear until ~6.0s. Two causes — the book page mounted with a cold TanStack cache (the creation wizard navigated without seeding `step-status`, and that fetch then sat 0.4-2.3s behind `extractPDF`'s synchronous `readFileSync` + mupdf WASM startup), and `TaskIndicator` keyed purely off a *step* being `running` while ignoring a queued *stage*.

The wizard now seeds `step-status` before navigating (rolling back if the kickoff throws), `queueRun` no longer drops its optimistic update on a cold cache via `if (!old) return old`, `TaskIndicator` and `StageRunCard` render a "Starting…" state for a queued stage with no reporting step yet, and `extractPDF` yields via `setImmediate` after `step-start` so the status write and SSE frame escape before the sync work blocks the loop.

Verified in a real browser: all feedback now lands at 147ms after click (first sample interval), versus a 123ms baseline on the already-warm click path, which was never the broken one; typecheck, lint, and 1903/1903 tests pass with all 5 locales translated.

Two caveats — the wizard→landing path is verified by construction rather than end-to-end (Playwright could not get past the "Visual Layout" wizard step; the mechanism is identical to the click path that *was* verified end-to-end, but it mounts a fresh `BookRunProvider`, so a manual check is worthwhile), and the `setImmediate` yield is a partial improvement only (634→375ms, 2198→1810ms, 2349→1960ms in paired trials) since extraction keeps the loop busy afterwards — the optimistic client state is what actually makes the UI immediate.

### DIFF
--- a/apps/studio/src/components/pipeline/components/StageRunCard.tsx
+++ b/apps/studio/src/components/pipeline/components/StageRunCard.tsx
@@ -97,6 +97,20 @@ export function StageRunCard({
     })
     .find((item) => item !== null)
 
+  // The run has been kicked off but no step has reported yet — the server is
+  // still in its synchronous stage startup (for Extract: reading and parsing
+  // the PDF, seconds on a large book). Spin the first pending sub-step so the
+  // card body doesn't sit inert and grey while the header already says
+  // "Running". Resolves into the real step state on the first step-start.
+  const anyStepRunning = subSteps.some(({ key }) => stepState(key) === "running")
+  const startingStepKey =
+    isRunning && !anyStepRunning
+      ? subSteps.find(({ key }) => {
+          const state = stepState(key)
+          return state !== "done" && state !== "skipped"
+        })?.key
+      : undefined
+
   return (
     <Card className={cn("overflow-hidden max-w-xl shadow-none", borderColor)}>
       {/* Colored header */}
@@ -132,7 +146,7 @@ export function StageRunCard({
               const numericProgressLabel = hasPages ? `${progress.page}/${progress.totalPages}` : null
               const messageProgressLabel = progress?.message?.trim()
               const progressLabel = messageProgressLabel || numericProgressLabel
-              const isSubRunning = state === "running"
+              const isSubRunning = state === "running" || key === startingStepKey
               const showInlineProgress = Boolean(
                 isSubRunning &&
                 progressLabel &&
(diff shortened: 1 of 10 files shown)


ksokolovic:
Reviewed this for stability and feature quality, and it holds up well. The mechanism is sound in the ways that matter:

- The optimistic "queued" state can't get permanently stuck — the status query polls every 2s precisely because a stage reads queued/running, so authoritative server state overwrites any stale optimistic value within ≤2s. It's self-correcting by construction.
- The "Starting…" rows resolve out cleanly: the sidebar shows them only while no step is reporting (`stageProgressRows + activeSteps === 0`), and the card's `startingStepKey` drops to undefined the moment a real step runs. Both transition into real state on the first `step-start`.
- The wizard's seed → kickoff → rollback-on-throw → navigate ordering is correct, and `removeQueries` is the right rollback (lets the book page refetch fresh rather than assume a state).
- The `setImmediate` yield is transparent — a one-tick pause on an already-async `extractPDF`, no change to extraction behaviour, extraction tests still pass.

i18n is complete and actually translated across all five locales, typecheck is clean, and the studio+pipeline suites pass. Also appreciated the honesty in the description about the two soft spots — the wizard path being verified by construction and `setImmediate` being only a partial win. That kind of candour makes the review easier to trust.

Two things, neither blocking:

**[1] No automated tests for the new state logic.** This is the one I'd want closed, here or as a fast-follow. The "Starting…" show/hide guard, the cold-cache seed in `queueRun`, and the wizard rollback are exactly the kind of state-transition logic a later refactor can silently break — and if it does, it reintroduces the dead-UI gap this PR fixes, invisibly. A couple of focused unit tests (the TaskIndicator "Starting…" conditions given a queued-but-no-step state; the `queueRun` cold-cache seed producing a valid optimistic entry) would lock the behaviour in. The browser verification is good for shipping now, but it doesn't guard against regression.

**[2] Minor nuance, not a defect.** Because "Starting…" now surfaces the optimistic-queued window, a *failed* run-button click will briefly flash "Starting…" for up to ~2s before the poll corrects it, where before it showed nothing. That's clearly the right trade against several seconds of dead UI on the common (successful) path — just flagging it as expected behaviour so it isn't mistaken for a bug later.

Looks good to me — approving.

PR #621, opened 2026-07-27
In-app Assistant chat — early prototype for feedback (not for merge)

caroline-ssantoss:
## What this is

This is **not ready to merge** — I'm opening it as a draft so you can see where my head's at, follow the reasoning, and poke holes in it. No expectation you review it like a normal PR; I mainly want you to read the intention and the open questions below.

The core idea: a persistent "Assistant" chat panel inside Studio (bottom-right bubble, opens a side panel) that helps a user get unstuck — either about the specific page they're looking at, or a general "how does ADT Studio work" question. In this version it's **guidance-only**: it tells the user what to click, it never calls a mutating endpoint itself.

Why I wanted this: while testing a PDF→activity-book conversion, the pipeline merged two distinct activities into one section, and there's genuinely no "split" action in the product today — only Merge/Duplicate/Delete. The only path is a manual workaround, and nothing in the UI tells you that. This is meant to close that gap generally, not just for that one case.

Full technical writeup (request flow, prompt design, data contract, how it plugs into the existing `@adt/llm` client and logging) is in [`docs/ASSISTANT.md`](https://github.com/unicef/adt-studio/blob/worktree-assistant-widget/docs/ASSISTANT.md) — that's the place for the "how does this actually work" detail; this description is more about intent and open questions.

## How I tested it

- Rebased onto latest `develop` (this branch had gone stale for a bit) — `pnpm typecheck` and `pnpm lint` both clean.
- Verified the widget's rendering (bubble, panel open/close, empty state, no-API-key gate, composer) with real browser screenshots, not just "it compiles."
- Tested against a real book with my own OpenAI key. This surfaced two real problems, both fixed before this PR:
  - **A hallucination.** I asked it something not covered by the current page or my prompt's "facts" section, and it confidently invented a UI flow that doesn't exist (a fictional "share this book with a coworker" settings menu). Root cause: I'd told it to answer broader questions but hadn't grounded it in what's actually true about the product, combined with a "never hedge" tone rule — confident wording made the wrong answer more convincing, not more correct. Fixed by adding an explicit facts section to the prompt (single-user app, book-level storage, versioning, caching, etc.) plus a rule that confidence is about *how* it says true things, never license to invent untrue ones. Worth reading as a cautionary note if this prompt gets extended further.
  - **A layout bug.** A UI tweak I made (see below) rendered a dropdown that was structurally present but visually clipped to invisibility by a parent's `overflow-hidden`. Only caught by actually opening it in a browser, not by typecheck/lint.
- Every one of the assistant's LLM calls shows up in the existing Debug → LLM Logs panel exactly like every other AI feature's calls do — I actually watched this happen against my own test book, so that's not just an aspirational claim in the docs.

## What I intentionally did not do

- **No automated tests.** This is deliberately left for whoever picks this up — `assistant-chat-service.ts` and the route have no `*.test.ts`, unlike sibling AI-feature services. I see this as your call on when/how, not something to guess at.
- **No deep documentation linking yet** — see below, this is one of the open points.
- **No action-capability.** It only advises; it can't invoke Duplicate/Merge/Delete/etc. itself. Whether that's worth building is a v2 question, not something I've tried to answer here.

## Small unrelated thing bundled in

While testing I also moved the "Debug Panel" button — it used to live in the app's top-left logo bar, which never made sense since it opens per-book data. It's now a dropdown on the book's own row in the sidebar, with "Debug Panel" and "API Keys" as the two options (that row's gear used to only open API Keys). Separate concern from the Assistant itself, just happened to ride along in the same branch since I found it while testing.

## Open points — things on my mind, not yet built

1. **I want every answer to link back to the docs.** Right now there's a `DOCS_BASE_URL` env var wired end-to-end (route → service → prompt) that lets the assistant point to the documentation site — but it's deliberately restricted to linking the docs *home page* only, because the prompt has no index of what pages actually exist, and I didn't want it fabricating a specific sub-page URL that 404s. To do this properly we'd need to feed it an actual index (title + path per doc page) so it can link to the *right* page, not just the front door. Also worth noting: the docs site (`apps/site`) lives on a totally separate branch from `develop`, so this isn't wireable yet in a real sense until that lands somewhere this app can see it.

2. **I want a way to see what people are actually asking it**, so the docs work can be prioritized by real usage instead of guesswork — e.g. if "how do I do X" comes up constantly and there's no doc page for X, that's a strong signal of what to write next. Nothing like this exists today: the LLM logs record individual calls for transparency/debugging, but there's no aggregation layer that would let someone look at "top questions this month" across users. This is the one I'd love your read on most — whether it belongs in this app at all, or as a separate analysis pulled from the logs.

Would genuinely like your take on either of these, or anything else that jumps out at you.

### DIFF
--- a/apps/api/src/app.ts
+++ b/apps/api/src/app.ts
@@ -34,6 +34,7 @@ import { createReviewerValidationRoutes } from "./routes/reviewer-validation.js"
 import { createTocRoutes } from "./routes/toc.js"
 import { createSignLanguageVideoRoutes } from "./routes/sign-language-videos.js"
 import { createTranslationEvaluationRoutes } from "./routes/translation-evaluations.js"
+import { createAssistantRoutes } from "./routes/assistant.js"
 
 // Resolve paths relative to monorepo root (2 levels up from apps/api/)
 const projectRoot = path.resolve(
@@ -47,6 +48,7 @@ const configPath = path.resolve(
 const configFolderPath = path.resolve(
   process.env.CONFIG_FOLDER_PATH ?? path.join(projectRoot, "config")
 )
+const docsBaseUrl = process.env.DOCS_BASE_URL
 
 let webAssetsDir: string
 const adtResourcesZip = process.env.ADT_RESOURCES_ZIP
@@ -118,6 +120,7 @@ app.route("/api", createSpeechConfigRoutes(configPath))
 app.route("/api", createReviewerValidationRoutes(booksDir, configFolderPath, configPath))
 app.route("/api", createSignLanguageVideoRoutes(booksDir))
 app.route("/api", createTranslationEvaluationRoutes(booksDir, configPath, taskService))
+app.route("/api", createAssistantRoutes(booksDir, promptsDir, configPath, docsBaseUrl))
 
 export default app
 export { booksDir }
--- a/apps/api/src/routes/assistant.ts
+++ b/apps/api/src/routes/assistant.ts
@@ -1,0 +1,48 @@
+import { Hono } from "hono"
+import { HTTPException } from "hono/http-exception"
+import { AssistantChatRequest, parseBookLabel } from "@adt/types"
+import { assistantChat } from "../services/assistant-chat-service.js"
+
+export function createAssistantRoutes(
+  booksDir: string,
+  promptsDir: string,
+  configPath?: string,
+  docsBaseUrl?: string
+): Hono {
+  const app = new Hono()
+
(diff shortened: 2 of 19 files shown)


ksokolovic:
Thanks for putting this prototype together and documenting the thinking behind it—it helped clarify the opportunity and tradeoffs. Since it was intentionally created for exploration rather than merging, I’m going to close the PR. We can carry the useful ideas forward into a focused implementation if we decide to pursue the Assistant feature.

PR #624, opened 2026-07-27
Feat/activity underline text

kinggamma:
Adds `activity_underline_text` as a new Storyboard activity type, per review feedback.

## What's included
- Prompts: `activity_underline_text` + `activity_underline_text_answers`
- Render strategy in `config.yaml`; section-type constants; wizard presets
- Disambiguation guidance in `page_sectioning.liquid`
- Structure-validator rules, plus a deterministic repair pass for when the model
  returns a sentence with no selectable spans
- Runtime initializer with keyboard operability and screen-reader semantics
  (role/name/state per token, question grouping, live-region announcements)
- Interface translations for en, es, fr, pt-BR, sq, sw

## Testing
`pnpm typecheck` clean; 84 tests across 7 suites. Screen-reader semantics verified
against Chromium's accessibility tree (accessible names, roles, and checked state)
and axe-core.

## Note for reviewers
added tsc and adt-runtime watch to pnpm dev; the runtime package wasn't being rebuilt in dev, which made iterating on runtime activity code impractical


elasticsounds:
- Words moving if you click to underline. Stops moving words.
- Underline only by color not accessible. Success criteria requires icons for validation. Added X and check icons to items.
- Prompt examples inherently omitted words like filler words (and, or) so not all words were clickable. The activity requires all words can be chosen.
- Raw text with underlines may not communicate clearly that words are clickable - added subtle grey border to words that respond to hover action to make this more obvious.
- Validator ensuring blank sentences are clickable extended to ensure all words or word-groups in sentences are clickable.
- Support added for "underline the correct sentence" so if it's not a word only it should still be functional.

PR #641, opened 2026-07-29
feat(tts): add ElevenLabs as a Speech step TTS provider

### LINKED ISSUE #630: Add ElevenLabs as a new TTS provider for the Speech step
### Preliminary Checks

### Problem / motivation
The Speech step currently supports three TTS providers — **Azure**, **Gemini**, and **OpenAI** — configurable per book via `speech.default_provider` / `speech.providers` in `packages/types/src/speech.ts`. **ElevenLabs** is a widely-used TTS provider with strong multilingual voice quality that is not yet available as an option, forcing users who prefer it to work around ADT Studio&#39;s speech pipeline entirely.

### Proposed solution
Add **ElevenLabs** as a TTS provider for the Speech step, following the exact same architecture and patterns already established for Azure/Gemini/OpenAI:

- **`packages/llm/src/speech.ts`**: add `createElevenLabsTTSSynthesizer()` implementing the existing `TTSSynthesizer` interface (`synthesize(options): Promise`), calling the ElevenLabs `text-to-speech` REST API, exported from `packages/llm/src/index.ts`.
- **`packages/types/src/speech.ts`**: no schema changes expected (provider is already a free-form string key in `providers`), but default model/voice/format helpers in `packages/pipeline/src/speech.ts` (`resolveModel`, `resolveVoice`, `resolveSpeechFormat`, etc.) need an `&#34;elevenlabs&#34;` branch with sensible defaults.
- **`packages/pipeline/src/pipeline-dag.ts`** and **`apps/api/src/services/stage-runner.ts`**: wire up `createElevenLabsTTSSynthesizer` alongside the existing Azure/Gemini construction in the provider-selection switch (`getSynthesizer` / batch run paths).
- **`apps/api/src/routes/tts.ts`**: add the ElevenLabs branch to the single-item regeneration/preview endpoint&#39;s synthesizer selection, mirroring the Azure/Gemini/OpenAI ternary.
- **API key plumbing**: add an `X-ElevenLabs-API-Key` header (mirrors `X-Gemini-API-Key` / `X-Azure-Speech-Key`) in `apps/studio/src/api/client.ts`, and route/validate it server-side alongside the other provider keys.
- **Studio UI**:
  - `apps/studio/src/hooks/use-api-key.ts`: add `elevenLabsKey` / `setElevenLabsKey` (localStorage-backed, same pattern as `geminiKey`/`azureKey`).
  - `apps/studio/src/components/settings/ApiKeysSettings.tsx` + `ApiKeyDialog.tsx` + `BookApiKeyDialogProvider.tsx`: add an ElevenLabs key field.
  - `apps/studio/src/components/pipeline/components/ModelSelect.tsx`: add `ELEVENLABS_TTS_MODELS` group and include it in `ALL_TTS_MODEL_GROUPS`.
  - `apps/studio/src/components/pipeline/stages/languages/LanguageSettings.tsx` / `LanguageView.tsx` and `VoiceMappingsEditor.tsx` / `SpeechPromptsEditor.tsx`: add ElevenLabs to provider selection / per-language routing and voice mapping UI, consistent with Azure/Gemini handling.
  - `apps/studio/src/lib/speech-routing.ts`: no changes needed (provider routing is already generic over provider name), but add coverage in `speech-routing.test.ts` for an `&#34;elevenlabs&#34;` provider.
- **i18n**: wrap any new user-visible strings in Lingui macros and run `pnpm --filter @adt/studio extract`, translating all locales (`en`, `pt-BR`, `es`, `fr`, `sq`).
- **Tests**: mirror `packages/llm/src/__tests__/speech.test.ts` (Gemini fetch-mocked tests) with an ElevenLabs equivalent, plus pipeline/stage-runner/tts route test coverage mirroring existing Azure/Gemini cases.
- **Docs**: update any provider-listing docs (e.g. `docs/GUIDELINES.md` / `docs/DECISIONS.md` if they enumerate TTS providers).

### Alternatives considered
- Using ElevenLabs only through a &#34;custom&#34; HTTP provider escape hatch — rejected because the other three providers are first-class, and Speech step configuration (voice mapping, rate limits, provider routing per language) is provider-name-driven, so ElevenLabs needs to be a recognized provider key (`&#34;elevenlabs&#34;`) to fit existing UI/config flows without special-casing.

Pipeline — Speech

### Additional context
Existing reference implementations to follow for architecture/pattern parity:
- `packages/llm/src/speech.ts` — `createAzureTTSSynthesizer`, `createGeminiTTSSynthesizer`, `createTTSSynthesizer` (OpenAI)
- `packages/pipeline/src/speech.ts` — provider-aware model/voice/format resolution, rate-limit resolution (see `resolveGeminiTtsRateLimit`)
- `packages/pipeline/src/pipeline-dag.ts` (~line 900) and `apps/api/src/services/stage-runner.ts` (~line 2600) — synthesizer construction per provider during a full run
- `apps/api/src/routes/tts.ts` (~line 800) — synthesizer construction for single-item regeneration
- `apps/studio/src/hooks/use-api-key.ts`, `apps/studio/src/api/client.ts` — per-provider API key storage/headers
- `apps/studio/src/components/pipeline/components/ModelSelect.tsx` — `AZURE_TTS_MODELS` / `GEMINI_TTS_MODELS` / `OPENAI_TTS_MODELS` groups

---

### Implementation note (post-review UI change)
While implementing this and addressing code review feedback in PR #641, we also moved **"Batch a whole page per request (experimental)"** (`speech.batch_by_page`) out of the global Speech settings panel and **into the per-language card, shown only when that language is routed to Gemini** — mirroring how the ElevenLabs-only tuning options above already render per-language. This setting only ever applied to Gemini-routed languages, so this makes that scoping explicit in the UI without changing the underlying config field or behavior. See [PR #641](https://github.com/unicef/adt-studio/pull/641) for details.

gaguerre-iugo:
Closes #630

## Summary

Adds **ElevenLabs** as a TTS provider for the Speech step, alongside Azure, Gemini, and OpenAI, following the same architecture/patterns already established for the existing providers — plus two ElevenLabs-specific tuning options.

### Base provider implementation
- `packages/llm/src/speech.ts`: `createElevenLabsTTSSynthesizer()` implementing the `TTSSynthesizer` interface, calling the ElevenLabs `text-to-speech` REST API.
- `packages/pipeline/src/speech.ts` + `pipeline-dag.ts`: `elevenlabs` branch in model/voice/format resolution and synthesizer construction, mirroring Azure/Gemini.
- `apps/api/src/services/stage-runner.ts`, `apps/api/src/routes/tts.ts`: ElevenLabs wired into the full-run stage and the single-item regeneration/preview endpoint, including `X-ElevenLabs-API-Key` header handling.
- Studio UI: API key tab (`ApiKeysSettings.tsx`, `ApiKeyDialog.tsx`, `BookApiKeyDialogProvider.tsx`, `use-api-key.ts`), `ELEVENLABS_TTS_MODELS` in `ModelSelect.tsx`, provider selection in `LanguageSettings.tsx` / `LanguageView.tsx` / `VoiceMappingsEditor.tsx`, and the Speech landing page.
- `config/voices.yaml`: default ElevenLabs voice — Rachel (`21m00Tcm4TlvDq8ikWAM`), kept in sync with the pipeline fallback and the UI summary via the shared `DEFAULT_ELEVENLABS_VOICE_ID` constant in `@adt/types`.
- Tests mirroring existing provider coverage in `packages/llm`, `packages/pipeline`, and `apps/api/src/routes/tts.test.ts`.

### ElevenLabs-only tuning options
Two additional options, visible in each per-language card **only when that language is routed to ElevenLabs**:
- **Text Normalization** (`elevenlabs_apply_text_normalization`, auto/on/off) — controls whether ElevenLabs expands abbreviations, numbers, and symbols into spoken words (e.g. &#34;N.º&#34; → &#34;Número&#34;).
- **Use adjacent text as context** (`elevenlabs_use_context`) — sends the neighboring sentence&#39;s text as `previous_text`/`next_text` so intonation carries across sentence boundaries.

Both are threaded through `types → llm → pipeline → stage-runner`, folded into `computeSpeechCacheKey` only for the `elevenlabs` provider (mirrors the existing Gemini temperature/seed gating), and covered by new tests.

### Reliability
- **Rate-limit / concurrency handling** (`stage-runner.ts`): ElevenLabs throttles by concurrent requests (plan-dependent, e.g. Free ~4, Starter ~5), not by RPM like Gemini. The Speech loop caps concurrency to `ELEVENLABS_TTS_MAX_CONCURRENCY` (4) whenever any work is routed to ElevenLabs, running that pass concurrently alongside a separate full-concurrency pass for all other providers (`ttsWorkItems` partitioned by provider, each processed via its own `processWithConcurrency` call) — so a book that mixes ElevenLabs with other providers no longer loses throughput on the non-ElevenLabs languages. Retries `429`/`5xx` responses with exponential backoff (2s → 30s cap), respecting cancellation. Gemini's adaptive limiter and retry path are unchanged.
- **Default-voice consistency**: the ElevenLabs default is unified on Rachel across `voices.yaml`, the pipeline fallback, and the UI summary, via a single `DEFAULT_ELEVENLABS_VOICE_ID` constant in `@adt/types` (no duplicated literals).
- **Cache-key parity across all three execution paths** (API stage-runner, CLI/`runFullPipeline`, single-item regeneration route): all three now forward the same ElevenLabs-specific fields (adjacent-text context, text normalization) and share one `findAdjacentSpeechText` helper in `@adt/pipeline`, so audio doesn't regenerate when a book crosses paths.
- **`sample_rate`/`bit_rate`**: now honored for ElevenLabs too, snapped to the nearest ElevenLabs-supported `output_format` combo (mp3/pcm/opus), matching how Azure already consumes these settings.

### UI change beyond the original scope
While addressing review feedback, we also moved **"Batch a whole page per request (experimental)"** (`batch_by_page`) out of the global Speech settings panel and **into the per-language card, shown only when that language is routed to Gemini** — mirroring how the ElevenLabs-only options already render per-language. This setting only ever applied to Gemini-routed languages, so showing it inline with the Gemini provider/model selection makes that scoping explicit in the UI. The underlying config field and behavior (`speech.batch_by_page`, gemini-only application in `stage-runner.ts`) are unchanged — this is a UI relocation only.

### i18n
All new user-visible strings extracted via Lingui and translated across `en`, `es`, `fr`, `pt-BR`, `sq` (0 missing).

### Validation
- `pnpm typecheck` — 0 errors
- `pnpm lint` — 0 errors (only pre-existing unrelated warnings)
- Targeted `vitest` runs across `packages/types`, `packages/llm`, `packages/pipeline`, `apps/api` (speech/tts), and `apps/studio` — all passing
- Verified in-browser: per-language ElevenLabs and Gemini options render correctly, only for their respective provider, and persist saved config values
- `pnpm build` + `pnpm dev` verified locally after the latest round of fixes

### Notes
- Fixed a stale help-text string in `SpeechPromptsEditor.tsx` that didn&#39;t mention ElevenLabs also ignores the `instructions`/accent-prompt field (only OpenAI/Gemini use it).
- See the [review response comment](https://github.com/unicef/adt-studio/pull/641#issuecomment-5169880391) for a full breakdown of which review items were addressed (and how) vs. deferred.


gbergengruen:
## Review: ElevenLabs TTS provider (#630)

Nice work — this follows the existing provider pattern closely and the risky part (cache-key compatibility) is handled correctly. The three new ElevenLabs fields are stripped from the hash and re-added only for `provider === "elevenlabs"` (`packages/pipeline/src/speech.ts:279-308`), with a test proving existing OpenAI/Azure/Gemini audio is **not** invalidated. **No regression to the existing providers found.**

All gates green on my machine:

| Check | Result |
|---|---|
| `pnpm typecheck` | ✅ |
| `pnpm test` | ✅ 181 files / 2250 tests |
| `pnpm lint` | ✅ |
| i18n `es` / `fr` / `pt-BR` / `sq` | ✅ no empty `msgstr`, all 24 new strings translated |

Findings below, most-severe first. Only **H1** is a genuine behaviour regression for existing books; everything else is a gap or polish item.

### 🔴 H1 — Mixed-provider books lose ~8× speech throughput

`apps/api/src/services/stage-runner.ts:3006-3012`

`ttsConcurrency` caps the **entire** per-entry TTS pass at 4 whenever *any* output language routes to ElevenLabs. `ttsWorkItems` is a single mixed-provider list (built at `:2838-2844`), so a book with 10 languages on OpenAI plus 1 on ElevenLabs drops from `effectiveConcurrency` (default 32) to 4 for **all** of them.

Suggested fix: partition `ttsWorkItems` by provider and run the two passes concurrently via `Promise.all`, each with its own concurrency — or hold a semaphore around only the ElevenLabs synth call. There's precedent for a *scoped* cap at `:2523` (`imageConcurrency`).

### 🟠 M2 — `elevenlabs_*` options are dropped on the `runFullPipeline` / CLI path

`packages/pipeline/src/pipeline-dag.ts:975-989`

`geminiTemperature` / `geminiSeed` are forwarded but the three ElevenLabs fields aren't (compare `stage-runner.ts:3049-3053`). Two consequences: the config is silently ignored on that path, and the computed `computeSpeechCacheKey` diverges from `runSpeechStep`, so audio regenerates when a book crosses paths. The comment at `:967-969` explicitly requires all three paths to resolve identically for cache-key consistency, so this breaks a documented invariant. There's also no `findAdjacentSpeechText` equivalent there — worth extracting the helper into `@adt/pipeline` and sharing it rather than duplicating.

### 🟠 M3 — Same omission in the single-item route

`apps/api/src/routes/tts.ts:799-830`

`generateEntry` passes `geminiTemperature` / `geminiSeed` but never the ElevenLabs three, while the comment at `:805-807` claims it "Mirrors stage-runner.ts so the single-item cache key matches the batch path". Only reachable via the Gemini-no-audio → ElevenLabs fallback so impact is small, but please either pass `config.speech?.elevenlabs_apply_text_normalization` or correct the comment.

### 🟠 M4 — `speech.sample_rate` / `speech.bit_rate` silently ignored for ElevenLabs

`packages/llm/src/speech.ts:506-513`

Azure consumes both (`stage-runner.ts:2660-2663`); the ElevenLabs mapper hardcodes `mp3_44100_128` / `opus_48000_128` / `pcm_24000`. Those two inputs sit in the same global Speech settings panel the user just used to select ElevenLabs, so they'll expect them to apply. ElevenLabs exposes `mp3_22050_32`, `mp3_44100_64`, `pcm_16000|22050|24000` — extend `ElevenLabsTTSConfig` the way `AzureAudioOptions` does.

Related: `speech.providers.elevenlabs.rate_limit` parses fine per the schema but is silently ignored — only `providerConfigs.gemini?.rate_limit` is ever read (`stage-runner.ts:2868-2872`). Either honour it or note the limitation in `config.yaml`.

### 🟠 M5 — No stage-runner tests for the new retry / concurrency logic

`apps/api/src/services/stage-runner.test.ts` has zero occurrences of `elevenlabs`, while the Gemini equivalents (`gemini-tts-failure` / `-retry` / `-transient` / `-cancel`, ~lines 973-1235) cover exactly the paths this PR adds. Untested: the 429/5xx exponential backoff, the retry cap, cancellation during backoff, the concurrency cap, `findAdjacentSpeechText` neighbour-skipping, and the `canReuseSpeechEntry` ↔ work-item cache-key sync. #630 asked for this coverage explicitly. The `@adt/llm` / `@adt/pipeline` / `@adt/types` unit tests you did add are good — this is the one layer missing.

### 🟡 Low / polish

- **A book-global setting is rendered per language** — `LanguageSettings.tsx:1573-1610`. `elevenlabs_use_context` / `elevenlabs_apply_text_normalization` are single global config fields, but the controls render inside *every* ElevenLabs-routed language card bound to one shared state. With 2+ such languages the same toggle appears N times and moving one moves all. Suggest moving it into the global Speech block next to the Gemini sampling box.
- **The default voice ID is duplicated three times** — `config/voices.yaml:113`, `packages/pipeline/src/speech.ts:31`, `LanguageView.tsx:379`, held together by "keep in sync" comments. `DEFAULT_ELEVENLABS_TTS_MODEL_ID` was correctly placed in `@adt/types` and imported by studio — please do the same for `DEFAULT_ELEVENLABS_VOICE_ID` and delete two copies.
- **A global `speech.voice` becomes a bogus ElevenLabs voice ID** — `packages/pipeline/src/speech.ts:78-99`. If the user set a non-generic voice (`nova`, `Kore`) and `voices.yaml` has no `elevenlabs` entry for the language, that name is sent verbatim as a voice ID → opaque 404/422. Consider falling back to the ElevenLabs default with a warning.
- **Provider name renders as "Elevenlabs"** — `LanguageView.tsx:1578` uses `<span className="capitalize">{speechSummary.provider}</span>`. Use a `PROVIDER_LABELS`-style lookup.
- **Onboarding key screen wasn't updated** — `apps/studio/src/components/onboarding/scenes/ApiKeyScene.tsx:18-26` duplicates `ApiKeyDialog`'s tab list, so a new user can't enter an ElevenLabs key during onboarding (only later via the gear icon).
- **`config.yaml` speech docs are Gemini-only** — the commented block at `config.yaml:324-346` is where users discover these settings; add an `elevenlabs` example plus the two new options and a note about the concurrency cap.
- **No `[elevenlabs-tts]` request logging** — Azure logs the outbound request and byte count (`packages/llm/src/speech.ts:329/344/349`); ElevenLabs logs nothing, so debugging a bad voice ID rests entirely on the thrown message. Principle 4 (Maximum Transparency) argues for parity.
- **Backoff has no jitter** — `stage-runner.ts:3077-3080`. Up to 4 workers that hit a concurrency 429 together retry in lockstep at 2s/4s/8s. ±20% jitter would de-synchronise them.
- **Stale help text** — `LanguageSettings.tsx:1467` still reads "Only affects languages routed to Gemini — OpenAI and Azure ignore these". Any edit needs `pnpm --filter @adt/studio extract` + 4 translations, so batch it with the UI move above.
- **`BookCreationWizard.tsx:327-334`** passes `azure` / `geminiApiKey` but not `elevenLabsApiKey`. Harmless today (extract-only kickoff) but asymmetric.
- **Unrelated JSX reflow** — the `batch-by-page` switch was moved out of the "Gemini voice consistency" `<div>` into a sibling. Looks like a legitimate nesting fix, but it's unrelated to #630; worth a line in the PR description so reviewers don't hunt for its purpose.
- **Pre-existing, but ElevenLabs makes it more likely:** a 200 with an empty body gets written to both `audio/` and `.cache/` (`packages/pipeline/src/speech.ts:470-478`) and then satisfies the `fs.existsSync(cachePath)` reuse check at `:439` forever. Gemini guards against this; OpenAI, Azure and now ElevenLabs don't. A one-line `if (audioBytes.byteLength === 0) throw` in `generateSpeechFile` closes it for all four.

### ✅ Things I specifically checked and found correct

- **The retry loop can't spin.** `attemptCount++` is at the top of `while (true)` (`:3033-3034`), so the ElevenLabs `continue` is bounded at 6 attempts, and cancellation is checked before and after each sleep.
- **`elevenLabsApiKey!` at `tts.ts:825` is unreachable with an undefined key** — the ElevenLabs attempt is only pushed when the trimmed header is truthy (`:296`). Same shape as the existing `azureSpeechKey!`.
- **`ELEVENLABS_API_KEY` env fallback is at parity** with `GEMINI_API_KEY` at all three construction sites.
- **Format mapping is sound** — `wrapPcmAsWave` at 24 kHz matches `pcm_24000`; the mp3/opus values are valid ElevenLabs `output_format`s, and the `wav` → PCM → RIFF round-trip is covered by a test.
- **Packaging and word-timestamps are provider-agnostic** — nothing to add there; Whisper alignment works on ElevenLabs output unchanged.
- **The page-batched path** uses uncapped `effectiveConcurrency` but is Gemini-only by construction (`:2733-2736`) and runs before the per-entry pass, so no ElevenLabs request escapes the cap.
- **No docs enumerate the TTS providers** (checked `docs/`, `docs/wiki/`, `README.md`), so #630's docs bullet is a no-op beyond the `config.yaml` item above.

### Two behaviours worth confirming are intentional

1. **ElevenLabs-routed languages get no per-item regeneration.** `tts.ts:742` rejects anything but Gemini and `LanguageView.tsx:3226` gates the button on `hasGeminiKey`; ElevenLabs only ever appears as a *fallback* when Gemini returns no audio. Same limitation as OpenAI/Azure, so not a regression — but given ElevenLabs 429s are concurrency-based and fairly likely, the "generate the missing ones one by one" recovery path isn't available to these users.
2. **Failed ElevenLabs items emit a per-item `step-error`** (`:3213-3219`) rather than the friendly "N items failed, generate them one by one" summary Gemini gets (`:3244-3254`). Consistent with OpenAI/Azure, so intentional as far as I can tell, just noting the asymmetry.

### Suggested merge order

**Before merge:** H1, M2/M3 (or corrected comments), M5.
**Follow-up PRs:** M4, then the Low items — the shared `DEFAULT_ELEVENLABS_VOICE_ID`, the empty-audio guard and the "Elevenlabs" capitalisation are the cheapest wins; batch the per-language UI move + the stale help text into one `extract` + translation pass.

gaguerre-iugo:
Thanks for the thorough review! Status on each item below — explicitly noting where we followed your suggested approach as-is.

### 🔴 H1 — Mixed-provider concurrency
✅ **Followed your suggestion exactly.** Partitioned `ttsWorkItems` by provider in `stage-runner.ts` and ran the ElevenLabs / non-ElevenLabs passes concurrently via `Promise.all`, each with its own concurrency cap — precisely the fix you proposed. Non-ElevenLabs books/languages now keep full `effectiveConcurrency`.

### 🟠 M2 — `elevenlabs_*` dropped on the CLI/`runFullPipeline` path
✅ **Followed your suggestion exactly.** Extracted `findAdjacentSpeechText` into `@adt/pipeline` (shared instead of duplicated, as you suggested) and wired all three ElevenLabs fields into `pipeline-dag.ts`'s `tts` executor to match `stage-runner.ts`, so the cache key resolves identically on both paths.

### 🟠 M3 — Same omission in the single-item route
✅ **Followed your suggestion (and went with the fuller of the two options you offered).** You said "either pass the field or correct the comment" — we did both: passed all three ElevenLabs fields (not just `elevenlabs_apply_text_normalization`) via the shared helper, and corrected the misleading "mirrors stage-runner.ts" comment.

### 🟠 M4 — `sample_rate`/`bit_rate` ignored for ElevenLabs
✅ **Followed your suggestion, adapted for ElevenLabs' constraints.** You said "extend `ElevenLabsTTSConfig` the way `AzureAudioOptions` does" — we did that (`ElevenLabsAudioOptions`), but since ElevenLabs only accepts a fixed set of `output_format` combos (unlike Azure's free-form values), we added nearest-match snapping instead of pass-through. Wired through `stage-runner.ts` and `pipeline-dag.ts`.

For `rate_limit`: ✅ **followed one of your two suggested options exactly** ("either honour it or note the limitation") — we chose to document it, since ElevenLabs' real constraint is concurrent connections, not requests/minute, so reusing the RPM-shaped field would be misleading. Added JSDoc + a `config.yaml` note.

### 🟠 M5 — No stage-runner tests for ElevenLabs retry/concurrency
⏭️ **Not addressed in this pass** — deferring to a follow-up.

### 🟡 Low / polish
- **Default voice ID duplicated 3×** — ✅ **followed your suggestion exactly.** `DEFAULT_ELEVENLABS_VOICE_ID` moved to `@adt/types` (same pattern you pointed to with `DEFAULT_ELEVENLABS_TTS_MODEL_ID`), both code copies deleted.
- **Provider renders as "Elevenlabs"** — ✅ **followed your suggestion exactly** ("Use a `PROVIDER_LABELS`-style lookup"), and went a bit further: shared the same lookup between `LanguageSettings.tsx` and `LanguageView.tsx` instead of adding a second local copy.
- **Book-global setting rendered per language** (`elevenlabs_use_context`) — ↔️ **different approach than suggested.** You proposed moving it to the global block; we instead moved `batch_by_page` (previously global) *into* the per-language card for Gemini, mirroring the ElevenLabs per-language pattern rather than the other way around. Kept ElevenLabs settings per-language as-is for now.
- Everything else in this section (bogus voice ID fallback, onboarding key screen, full `config.yaml` example, request logging, backoff jitter, stale help text, `BookCreationWizard` key, empty-audio-body guard) — ⏭️ **not addressed**, by choice, in this pass.

### Two behaviours confirmed intentional
Confirmed both as accepted, consistent with existing OpenAI/Azure behavior — not a regression from this PR. Leaving as-is, no code changes.

---
All items marked ✅ above verified with `pnpm typecheck`, `pnpm lint`, and the relevant `vitest` suites (green), plus a fresh `pnpm --filter @adt/studio extract` pass (0 missing translations across es/fr/pt-BR/sq).

gbergengruen:
Hey @gaguerre-iugo — heads up on two things before this goes in.

**1. I merged `develop` into this branch.** It had drifted 28 commits behind, and `develop` has moved a lot in the storyboard/sectioning/activity areas since you opened this. The merge was clean (no conflicts), and I verified the merged tree: `pnpm typecheck` passes, 2494 tests pass across 199 files, `pnpm lint` is at 0 errors, and `lingui extract` reports 0 missing strings across es/fr/pt-BR/sq. Doing it here rather than at merge time keeps the eventual merge back into `develop` a fast-forward-ish no-brainer.

**2. I also pushed a set of fixes on top of your work.** Thanks for building this out — the provider abstraction and routing were the right shape, and everything below was found by reviewing the three execution paths against each other rather than by anything being structurally wrong. Each commit message has the full reasoning; the biggest ones:

**Correctness / would-have-failed-in-production**

- **`voice_settings` was never sent.** ElevenLabs documents this field as overriding whatever the voice has stored in their dashboard, so omitting it handed delivery control to arbitrary stored values. In practice this surfaced as filler sounds ("ehm", "uh") the source text never contained — and there's no prompt-side fix, since ElevenLabs has no negative prompting. Now always sent, defaulting to their own audiobook/narration recommendation.
- **The CLI / `runFullPipeline` DAG had no ElevenLabs protections.** One pool at the default concurrency of 32, against plan ceilings of ~4–5 concurrent requests, no retry — so a CLI run against ElevenLabs mostly failed. Now partitions by provider like the API stage-runner does.
- **Retry classification matched `/try again/i` against free response text**, retrying permanent 4xx five times and burning ~60s of backoff on requests that could never succeed.
- **Single-item regeneration was impossible for every provider but Gemini** — `generate-one` demanded `X-Gemini-API-Key` and hard-failed on `provider !== "gemini"`, even though the synthesizer ternary below already handled all four.
- **The default voice didn't match between config and code** — `voices.yaml` had the community clone while `DEFAULT_ELEVENLABS_VOICE_ID` said Rachel, and config wins in `resolveVoice`. Kept your ID under `es-uy` since it's clearly chosen for the Uruguayan book.
- **An unsupported `speech.format` fell through to mp3**, writing mp3 bytes into `.ogg` files.
- **Adjacent-text context pulled the wrong text** — raw vs. emoji-stripped, and it scanned across the Easy Read variants the stage-runner appends.

**Transparency and UI honesty** — voice IDs now resolve to names (new `/v2/voices` proxy + combobox on the existing `FontFamilyCombobox` pattern, no new dep); per-provider settings no longer render per-language when they're stored book-wide; the debug Log tab records effective request settings; two silent traps (v2.5 Enterprise normalization, English-default fallback) now warn at edit time.

**Two bugs that predate this PR but surfaced here** — a Lingui macro passed `t` as a function argument wasn't transformed, blanking the whole label column (the test was complicit, mocking `useLingui` with a working stub); and a locale case mismatch (`es-uy` vs `es-UY`) broke per-language display for *every* provider.

Nothing here needs action from you. Flagging it mainly so the diff isn't a surprise, and because two decisions are judgement calls I'd happily revisit: the 2-retry budget for single-item regeneration (shorter than the batch paths' 5, since it runs inside an HTTP request), and my choice *not* to add per-locale ElevenLabs voice IDs to `voices.yaml` — an unverified ID 400s at synthesis, which seemed worse than an English-accent fallback.

Issue #627, opened 2026-07-28
[Sectioning-Storyboard] Cross-page merge does not reliably auto-rerender affected sections


elasticsounds:
**Priority:** P0
**Reported by:** Ninas Gama

## Problem

After a cross-page merge, the affected sections do not consistently auto-rerender. The user sometimes has to trigger a rerender manually, and there's no indication that the merged content is stale.

## Options considered

1. **Automatic rerender** of the affected target/source section immediately after merge.
2. **Post-merge action banner** with one-click rerender.

## Decision: post-merge action banner

Automatic rerender is rejected — re-rendering the storyboard can block the interface. A banner lets the user perform several cross-page merges across multiple sections and then rerender **once**, at a moment of their choosing.

## Proposed behavior

- After a cross-page merge, mark the affected target and source sections as stale.
- Show a persistent post-merge banner indicating which sections need rerendering.
- Provide a single-click action to rerender all pending sections in one pass.
- Accumulate across multiple merges rather than prompting per merge.
- Dismiss/clear the banner once the rerender completes.

## Acceptance criteria

- [ ] Affected sections are flagged as stale immediately after any cross-page merge
- [ ] Banner appears and persists until the user rerenders or dismisses it
- [ ] Multiple merges accumulate into a single rerender action
- [ ] One-click rerender updates all pending sections
- [ ] Interface is not blocked between merge and rerender

## Open questions

- Should the banner be dismissible without rerendering, and if so, does the stale flag persist?
- Is there a point (e.g. export, save, navigating away) where a rerender should be forced?

gbergengruen:
E, PR #642 fixes this. In that PR, merging, splitting and other actions put the following stages into a stale state. This is consistent to other things we do in other stages. Does this PR fix this bug?

elasticsounds:
**Fixed by #642** (merged into `develop` 2026-08-03) — closing.

Answering the open question on this thread (*"Does PR #642 fix this bug?"*): **yes**, the underlying defect is resolved, though via a different mechanism than the one proposed above. Worth being explicit about what did and didn't ship so we can decide whether any follow-up is wanted.

### What the bug actually was

After a cross-page merge, the merged content wasn't reliably re-rendered and there was no signal that it was stale — so a merged section could silently ship the old (or empty) content to the exported book while the stage still read *"done"*.

### How it's fixed

Writing `page-sectioning` now marks the **storyboard stage and everything downstream** as needing a re-run, and empties both affected pages' renderings:

[...]

PR #642, opened 2026-07-29
fix(sectioning): unsaved-changes guard, downstream reset warnings, structural-op confirmations

### LINKED ISSUE #622: Unsaved changes alert in Sectioning
### Preliminary Checks

### Problem / motivation

In "sectioning," nothing happens when you exit without saving—unlike in "extract" or "storyboard," where an "unsaved changes" alert pops up.

### Proposed solution

Add a "Unsaved changes" popup alert in "Sectioning"

### Alternatives considered

_No response_

Pipeline — Sectioning

### Additional context

_No response_

gbergengruen:
**Fixes #622** — "Unsaved changes alert in Sectioning".

> Note: this PR targets `develop`, and GitHub only auto-closes linked issues when
> a PR merges into the **default** branch. #622 will therefore need closing by
> hand, or it will close when `develop` reaches `main`.

The issue asked for one thing: an "unsaved changes" alert in Sectioning
(bug 1 below). Implementing it surfaced a chain of related defects — several of
them silent data-loss paths, including sectioning edits never reaching the
exported book — which are all fixed here. **11 bugs total**, in five
self-contained commits.

### 1. Sectioning discarded pending edits silently (the reported issue)

`SectioningPageDetail` tracked its own dirty state but never registered with the
shared floating-save registry, so `hasUnsaved` stayed false and **none** of the
three guards fired: the TanStack route blocker, the browser `beforeunload`
prompt, and the Electron quit dialog. Extract and Storyboard all register; only
Sectioning did not.

Also fixed: the editor was the only stage rendering its own Save/Discard buttons
in the step header, against the convention documented in `VersionPicker`
("Save/Discard always live in the floating bar").

### 2. The Overview toggle dropped edits with no warning

`overviewMode` is local component state, so switching to Overview unmounted the
editor **without a route change** — the blocker could never see it. Now blocked
while edits are pending.

### 3. "Save & leave" could navigate away after a *failed* save

`UnsavedChangesGuard` proceeds if the save promise resolves. Sectioning's save
caught its error into local state, so the promise resolved even on failure — the
guard would navigate away and the edits were lost. The save now rethrows, and
the `onSaveStay` contract documents that it must.

### 4. `willRerun` was keyed off the wrong field

`willRerun` tested `onSaveStay`, so any surface that saves in place without
queueing a re-run was mislabelled **"Save & Re-run"**. Now keyed off
`onSaveAndRerun`. All three existing providers pass both fields, so settings
surfaces are unchanged (verified live).

### 5. The unsaved-changes dialog degraded mid-save

The dialog read its dirty entries live. Each surface clears its entry as soon as
it persists, but the blocked navigation only resumes once every save resolves —
so for the duration of the save the dialog visibly became a generic,
button-less **"HEADS UP"** with no pending-edit chips. It now renders from the
last populated snapshot.

### 6. The save promise waited on cache refetches, so that window lasted *seconds*

The save `await`ed `invalidateQueries`, which only settles once the refetches
complete. That held the (degraded) dialog open for seconds after the data was
already safe. The invalidations now run in the background; the PUT is still
awaited and still rethrows. Verified: dialog close time went from several
seconds to **~680ms**.

### 7. Sectioning edits never reached the exported book

`web-rendering` is a materialized snapshot of the sectioning text, and every
later output is re-derived from that HTML:

```
page-sectioning → web-rendering (HTML) → text-catalog → easy-read / translate / tts → package
```

The sectioning save cleared only the `IMAGE_SET_CHANGE_*` set, leaving
**storyboard, quizzes, glossary and toc** reporting `done` against stale input.
Clearing `text-catalog`/`tts`/`translation` only made the cascade *look*
handled — they were rebuilt from the stale HTML and landed on the old text
again. So a text edit was silently dropped from translations, audio and the
exported bundle.

### 8. The structural ops were worse — sections vanished from the output

A **split** deletes both halves' HTML (`filter(s => s.sectionIndex !== idx)`)
and a **cross-page merge** empties both pages, yet the storyboard stage still
reported `done`. Since packaging reads `web-rendering` per section, those
sections were silently **missing from the exported book**.

Fixed for the save and all five structural ops: writing `page-sectioning` now
marks the storyboard-and-later chain as needing a re-run. **Non-destructive** —
only `step_runs` are cleared, so renderings and their version history survive
and manual storyboard edits elsewhere in the book are not destroyed. A
`web-rendering` save is the storyboard's own output and does not mark storyboard
stale.

### 9. Split and Duplicate applied instantly with no confirmation

Merge and Delete confirmed; **split and duplicate just did it**. None of the
five mentioned that they invalidate downstream stages. All five (plus the save)
now share one `CascadeResetDialog` with op-specific titles and icons.
**Cancelling applies nothing** — previously it was possible to save the edits
and then decline the follow-up, leaving the book saved-but-inconsistent.

Nothing is ever re-run automatically; the user re-runs Storyboard when ready.

### 10. Split was clickable while dirty, failing with an easily-missed error

The dropdown ops were properly disabled when there were unsaved edits, but the
split control was not — it wrote a small error into the step header instead. Now
disabled with the "Save or discard your edits first" tooltip.

### 11. `CascadeResetDialog` was hardcoded for re-run flows

Its confirm button always showed a `Play` icon (wrong for "Delete section"),
could not be disabled, and always rendered the "Will be reset" block — showing
a misleading "Will be reset · 0" when nothing was affected. Three additive,
backwards-compatible props fix all three.

## Deliberately out of scope (worth separate issues)

- **`two_column` renders as a raw slug in the Storyboard strategy picker.**
  `STRATEGY_LABEL_MSGS` has no entry for it and `strategyDisplayName` falls back
  to the slug. It *is* the shipped default (`config.yaml: default_render_strategy:
  two_column`) but the wizard marks it `hidden: true`, so there is no card for it
  either — switching away is a one-way door. Separately, the picker lists
  `single_column` while the config key is `one_column`, so that card never
  appears. Pre-existing; unrelated to this branch.
- **Storyboard bounces to its landing page once marked stale.** `StoryboardIndex`
  renders the view only when the stage is `done`, so after a sectioning save the
  editor is replaced by the "Run" landing page even though the renderings are
  intact. Gating on rendering data existing (as `SectioningView` does with
  `hasPageData`) would keep it reachable.
- **`SectioningOverview` is not covered.** It PUTs sectioning from six places
  with immediate mutations and no dirty state, so it has the same staleness gap.
- **Section `isPruned` → true is the one edit that needs no re-run** (packaging
  and text-catalog skip pruned sections at read time); we still warn.
- **Fixed-layout books**: sectioning text edits never reach output even after a
  full re-run, because `web-rendering` is rebuilt from `positioned-text`. The
  reset warning is suppressed for them rather than lying; the real bug is that
  the text is editable there at all.

## Verification

- `pnpm typecheck` clean, `pnpm lint` **0 errors** (17 pre-existing warnings,
  none in changed files), **2144 tests / 173 files pass**.
- i18n: 5 new strings, all translated in `es`, `fr`, `pt-BR`, `sq` (0 missing);
  obsolete strings removed; stale-key count unchanged from `develop`;
  `compile --strict` passes.
- New tests: `floating-save.test.tsx`, `UnsavedChangesGuard.test.tsx`,
  `CascadeResetDialog.test.tsx` (first tests for all three modules), plus three
  API tests covering the staleness marking, the `web-rendering` exemption, and
  split specifically.
- Driven end-to-end in the running app with Playwright: guard dialog attribution
  and copy, save/cancel/discard paths, failed-save handling, `beforeunload`,
  split disabled-while-dirty, split cancel being a true no-op, split confirm
  flagging the storyboard, and **zero `/stages/run` calls** across every flow.

### Not verified

- The Electron quit dialog was exercised only via the web `beforeunload` path,
  which shares `hasUnsaved`; the desktop-specific path needs a packaged run.
- No storyboard edit was driven end-to-end (`willRerun` is provably unchanged
  there — it passes neither field — and both branches are unit-tested).


ksokolovic:
## Follow-up review update

We addressed all five findings from the latest review round in separate commits:

- e8b97393 — reject Sectioning mutations while pipeline work is active, preventing stale task completion from restoring invalid downstream state.
- 0adcc3bb — disclose the completed downstream stages that Save & leave will reset.
- 23ad900f — share and await the active Sectioning save promise so navigation cannot proceed before the PUT finishes.
- a1bc28a6 — lock the section-type selector and tree editor for the full duration of structural operations.
- 48e995c7 — freeze unsaved-dialog attribution, pending edits, and actions while navigation is blocked and multi-entry saves settle.

### Verification

- Node 24 local suite: 185 test files and 2,260 tests passed.
- All 65 directly affected tests passed.
- Typecheck passed.
- Lint passed with 0 errors; only 9 unrelated existing unused-suppression warnings remain.
- GitHub test, docker, and i18n checks passed on head 48e995c7.
- GitHub reports the PR clean and mergeable.

No remaining merge blockers were identified in this review scope.

Issue #627 remains open intentionally. This PR now marks the Storyboard pipeline stale and warns about downstream resets, but it does not implement the persistent post-merge banner, accumulated section-specific stale tracking, or one-click rerender workflow described by that issue.

elasticsounds:
Found an issue caused by this PR that invalidates the entire storyboard stage if you hit save on the storyboard for any change

PR #646, opened 2026-07-30
feat(reader): responsive mobile reader + preview device viewport

Eliezir:
## Summary
Makes the packaged **book reader** (`apps/adt-runtime`) fully usable on mobile phones — every feature is reachable and the chrome adapts to small screens — and makes the Studio **preview** render at the device viewport selected in the storyboard instead of always defaulting to desktop.

Previously the reader's dock and its panels were desktop-only: two `min-w-64` sections plus five tool icons overflowed a phone, and the feature panels rendered in fixed-width popovers that were clipped off-screen.

### Studio — preview inherits the device viewport
- The preview/"Run" pane now reads the same persisted device selection as the storyboard editor and renders mobile/tablet inside real device chrome (scaled to fit), instead of always desktop.
- Adds a surface-styled `ViewportToggle` variant matching the preview's cards.

### Reader — mobile-responsive chrome
New `useIsMobile` hook (`sm`/640px breakpoint); desktop layout unchanged.
- Compact mobile dock bar: `Contents · ◀ 3/24 ▶ · Tools`, with all features behind a **Tools bottom sheet** (44px touch targets preserved).
- Feature panels (TOC, Glossary, Language, Settings) present as **bottom sheets** on mobile instead of clipped popovers, and scroll correctly (mobile panels get a definite height so their inner `ScrollArea` bounds and scrolls).
- The bottom-sheet **overlay is fully transparent, no blur** — the book stays visible behind the sheet (a tap still dismisses it).
- **Read-aloud (TTS)** is a **non-modal floating bar** — no backdrop, so the book stays visible and interactive while it reads.

### TTS behavior
- TTS controls **stay open across page navigation** while read-aloud is on (driven by the persisted `readAloud` state, not the transient dock value); **Stop** or the **TTS toggle** turn it off and hide it. Works on desktop and mobile.

### Bug fixes
- Inline TTS bar was mis-placed (off-screen) when the dock sits at the top — it was trapped by the dock's `will-change: transform` containing block. Now portaled to the chrome container so `fixed` resolves against the viewport, offset from top/bottom by dock position.
- Opening the TTS **speed/volume menus scrolled the page back to the top** — they used `absolute` positioning while anchored inside the `fixed` bar. Both now use `positionMethod="fixed"`.

## Testing
- `pnpm test`: **2062 passing** (164 files).
- Repo typecheck clean; runtime typecheck only 8 pre-existing errors (unrelated files); studio lint 0 errors.
- Playwright E2E (mobile + desktop, 0 console errors): TTS persistence across nav + stop/hide; panels as sheets with working scroll (TOC, a 19,677px Glossary list, Settings); transparent overlay (book visible); speed/volume no scroll-jump; inline bar placement for top/bottom dock; desktop regression (icon row + popovers, persistent TTS bar); glossary term popover stays in-viewport; modal layering.
- Verified live on a physical phone over LAN.


Eliezir:
Splitting this into two focused PRs: one for the mobile reader UI, one for the preview device viewport.

PR #657, opened 2026-07-31
Fix preview refresh after sign-language video changes

### LINKED ISSUE #655: Preview does not refresh after sign-language video changes
## Problem

Uploading, assigning, reassigning, or deleting a sign-language video updates the book storage and the Sign Language stage, but the packaged ADT Preview can remain stale. The preview therefore may not include the updated `videos.json` mapping, and the Sign Language hand icon is absent even though a video is assigned to the current section.

## Steps to reproduce

1. Open a processed book and go to the Sign Language stage.
2. Upload a supported video and assign it to a Storyboard section.
3. Open Preview and navigate to that section.
4. Observe that the Sign Language hand icon may be missing because Preview is still serving the previous packaged bundle.

## Expected behavior

Sign-language video upload, assignment/reassignment, and deletion invalidate the packaged ADT status and cause Preview to rebuild/reload so its video manifest reflects the latest book state.

## Actual behavior

Only the sign-language-video query is invalidated; Preview can continue serving a stale package.

## Acceptance criteria

- Upload, assignment/reassignment, and deletion invalidate Preview/package state.
- An open Preview is repackaged after a video mutation.
- Navigating to Preview after a mutation cannot reuse stale package status.
- Regression tests cover the invalidation/refresh behavior.

KaReeeeeeeeEM:
## Summary

- refresh the sign-language video query after every video mutation
- invalidate packaged ADT preview status after upload, assignment/reassignment, and deletion
- dispatch the existing `adt:repackage` event so an open Preview rebuilds immediately
- add regression coverage for all three mutation hooks

Closes #655

Supersedes #656, which GitHub automatically closed when its source branch was renamed.

## Verification

- `pnpm exec vitest run apps/studio/src/hooks/use-sign-language-videos.test.tsx`
- `pnpm typecheck`
- `pnpm --filter @adt/studio lint` (passes with 9 pre-existing unused-disable warnings)

### DIFF
--- a/apps/api/src/routes/package.test.ts
+++ b/apps/api/src/routes/package.test.ts
@@ -204,6 +204,63 @@ describe("Package routes", () => {
       expect(JSON.parse(fs.readFileSync(configPath, "utf-8")).bundleVersion).toBe(secondBody.version)
     })
 
+    it("repackages when a sign-language video assignment changes", { timeout: 20_000 }, async () => {
+      const label = "book-sign-language-cache"
+      createRenderedBook(label)
+      createWebAssets()
+
+      const storage = createBookStorage(label, tmpDir)
+      storage.putSignLanguageVideo(
+        "video-1",
+        Buffer.from("video-content"),
+        "video.mp4",
+        "video/mp4",
+      )
+      storage.close()
+
+      const packageBook = async () => {
+        const response = await app.request(`/api/books/${label}/package-adt`, {
+          method: "POST",
+        })
+        expect(response.status).toBe(200)
+        return await response.json() as { version: string }
+      }
+      const videosPath = path.join(
+        tmpDir,
+        label,
+        "adt",
+        "content",
+        "i18n",
+        "en",
+        "videos.json",
+      )
+      const configPath = path.join(tmpDir, label, "adt", "assets", "config.json")
+
+      const unassignedPackage = await packageBook()
+      expect(JSON.parse(fs.readFileSync(videosPath, "utf-8"))).toEqual({})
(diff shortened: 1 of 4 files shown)


ksokolovic:
### Follow-up: packaging cache fix

Review found that frontend invalidation alone was insufficient: assigning or reassigning a video only changes its SQLite `sectionId`, while the packaging cache previously fingerprinted video files only. The package endpoint could therefore reuse a stale bundle.

Commit `32b37905`:

- includes stable sign-language assignment metadata in the packaging hash;
- adds an API-level regression covering unassigned → assigned → unassigned;
- verifies package versions change and `videos.json`/`features.signLanguage` reflect each state.

Full CI is green: tests, typecheck, Docker build, lint, and i18n.

gbergengruen:
Quick question, wouldn't the inmediate repackaging be contradictory about other invalidation of other stages we already have? I wouldn't re-package the preview until the user manually does so. @ksokolovic

ksokolovic:
Valid point. The cache fix in `32b37905` does not itself trigger packaging—it only ensures that the next package request detects assignment/reassignment changes instead of returning the stale cached bundle.

Immediate repackaging comes from the original `adt:repackage` event and was added to satisfy #655’s acceptance criterion that an open Preview refreshes after a mutation. It only has an effect while `PreviewView` is mounted. With the current single-view routing, normal Sign Language/Glossary edits simply invalidate state, and entering Preview already triggers packaging automatically.

If the intended product rule is manual-only, we can remove the event dispatch and update #655’s acceptance criteria accordingly. The packaging-hash fix should remain either way, since otherwise even a manual repackage can reuse stale output.

Should we keep the behavior specified in #655, or align this with the manual invalidation behavior used elsewhere?

PR #658, opened 2026-07-31
fix(storyboard): make saved reading order authoritative

### LINKED ISSUE #227: [Storyboard] Simple re-ordering support for text and images

### LINKED ISSUE #250: [Storyboard] Add controls to the storyboard to move objects back or next.

ksokolovic:
## Summary

Storyboard users can save a reordered section tree, but an LLM rendering or visual-review pass can still restore the source PDF's apparent order. This is especially likely for overlay rendering, whose previous instructions treated the PDF's top-to-bottom arrangement as authoritative.

This PR makes the persisted content tree the unconditional source of truth for reading order. Generated HTML must preserve the tree's depth-first text-and-image leaf order, regardless of whether the render is entered from drag-and-drop, a later rerender, cache reuse, or another workflow.

## Why this supersedes #573

#573 models a reorder as transient intent and threads an `intentionalReadingOrder` flag through Studio, API, and pipeline layers. That makes correct behavior conditional on every entry point preserving the flag.

This implementation replaces that conditional state with a stable invariant:

> Once a section tree is saved, its leaf order is authoritative.

The source PDF remains the reference for styling and composition, but it cannot override persisted content order. A deterministic validator enforces the invariant after generation, so prompt compliance is verified rather than assumed.

## Changes

- Instruct standard and overlay web generation to emit text and image elements in saved tree order.
- Instruct overlay generation to move positioned groups when necessary instead of restoring their former PDF positions.
- Give visual review the authoritative content tree and prevent it from undoing saved order while it continues to assess clipping, overlap, readability, responsiveness, and visual quality.
- Collect the non-pruned leaf `data-id` sequence from the render tree.
- Validate generated text and image elements against that sequence in DOM order, while preserving existing rules for optional text and omitted images.
- Add editor regressions for native drag/drop handlers covering:
  - an individual image moved between text groups;
  - an image-and-caption container moved as one reading-order unit.
- Add prompt and HTML-validation regressions for standard, overlay, and visual-review paths.

### Representative storybook

Verified a paired `Dog image → Rescue → Snake image → Sense` reorder through:

- section save and Studio reload;
- fresh model rerender;
- cached rerender;
- API-process restart;
- desktop, tablet, and mobile previews;
- DOM and screen-reader order;
- packaged ADT output.

The saved tree, generated HTML, accessible order, and packaged HTML all matched.

### Independent textbook and model

Ran pages 7-8 of `weather-climate.pdf` from Extract through Storyboard using `openai:gpt-5.4-mini`. After moving a complete image-and-caption group below the main text, the rerendered section matched all 14 saved leaf IDs in exact DOM order. Local LLM audit metadata confirmed the alternate model was used.

## Automated verification

- `pnpm typecheck` — passed.
- `pnpm lint` — passed with 0 errors (9 pre-existing unused-disable warnings).
- `pnpm test` — build passed; 2,212 tests passed. Five `part-service.test.ts` cases reached their existing 5-second timeout under full parallel load.
- `pnpm exec vitest run apps/api/src/services/part-service.test.ts` — all 13 tests passed in isolation.
- New drag/drop component regressions — 2/2 passed.

## Reviewer check

The native HTML5 drag/drop handlers are covered by component-level browser events. A brief physical Chrome drag smoke test is still recommended because the available in-app automation driver cannot synthesize this browser drag implementation reliably.

Fixes #227
Fixes #250
Supersedes #573

### DIFF
--- a/apps/studio/src/components/section-tree-editor/SectionTreeEditor.test.tsx
+++ b/apps/studio/src/components/section-tree-editor/SectionTreeEditor.test.tsx
@@ -1,0 +1,224 @@
+// @vitest-environment jsdom
+import React, { useState } from "react"
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest"
+import { cleanup, fireEvent, render, screen, waitFor } from "@testing-library/react"
+import type { PageSectioningSection } from "@adt/types"
+import { SectionTreeEditor, TREE_DRAG_TYPE } from "./SectionTreeEditor"
+
+vi.mock("@lingui/react/macro", () => ({
+  useLingui: () => ({
+    t(strings: TemplateStringsArray, ...values: unknown[]) {
+      return strings.reduce(
+        (text, part, index) => text + part + (index < values.length ? String(values[index]) : ""),
+        ""
+      )
+    },
+  }),
+}))
+
+vi.mock("@/components/ui/action-menu", () => ({
+  ActionMenu: () => null,
+}))
+
+const initialSection: PageSectioningSection = {
+  sectionId: "section-1",
+  sectionType: "standard",
+  backgroundColor: "#ffffff",
+  textColor: "#000000",
+  pageNumber: 1,
+  isPruned: false,
+  nodes: [
+    { nodeId: "snake-image", role: "image", isPruned: false },
+    {
+      nodeId: "rescue-group",
+      structure: "group",
+      isPruned: false,
+      children: [
+        { nodeId: "rescue-text", role: "text", text: "Rescue", isPruned: false },
+      ],
+    },
+    {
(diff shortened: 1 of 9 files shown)


gbergengruen:
This looks great to me. Let's close #573 and merge this. Thanks!

PR #659, opened 2026-07-31
fix(storyboard): resolve AI-edit model id with the standard fallback chain

gbergengruen:
## Problem

Storyboard AI edits fail on any book whose config doesn't set `page_sectioning.model` explicitly — which is the default, since the root `config.yaml` only sets `prompt` and `max_refinements` under that key.

Observed on `cuaderno3`:

```
POST /api/books/cuaderno3/pages/pg008/sections/0/ai-edit
--> POST /api/books/cuaderno3/pages/pg008/sections/0/ai-edit 200 18ms
[LLM] web-rendering pg008 | error (attempt 1/4) | Cannot read properties of undefined (reading 'indexOf') | retrying in 1072ms
...
[LLM] web-rendering pg008 | error (attempt 4/4) | Cannot read properties of undefined (reading 'indexOf')
```

## Root cause

`aiEditSection` in `apps/api/src/services/page-edit-service.ts` resolved its model id with a ternary that branched on the **presence of the `page_sectioning` section**, not on whether that section has a **`model` key**:

```ts
const modelId = (config as Record<string, unknown>).page_sectioning
  ? ((config as Record<string, unknown>).page_sectioning as Record<string, unknown>).model as string
  : "openai:gpt-4o"
```

With `page_sectioning` present but no `model`, `modelId` was `undefined`. `resolveModel` in `packages/llm/src/client.ts` then does `modelId.indexOf(":")` → `Cannot read properties of undefined (reading 'indexOf')`. The failure is deterministic, so all 4 attempts failed and the `ai-edit` task died — while the HTTP POST had already returned `200` (task accepted), leaving the user with an edit that silently never applied.

The `as string` cast is what let this past `tsc`.

## Fix

Use the same fallback chain every other thoughtful LLM step already uses (`page-sectioning.ts`, `translation.ts`, `easy-read.ts`, `glossary.ts`, `toc-generation.ts`, …):

```ts
const modelId =
  config.page_sectioning?.model ?? config.default_model ?? DEFAULT_LLM_MODEL_ID
```

Dropping the two `Record<string, unknown>` casts means the typed `AppConfig` schema guards this class of bug going forward. Note the old literal fallback was also stale (`openai:gpt-4o`); the chain now honours the book's `default_model`.

## Testing

- Added a regression test that builds a config with `page_sectioning` but no `model` and asserts `createLLMModel` receives a defined model id. Verified it fails on `develop` (`modelId: undefined`) and passes with the fix.
- `apps/api/src/services/page-edit-service.test.ts` — 4/4 pass.
- `pnpm typecheck` clean.

## Follow-up (not in this PR)

This failed for ~30s of retries with nothing surfaced in the storyboard UI beyond the edit not applying. Worth a separate look at how `ai-edit` task errors propagate to `StoryboardSectionDetail`.

### DIFF
--- a/apps/api/src/services/page-edit-service.test.ts
+++ b/apps/api/src/services/page-edit-service.test.ts
@@ -314,5 +314,62 @@ describe("page-edit-service", () => {
       expect(capturedHtml).not.toContain("img-first")
       expect(result.html).toContain("img-dup")
     })
+
+    // Regression: `page_sectioning` present without a `model` key used to yield
+    // `modelId: undefined`, which blew up inside resolveModel with
+    // "Cannot read properties of undefined (reading 'indexOf')".
+    it("falls back to default_model when page_sectioning has no model", async () => {
+      const pageId = `${label}_p1`
+      const storage = createBookStorage(label, tmpDir)
+      try {
+        storage.putNodeData("web-rendering", pageId, {
+          sections: [
+            {
+              sectionIndex: 0,
+              sectionType: "content",
+              reasoning: "ok",
+              html: `<section data-id="first">hello</section>`,
+            },
+          ],
+        })
+      } finally {
+        storage.close()
+      }
+
+      const configPath = path.join(tmpDir, "config.yaml")
+      fs.writeFileSync(
+        configPath,
+        [
+          'default_model: "openai:gpt-5.4"',
+          "structure_types: {}",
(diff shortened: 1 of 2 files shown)


ksokolovic:
Added test-only coverage for all three model-resolution paths. No production implementation changes.

PR #665, opened 2026-07-31
fix(styleguide): apply text colors, add missing templates, fix prompt precedence, invalidate stale preview

### LINKED ISSUE #664: StyleGuide text colors are not reliably applied to generated content, and the StyleGuide preview thumbnail goes stale
## What happened?

When configuring distinct heading and body colors in a StyleGuide and generating content with it, the selected colors are not always reflected in the final HTML. Several independent root causes contribute to this, plus a stale-preview issue when replacing an existing StyleGuide. They are grouped here because they affect the same functional area: applying and presenting the selected StyleGuide.

### 1. `data-text-color` is never applied as real CSS

The HTML generated by the LLM can correctly include `data-text-color="#RRGGBB"` on headings and paragraphs. However, neither the Studio Storyboard preview nor the packaged web reader applies that semantic attribute as a CSS color.

Result: a configured text color may have no visual effect, especially when an incidental Tailwind color class is also present.

This attribute pattern (`data-text-color` / `data-background-color`) already exists in `prompts/styleguide_generation.liquid`; this is a pre-existing application gap, not a recent regression.

Links nested inside colored text should retain their own intentional link color rather than being flattened into the surrounding heading or body color.

Likewise, when a user explicitly changes an element's text color in the Storyboard style editor, that manual choice must take precedence over the original semantic marker in both preview and packaged output.

### 2. Body text does not adopt its configured StyleGuide color

On LLM-generated pages, generated HTML may place one `data-text-color` at the section level or apply it only to headings, while body paragraphs retain arbitrary Tailwind color classes.

Root cause: `assets/styleguides/default.md` does not demonstrate separate heading and body colors at the individual element level. Without a concrete per-role example, the LLM does not consistently reproduce that pattern.

### 3. Missing style template for text-only pages (`text_only`)

There is no Page Template in `assets/styleguides/default.md` for pages with `section_type: text_only`. Existing templates cover chapter starts, regular content with images, side-by-side layouts, and the Table of Contents.

Without a concrete text-only example, the LLM can fall back to a generic layout that does not consistently follow the selected StyleGuide.

### 4. Base generation prompts can override the selected StyleGuide

The generation prompts include HTML examples with fixed colors such as `text-gray-900` and `text-gray-800`. The LLM may imitate those concrete examples instead of the colors, alignment, and typography configured in the selected StyleGuide.

This affects both standard LLM generation and `llm-overlay`. In addition, strict Match Design visual review can favor the source scan unless the prompt explicitly gives the selected StyleGuide precedence.

### 5. Re-uploading a StyleGuide can leave a stale cached preview

Some StyleGuides have a cached `${name}-preview.html`. When a `.md` StyleGuide is uploaded again using the same name, that cached preview is not invalidated, so the selector can continue showing content derived from the previous version.

The required behavior is not automatic live thumbnail generation while a file is being edited. On re-upload, the stale static preview must be removed so the preview endpoint falls back to rendering the latest uploaded markdown.

## Expected behavior

- `data-text-color` is applied consistently in both the Storyboard preview and packaged web output.
- Heading and body colors remain independent, while nested links and explicit manual color edits preserve their own color.
- Generated `text_only` pages have a concrete StyleGuide template to follow.
- StyleGuide colors, alignment, and typography take precedence over generic prompt examples and Match Design defaults.
- Re-uploading a StyleGuide under the same name invalidates both its cached static preview file and the active client-side preview query, exposing the latest uploaded content without requiring a page reload.

## Actual behavior

- Configured text colors can be ignored or only partially applied.
- Heading colors can leak into body text, or body text can retain LLM-selected utility colors.
- Text-only pages can fall back to a generic layout.
- Hardcoded prompt examples can override the selected StyleGuide.
- Re-uploading a StyleGuide can leave its previous cached preview in place.

## Steps to reproduce

1. Upload or select a StyleGuide with clearly different heading and body colors.
2. Generate or regenerate a Storyboard page using `llm` or `llm-overlay`.
3. Observe that `data-text-color` may be present without affecting the rendered color, or that body text uses an unrelated utility color.
4. Generate a `text_only` page and observe that it lacks a dedicated template to follow.
5. Re-upload an existing StyleGuide `.md` with changed content but the same filename.
6. Open its preview and observe that the cached version can still be served.

## Acceptance criteria

- Studio preview and packaged output render the same `data-text-color` behavior.
- The declared color wins over incidental descendant utility classes, except within links and elements whose color was explicitly edited by the user.
- A manual Storyboard text-color edit is preserved in saved HTML and remains authoritative in packaged output.
- Applicable StyleGuide examples put separate colors on heading and body elements.
- A `text_only` Page Template is available.
- Generation and visual-review prompts explicitly prioritize the selected StyleGuide.
- Re-uploading a StyleGuide deletes any cached `${name}-preview.html` and invalidates `["styleguide-preview", name]`; the open preview then refetches the current markdown.

## Version

v0.7.4-beta.4

## Platform

Windows - x64

## Area

Pipeline -- Storyboard

gaguerre-iugo:
## Issue

Fixes #664

## Summary

Fixes the five StyleGuide color/template issues reported in #664:

1. `data-text-color` was emitted by the LLM but not applied as CSS.
2. The default StyleGuide did not demonstrate independent heading and body colors.
3. `text_only` pages had no template to follow.
4. Generic examples in generation and visual-review prompts could override the selected StyleGuide.
5. Re-uploading a StyleGuide under the same name could leave a stale cached preview.

## Changes

- **Apply `data-text-color` in preview and packaged output**
  - `BookPreviewFrame.tsx` applies the attribute in the Storyboard preview.
  - `packages/pipeline/src/packaging/web.ts` injects equivalent runtime behavior into exported/reader HTML.
  - The declared color is forced with `!important` on the marked element and inherited by descendants governed by that marker, overriding incidental Tailwind color utilities.
  - Anchors and their contents are excluded from the inheritance pass so links retain their intentional color.
  - A manual text-color edit in the Storyboard removes the element's semantic marker and persists a manual-color marker, so the user's Tailwind color remains authoritative in both preview and packaged output.
  - Preview-only inline colors are restored before edited HTML is serialized, preventing runtime `!important` styles from leaking into saved content.

- **Add per-role colors and the missing template to `default.md`**
  - Removes the section-wide `data-text-color`, which cannot represent distinct heading and body values.
  - Adds explicit Visual Defaults for headings and body text using neutral defaults (`#111827` and `#1F2937`).
  - Aligns the declared reading width and paragraph spacing with the concrete templates (`max-w-5xl`, `space-y-4`, and `space-y-3`).
  - Applicable heading/body examples now place `data-text-color` on individual elements.
  - Adds a Text-Only Page template for `section_type: text_only`.
  - Keeps the Table of Contents on its intentional custom color scheme instead of applying the generic heading/body defaults.

- **Make StyleGuide precedence explicit**
  - `web_generation_html.liquid`, `web_generation_html_overlay.liquid`, and `visual_review.liquid` state that the selected StyleGuide's colors, alignment, and typography take precedence over generic example classes and source-page matching defaults.
  - `render-llm.ts` passes the StyleGuide into the visual-review prompt context.

- **Invalidate stale StyleGuide previews**
  - Re-uploading a StyleGuide deletes an existing `${name}-preview.html`.
  - `GET /styleguides/:name/preview` then falls back to rendering the latest uploaded markdown.
  - The upload mutation invalidates the matching TanStack Query preview key so an already-open preview refetches immediately.
  - This intentionally invalidates stale cached content; it does not introduce automatic live thumbnail generation while editing.

- **Add regression coverage**
  - Covers stale-preview deletion and first-time upload behavior.
  - Covers client-side preview cache invalidation.
  - Covers conditional export-script injection, link-color exclusion, and manual-color precedence.
  - Covers restoration of preview-only inline styles before persistence.

## Testing

- `pnpm typecheck` — clean
- `pnpm lint` — 0 errors
- API preset route tests — 8/8 passing
- Package web rendering tests — 111/111 passing
- Targeted Studio regression tests — 4/4 passing

### DIFF
--- a/apps/api/src/routes/presets.test.ts
+++ b/apps/api/src/routes/presets.test.ts
@@ -65,6 +65,50 @@ describe("global default model", () => {
   })
 })
 
+describe("styleguide upload preview invalidation", () => {
+  it("deletes a stale cached preview when the styleguide is re-uploaded", async () => {
+    const styleguidesDir = path.join(tmpDir, "assets", "styleguides")
+    fs.mkdirSync(styleguidesDir, { recursive: true })
+    fs.writeFileSync(path.join(styleguidesDir, "custom.md"), "# Custom\n", "utf-8")
+    const stalePreviewPath = path.join(styleguidesDir, "custom-preview.html")
+    fs.writeFileSync(stalePreviewPath, "<html>stale</html>", "utf-8")
+
+    const app = createPresetRoutes(configPath)
+    const formData = new FormData()
+    formData.append(
+      "file",
+      new File(["# Custom v2\n"], "custom.md", { type: "text/markdown" }),
+    )
+    const response = await app.request("/styleguides/upload", {
+      method: "POST",
+      body: formData,
+    })
+
+    expect(response.status).toBe(200)
+    expect(await response.json()).toEqual({ name: "custom" })
+    expect(fs.existsSync(stalePreviewPath)).toBe(false)
+    expect(fs.readFileSync(path.join(styleguidesDir, "custom.md"), "utf-8")).toBe(
+      "# Custom v2\n",
+    )
+  })
+
+  it("does nothing special when there is no cached preview yet", async () => {
+    const app = createPresetRoutes(configPath)
+    const formData = new FormData()
+    formData.append(
+      "file",
+      new File(["# Brand New\n"], "brand-new.md", { type: "text/markdown" }),
+    )
+    const response = await app.request("/styleguides/upload", {
+      method: "POST",
+      body: formData,
+    })
+
+    expect(response.status).toBe(200)
+    expect(await response.json()).toEqual({ name: "brand-new" })
+  })
+})
+
 describe("global specialized model defaults", () => {
   it("returns built-in defaults when no overrides are configured", async () => {
     const response = await createPresetRoutes(configPath).request(
--- a/apps/api/src/routes/presets.ts
+++ b/apps/api/src/routes/presets.ts
@@ -107,6 +107,14 @@ table{border-collapse:collapse;width:100%;margin:0.75rem 0;}th,td{border:1px sol
     fs.mkdirSync(styleguidesDir, { recursive: true })
     const content = await file.text()
     fs.writeFileSync(path.join(styleguidesDir, `${result.data}.md`), content, "utf-8")
+    // Re-uploading an existing styleguide changes its content, so any
+    // previously captured static preview HTML is now stale. Delete it so
+    // GET /styleguides/:name/preview falls back to rendering the fresh
+    // markdown content instead of serving an outdated snapshot.
+    const stalePreviewPath = path.join(styleguidesDir, `${result.data}-preview.html`)
+    if (fs.existsSync(stalePreviewPath)) {
+      fs.rmSync(stalePreviewPath)
+    }
     return c.json({ name: result.data })
   })
 
--- a/apps/studio/src/components/pipeline/stages/storyboard/components/BookPreviewFrame.tsx
+++ b/apps/studio/src/components/pipeline/stages/storyboard/components/BookPreviewFrame.tsx
@@ -535,7 +535,15 @@ ${autoFitScript}
    * the attribute, and force any descendant that does NOT have its own
    * `data-text-color` to inherit it (!important) too, so it wins over
    * incidental Tailwind color utility classes (e.g. `text-black`) the LLM
-   * may have added on its own.
+   * may have added on its own. Anchors (`<a>`) and their contents are
+   * deliberately excluded so links keep their own (usually distinct) color
+   * instead of being flattened into the surrounding body/heading color.
+   *
+   * NOTE: this mirrors the injected `textColorScript` in
+   * packages/pipeline/src/packaging/web.ts so the Studio preview renders
+   * identically to the packaged/exported HTML. The two intentionally can't
+   * share one implementation because the frontend may not import from
+   * `packages/*` (see the layer rule in AGENTS.md); keep both in sync by hand.
    */
   function applyTextColors(doc: Document) {
     const colorEls = doc.querySelectorAll<HTMLElement>("[data-text-color]")
@@ -545,7 +553,7 @@ ${autoFitScript}
       el.style.setProperty("color", color, "important")
       const descendants = el.querySelectorAll<HTMLElement>("*")
       for (const d of descendants) {
-        if (d.closest("[data-text-color]") === el) {
+        if (d.closest("[data-text-color]") === el && !d.closest("a")) {
           d.style.setProperty("color", "inherit", "important")
         }
       }
--- a/apps/studio/src/components/pipeline/stages/storyboard/components/BookPreviewFrame.tsx
+++ b/apps/studio/src/components/pipeline/stages/storyboard/components/BookPreviewFrame.tsx
@@ -528,6 +528,30 @@ ${autoFitScript}
     if (h > 0) setContentHeight(h)
   }
 
+  /**
+   * `data-text-color` is a semantic marker the styleguide/LLM stamps onto
+   * headings/paragraphs — by itself it has no visual effect. Apply it as a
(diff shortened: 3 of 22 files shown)


ksokolovic:
I reviewed the current PR head (`6e731767`) against its exact `develop` base (`b004859e`) and traced issue #664 through Storyboard preview, persistence, prompt generation, StyleGuide generation, and packaged output. I found five gaps that I think should be addressed before merging:

1. **[P1] Existing and newly generated StyleGuides can lose intentional child colors.** The new runtime forces every descendant of a `data-text-color` element to `color: inherit !important`. However, `prompts/styleguide_generation.liquid` still tells the model to put `data-text-color` on the outer section, and bundled StyleGuides such as `sri-lanka-grade2.md` combine that section marker with deliberate child heading colors. I reproduced a maroon heading being flattened to the section's brown color. The generator also still omits the required `text_only` template.

2. **[P1] Manual Storyboard color edits do not override existing inline colors.** A manual edit adds a Tailwind class and marker, but the serialization path restores the source inline `style="color: ..."`. That inline declaration continues to beat the new class. Existing generated StyleGuides use inline colors extensively, so the manual choice is not actually authoritative in this common case.

3. **[P2] Responsive color edits remove the semantic desktop color.** Tablet/mobile edits add a responsive Tailwind class but also remove `data-text-color` globally. Desktop then has neither the semantic color nor an unprefixed replacement. Resetting a responsive override has the same issue.

4. **[P2] The new authoritative defaults contradict their own templates.** `default.md` declares centered, `leading-loose` headings and italic body text, while its templates use left-aligned `leading-tight` headings and non-italic paragraphs. The updated prompts explicitly make these defaults authoritative, so this contradiction can produce inconsistent output.

5. **[P2] The packaging format version was not bumped.** `renderPageHtml` now emits a new runtime script, but `PACKAGING_FORMAT_VERSION` remains `4`. Existing books with a matching `.build-hash` can therefore keep cached pages that lack the new behavior.

The stale-preview deletion/query invalidation, visual-review context wiring, default `text_only` template, link exclusion, and preview/export mirroring are otherwise correctly connected. The focused tests, build/typecheck, lint, merge simulation, and `git diff --check` passed; the only full-suite failures were the known parallel-sensitive `part-service.test.ts` timeouts, and that file passed 13/13 in isolation.

I’m going to follow up on this branch with fixes and regression coverage for these paths.

ksokolovic:
I completed the fix pass in `e9e33483` and then re-reviewed that exact published head against the exact `develop` base (`b004859e`).

## Follow-up review result

I found **no remaining actionable code findings in issue #664's scope** after the changes:

- Legacy section-level `data-text-color` now behaves as a fallback instead of flattening intentional child colors, while new StyleGuides put the semantic marker directly on each text role.
- Manual Storyboard color edits remove the original inline `color`, so the persisted manual class is authoritative.
- Tablet/mobile edits materialize the semantic/inline desktop color before removing the global marker, preserving wider viewports and reset behavior.
- The generated StyleGuide contract now requires per-role colors and a `text_only` template; the bundled defaults now agree with their templates.
- `PACKAGING_FORMAT_VERSION` is now `5`, invalidating already-packaged pages that lack the runtime behavior.

## Verification

- Node 24 build/typecheck: passed.
- Focused regressions: **132/132 passed**.
- Full suite: **2,312/2,316 passed**; four unrelated `part-service.test.ts` cases exceeded their existing 5-second timeout under parallel load.
- Isolated `part-service.test.ts` rerun: **13/13 passed**.
- Studio lint: **0 errors** (9 existing unused-disable warnings).
- Lingui extraction: all five catalogs at 3,123 messages, 0 missing translations, no catalog diff.
- `git diff --check`: passed; isolated checkout is clean.

## Current GitHub status

GitHub has not reported a new check run for `e9e33483` yet, so hosted CI remains pending/unverified. The PR API currently says `mergeable_state: dirty`, but GitHub's compare API reports the head **5 commits ahead / 0 behind** with `b004859e` as the merge base, and the same commits pass a conflict-free local `merge-tree` check. I would treat the mergeability label as inconsistent/stale until GitHub recomputes it, rather than as a reproduced source conflict.

PR #667, opened 2026-08-01
Enforce one section in By Page mode

### LINKED ISSUE #666: Enforce one section per source page in By Page sectioning mode
## Problem

The Sectioning UI offers a **By Page** mode and the prompt asks the model to emit exactly one section, but the runtime validator does not enforce that invariant. If the LLM returns multiple sections anyway, they are accepted and Storyboard renders/packages them as separate pages.

This breaks the expected one-source-page-to-one-storyboard-page behavior and can increase a book page count substantially.

## Reproduction

1. Select **By Page** in Sectioning.
2. Run Sectioning on a page containing multiple activities or visual regions.
3. If the LLM emits more than one section, observe that validation accepts it.
4. Run Storyboard; each accepted section is rendered as a separate page.

Observed in a Science Standard Five book: 88 PDF pages produced 153 section renderings.

## Expected behavior

When `page_sectioning.mode` is `page`, validation must require exactly one section. A multi-section result should enter the existing validation-feedback/retry flow instead of reaching Storyboard.

## Proposed fix

Propagate the configured sectioning mode into the page-sectioning validator and add a validation error whenever page mode returns anything other than one section. Add regression coverage for the invariant.

KaReeeeeeeeEM:
## Summary

- pass `page_sectioning.mode` into the runtime validator
- reject multi-section LLM output when **By Page** mode is selected
- route invalid output through the existing validation-feedback/retry loop
- add regression coverage for the one-section invariant

Fixes #666.

## Why

The prompt already instructs the model to emit one section in page mode, but this was not programmatically enforced. A model response containing multiple sections could therefore reach Storyboard, where each section becomes a separate rendered page.

## Verification

- `pnpm build`
- `pnpm exec vitest run packages/pipeline/src/__tests__/page-sectioning.test.ts` (48 tests passed)
- `pnpm typecheck`
- `git diff --check`

### DIFF
--- a/packages/pipeline/src/__tests__/page-sectioning.test.ts
+++ b/packages/pipeline/src/__tests__/page-sectioning.test.ts
@@ -24,6 +24,7 @@ function makeCtx(overrides?: {
   roleKeys?: string[]
   sectionTypeKeys?: string[]
   availableImageIds?: string[]
+  mode?: "page" | "dynamic"
 }) {
   return {
     structureKeys: new Set(
@@ -40,6 +41,7 @@ function makeCtx(overrides?: {
       overrides?.sectionTypeKeys ?? ["text_only", "images_only"]
     ),
     availableImageIds: new Set(overrides?.availableImageIds ?? []),
+    mode: overrides?.mode,
   }
 }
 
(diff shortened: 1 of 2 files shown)


KaReeeeeeeeEM:
Replaced by #692 after renaming the source branch to the required `feat/` prefix.

PR #700, opened 2026-08-03
feat(export): offline TTS regeneration for exported ADT bundles

elasticsounds:
**Requirement**
In the Tanzania ADT Project, the ADTs are edited in codex. The team requested a way to be able to regenerate TTS outside of ADT Studio

Exported ADT bundles now ship a dependency-free `tools/regenerate-tts.mjs` script, an editable `tools/tts.config.json`, a README, and a `regen/manifest.json` of per-unit baseline cache keys, so a user can edit a book's `content/i18n/<lang>/texts.json` outside ADT Studio and re-record only the lines whose text (or voice/model) changed via OpenAI TTS — re-aligning word highlighting for just those lines. Change detection is content-addressed against the manifest, so no audio is duplicated and manually uploaded clips are never regenerated. `packageAdtWeb` emits these assets and now threads a `configDir` (through the export service, package route, and pipeline DAG) so the shipped per-language speech settings match what produced the audio. Gemini/Azure providers and output compression are recognized but deferred to a follow-up.


elasticsounds:
Suggested update:v Refactor standalone TTS regenerator to share core speech implementation

The standalone TTS regeneration script currently duplicates parts of the main OpenAI, Gemini, and Whisper implementations. Although some helpers have parity tests, changes to the primary TTS clients can still drift from the exported script.

Goal

Keep the exported CLI dependency-free and compatible with Node 20 while sharing provider request/response logic with the main application.

Proposed work

Add characterization tests for current OpenAI, Gemini, Whisper, cache-key, and audio-decoding behavior.
Extract zero-dependency speech primitives into @adt/llm.
Make both the normal TTS clients and regeneration CLI consume those primitives.
Extract shared cache-key and text-normalization helpers.
Build the CLI with esbuild into a single dependency-free .mjs.
Add regen:build and regen:check commands for deterministic generation and CI drift detection.
Include the generated script hash in packaging cache invalidation.
Remove duplicated provider code and document intentional differences.
Intentional standalone differences

The regenerator should retain manifest-based caching, serial execution, bundle-specific filesystem handling, and no ADT database dependency. Page batching and Azure support can remain unsupported until explicitly implemented.

Acceptance criteria

One implementation each for OpenAI, Gemini, Whisper, and speech cache keys.
Generated CLI runs using only Node 20.
Shared speech changes automatically trigger generation checks and package invalidation.
Existing manifest v2 bundles remain compatible.

kinggamma:
i tested it and it works as expected for programmers with the cli

elasticsounds:
Since this has been tested and accepted - it is being merged for use in web export only. Tools will be disabled by default.

PR #707, opened 2026-08-03
feat: add publishing, selective speech, and storyboard editing

### LINKED ISSUE #696: Enhancement: prompt and regenerate a single voice entity
## Gap
The Speech stage currently forces experts to rerun broad speech generation or regenerate only a missing Gemini clip with the global speech prompt. Existing phrase audio cannot be adjusted using a phrase-specific instruction.

## Proposed behavior
- Expose an **Adjust voice** action for each generated speech entity.
- Let the expert enter a natural-language instruction such as “speak more slowly and emphasize the final word.”
- Regenerate only that entity while preserving all unaffected clips.
- Store the result as a new TTS node-data version for rollback.
- Include the expert instruction in the inspectable LLM log.
- Invalidate only that entity’s stale timestamps.

## Accessibility and validation value
Phrase-level regeneration makes validator corrections affordable and prevents a full-book TTS rerun after small content or pronunciation changes. It also keeps downstream inclusive outputs aligned with validated text.

### LINKED ISSUE #703: Enhancement: add an in-app GitHub publishing stage for accessible books
## Problem
ADT experts currently leave ADT Studio to initialize a repository, inspect changes, commit, push, configure GitHub Pages, and share a validation link. This breaks the pipeline and makes regenerated media/sign-language updates harder to keep synchronized.

## Proposed enhancement
Add a dedicated **Publish** stage after Export that provides:
- persistent GitHub account configuration with clear PAT permission guidance
- a local Git history/diff view driven by the actual book worktree
- generated commit messages, commit/push controls, and optional automatic synchronization
- a sequential, inspectable deployment workflow with retry and failure diagnostics
- GitHub Pages configuration, deployment history, public preview link, and deployment inspection
- visible task progress in the pipeline sidebar

## Acceptance criteria
- Publish is defined through the pipeline single source of truth.
- All book artifacts remain inside the book directory.
- Tokens are never logged or placed in URLs.
- Git operations expose progress and actionable failures.
- Published deployments and their changed-file diffs are inspectable in ADT Studio.

### LINKED ISSUE #704: Prevent Publish sync from overwriting local book edits and expose Pages build failures
## Problem
Publish synchronization can report remote activity indefinitely, obscure local changes, or attempt synchronization while the book worktree is dirty. GitHub Pages build failures are not always surfaced with enough detail to act on.

## Expected behavior
- On every Publish-page entry, detect local changes first and stop the loading state when results arrive.
- Use real Git status/diffs for all pipeline-created or externally edited book artifacts.
- Never pull over a dirty worktree; detect divergence/conflicts and offer opening the repository in the user's editor.
- Show synchronization as a persistent bottom status item rather than repeated toasts.
- Run synchronization/deployment stages strictly in sequence with bounded retries.
- Surface the failing GitHub Pages workflow step, message, and logs URL.
- Allow reconnecting/replacing expired credentials and rerunning a failed deployment.

### LINKED ISSUE #706: Enhancement: canvas-style component editing in Storyboard
## Problem
Fine-tuning storyboard fidelity currently requires editing generated HTML/CSS manually. ADT experts need direct manipulation controls without losing accessibility metadata or responsive behavior.

## Proposed enhancement
Add an editing mode that supports:
- dragging components into position
- Command/Ctrl multi-selection and marquee selection
- grouped movement and deletion
- component rotation
- image resize handles
- adding new components
- keyboard-accessible alternatives for every pointer interaction
- persistence through entity-level versions so edits can be rolled back

## Acceptance criteria
- Changes persist to the book directory and appear in page history.
- Selection and transforms work across iframe boundaries.
- Editing controls preserve semantic HTML, focusability, captions, and accessible names.
- Mobile/responsive layouts remain usable after transformations.
- The interaction layer has automated tests for drag, resize, rotate, multi-select, and delete.

KaReeeeeeeeEM:
### Publish stage
- adds a dedicated Publish stage with GitHub configuration, local-first Git status and file diffs
- supports commit/push, GitHub Pages deployment, deployment inspection, public preview links, sequential workflow progress, retry, and actionable failure diagnostics
- prevents remote synchronization from overwriting dirty local book edits and exposes editor conflict recovery

### Speech enhancements
- restores individual and multi-select speech regeneration
- forces cache bypass for explicitly regenerated clips
- preserves the selected entity text and adds speech preparation/instruction controls
- improves retry and progress behavior for selective generation

### Storyboard editing
- adds canvas-style drag, multi-select, resize, rotation, deletion, and keyboard movement across the preview iframe
- adds a global book font selector and image resizing controls
- documents reusable storyboard remediation practices

Closes #703
Closes #704
Closes #696
Closes #706

## Verification
- pnpm typecheck
- pnpm lint (0 errors; 9 pre-existing warnings)
- 7 focused files / 122 tests passed
- full suite: 2,331 tests passed; 3 parallel-load timeouts subsequently rerun independently
- rerun of the 3 affected files: 74/74 tests passed
- pnpm --filter @adt/studio extract
- git diff --check origin/develop...HEAD


elasticsounds:
Johnson this is good work moving in the right direction - typically we require though that different features are broken into different pull requests so we can better manage them before they are merged in - best for testing etc. I'm going to test and review and give you feedback. If it is possible to break into the publish stage, speech enhancements, storyboard editing so we are able to better review test and manage with team.

Let me know if you need assistance in breaking these into separate PRs.

elasticsounds:
The publish stage is very cool - the deployment was easy and worked. Only at the end it didn't take me to the link. The link should also be on the home page in the project as a way to open the preview of the book. The PAT was easy to create, just may not be possible if someone has a corporate account.

The storyboard editor seems easy as in I can just drag and it uses CSS transform and rotation. However it is not compatible with the tailwind and responsive model. We have to reconcile how we want that to work. I'm also not clear on what the editors want to do dragging things around - almost every time it makes the layout worse? 

The TTS regenerate button is great - the check marks to select more than one file seems a bit overkill and not as necessary. Needs a UX redesign for that, but regenerate a single file makes sense.

elasticsounds (commented):
Thanks — there is a lot of genuinely good work here. The publish flow is really cool and works end-to-end, and selective speech regeneration is very welcome. Before line-level fixes, though, one structural request, then feedback per area.

## 1. Please split this into three PRs

This bundles three independent features — GitHub publishing, speech regeneration, and storyboard canvas editing — at 8.6k added lines across 48 files, closing four unrelated issues. They share no code beyond incidental wiring, and as the feedback below shows they are in very different states (one is close, one needs cache-correctness fixes, one needs a design rework). Splitting lets us review, merge, revert, and release each independently:

1. `feat: GitHub publishing stage`
2. `feat: selective speech regeneration`
3. `feat: storyboard canvas editing + book font selector`

---

## 2. Storyboard canvas editor

### Responsiveness must keep working (blocking)

Drag/rotate/resize persist as viewport-agnostic inline styles: `el.style.translate = '<N>px <N>px'` (`iframe-interactive.ts:302`), `el.style.rotate = '<N>deg'` (`:345`), and image resize writes fixed px `width`/`height` plus `maxWidth:'none'` (`:381-388`). One transform applies at every screen size, and the offsets are measured at the current preview width (`renderWidth={DEVICE_WIDTHS[deviceView]}`, `StoryboardSectionDetail.tsx:2811`) — so a layout arranged on the desktop preview misplaces or overflows content on tablet/mobile, and `maxWidth:'none'` defeats the responsive image constraints. This also contradicts the PR's own remediation guide ("Avoid fixed pixel dimensions", `docs/STORYBOARD-BOOK-REMEDIATION-GUIDE.md:76-79`).

Breaking out of flow can be fine — but only if it is explicit and per-device. What I'd like to see:

- Treat dragging as an explicit **canvas mode**, and on drop **convert the position into responsive values** instead of raw inline px. We already have the machinery: the style inspector writes breakpoint-scoped Tailwind variants (`max-lg:` / `max-sm:`, `style-editor/StyleEditorPanel.tsx:131-137` + `device-breakpoint.ts`). The Tailwind equivalents of these transforms are `translate-x-[24px] translate-y-[13px] rotate-[7deg]` — scoped per device as e.g. `max-sm:translate-x-[5%]`. Percent/rem-based offsets will survive reflow far better than px.
- The user should be able to position elements separately for desktop, tablet, and mobile (the device toggle already exists), with the un-overridden breakpoints keeping the flow layout.
- **The translate/rotation values must be visible and editable.** Right now there is no numeric readout anywhere — after a drag the values exist only in the `style` attribute (the style-editor panel has no transform/position section). Add X/Y/angle fields to the inspector, per breakpoint.

### Editor UX

- The selection handles are `<button>`s with text glyphs ✥ / ↻ appended to the iframe body (`iframe-interactive.ts:165-216`). They don't match the app's visual language — proper icons, consistent hit areas, and resize handles that work for more than single-selected images. Also inconsistent: containers selected via `select-container` can be Delete-key removed but never dragged or rotated (`:671-681`).
- Clicking any element auto-expands the style sidebar via a `w-0 → w-[300px]` width transition (`StyleEditorPanel.tsx:112-115`), which reflows and rescales the preview on every selection — frustrating. Keep the layout stable: reserve the panel space, overlay it, or make expansion manual.

### Bugs

- **Delete/Backspace while typing deletes the page element.** The delete branch (`iframe-interactive.ts:773-781`) never checks whether focus is in a form control, and clicking into an activity input doesn't clear the selection (`:648-656`) — pressing Backspace in a fill-in-the-blank removes the selected element and blocks the text edit.
- **Keyboard movement self-destructs.** Every `element-moved` (including each arrow-key nudge, `:721-724`) reserializes and re-injects the whole body (`StoryboardSectionDetail.tsx:1538-1552` → `BookPreviewFrame.tsx:635`), destroying the focused handle; the next arrow key falls through to the page-navigation branch (`iframe-interactive.ts:782-784`) and can navigate the user away. Each nudge also fires a `/convert-math` POST and an auto-fit pass.
- **Multi-select rotation** rotates each element around its own center while measuring the angle against the group bounds (`:330-346`) — the group's arrangement isn't preserved. Resizing an already-rotated image computes from the rotated AABB (`:364,746`), inflating dimensions.
- **No undo.** Moves/rotates/deletes have no Ctrl+Z; an accidental delete is only recoverable via Discard, which throws away *all* pending edits. (Versioning on save is fine — `PUT .../rendering` appends new versions.)
- Duplicate pointer + mouse listener registrations and no `setPointerCapture` (`:416-433`): touch devices can double-fire `element-moved`, and releasing the button outside the iframe strands the drag.
- Serialized HTML is written back with no re-sanitization (DOMPurify runs only on load — `BookPreviewFrame.tsx:262-275` vs `:419-421`), and "Add text" persists editor scaffolding classes (`border-dashed border-violet-300 …`) into the published book (`:283-289`).
- Coverage: one 55-line smoke test for ~500 new lines of untyped ES5-in-a-template-string; the drag/rotate/resize math, serialization round-trip, and the delete-while-typing case are untested. `iframe-interactive.ts` (876 lines), `StoryboardSectionDetail.tsx` (3,356), and `BookPreviewFrame.tsx` (1,114) all keep growing against the split-components convention.

## 3. Book font selector

`BookFontSelector` is mounted in the shared step header (`StepViewRouter.tsx:152`), so a font dropdown appears on **every** pipeline page — extract, quizzes, publish — firing a `book-fonts` query everywhere, and fixed-layout books show a permanently disabled control. Having the font on screen at all times is unnecessary.

Suggestion: fold it into a **"Book design" options panel on the storyboard stage** (font plus related design settings). The existing settings under the storyboard stage are hard to find and navigate — this could be where they get consolidated and made discoverable.

Minor: the three new fonts (`packages/types/src/reflowable-fonts.ts:45-55`) are marked `google: false`, whose documented meaning is "bundled locally" — but Helvetica/Times/Sassoon Primary are neither loaded nor bundled, so selecting them silently falls back via `reflowableFontFamilyChain`, and the field's comment is now wrong.

## 4. Speech regeneration

The regenerate button is great. One UX note: the always-visible per-row checkboxes (`h-4 w-4 accent-rose-600`, `LanguageView.tsx:2386-2394`, master at `:1809`) feel heavy — consider hover-reveal or an explicit select mode.

Correctness issues that need fixing before merge:

- **Page-batched Gemini languages silently revert regenerations.** Batchable entries skip the per-entry reuse check and are re-sliced from the page-level transcript cache (`stage-runner.ts:2744-2758`) — the next full Speech run copies the old sliced audio over the regenerated clip.
- **Regenerating with edited text is also reverted later**: the new audio is cached under the edited-text key, but a full run recomputes the key from catalog text and restores the stale entry (`tts.ts:756-760` + `stage-runner.ts:668-685`).
- **Stale timestamps**: neither regenerate mutation re-transcribes or invalidates `tts-timestamps` (`LanguageView.tsx:1188-1240`), so read-along highlighting and EPUB SMIL desync from the new audio.
- `forceRegenerate` overwrites the audio file *and* the cache entry in place (`packages/pipeline/src/speech.ts:583,615-619`) — the old clip is unrecoverable, which conflicts with the never-overwrite/rollback principle.
- `regenerateSelectedMutation` has no `onError` (`LanguageView.tsx:1218-1241`): the first failure silently aborts the loop with no UI error and no query invalidation for the clips that did succeed.
- Untranslated entries fall back to source-language text with the target-language voice (`LanguageView.tsx:1849,2698`).
- Heads-up: routing all `sw*` text through `prepareTextForSpeech` (`speech.ts:542`) rotates every existing Swahili cache key — the next full run regenerates *all* Swahili audio at full API cost, with no opt-out (contrast the Azure change, which got a `synthesisRevision` gate). Also `swahiliNumberWords` reads "1,500" as a decimal ("moja koma tano sifuri sifuri", `speech.ts:73-88`).

## 5. GitHub publishing

Really cool feature, and it works. The PAT requirement is a real barrier for individuals — near-term, the in-app guide (`GitHubPublisher.tsx:569-596`) should steer to a **fine-grained token** (single repo, Contents + Pages) instead of a classic full-`repo`-scope token; longer-term a GitHub App / device-flow login would remove the friction entirely. Token handling is otherwise clean (header-only transport, tokenless remote URL, memory-only server-side) — though it's stored in plaintext `localStorage` with no expiry (`GitHubPublisher.tsx:45,691`); on desktop, Electron `safeStorage` would be better.

Issues to fix:

- **Path traversal on pull**: remote tree paths are joined onto `pullRoot` without validating the result stays inside it (`github-publishing-service.ts:169-171`) — a crafted tree containing `..` (pushable by any repo collaborator) writes outside the book directory. The diff endpoint does this check correctly (`:347-348`); the pull path must too.
- **Publish race**: `/publish` reads the `active` guard before `await c.req.json()` and sets it after (`routes/github-publishing.ts:275-293`) — two concurrent POSTs both pass the 409 check and interleave git operations on the same repo.
- **Pull leaves a permanent phantom conflict**: a successful pull replaces `adt/` but never updates the local repo refs or the recorded commit (`service:143-184` + `routes:231-246`), so remote-status reports `conflict: true` and auto-sync stays paused until the next full publish.
- **Publish can silently revert collaborator commits**: `synchronizeLocalGitRepository` force-resets the branch to the remote head without touching the worktree (`service:376-410`), so commits made directly on GitHub read as local edits and the next publish reverts them; `/publish` never consults `behind`/`conflict` first.
- `GET /changes` has side effects (`prepareExport`) and is polled every 3s (`routes:143-156`, `GitHubPublisher.tsx:677`) — concurrent rebuilds can mutate the worktree mid-commit.
- Version record filenames derive from the current file count (`service:555-561`) — a concurrent write or deleted record collides and overwrites a prior version.
- Tests cover 3 of 9 route endpoints; the dirty-edit pull guard, the publish-concurrency 409, and the `behind`/`conflict` computation have none. (The service tests using real temp git repos are good.)
- `isomorphic-git` as the one new prod dependency is justified (pure JS, fits the no-native rule), but the system-git push fallback will throw `spawn git ENOENT` on machines without git installed — worth a graceful message.

---

CI is green (docker / i18n / test), typecheck and lint pass. Happy to go through any of these in more detail on the split-out PRs.

KaReeeeeeeeEM:
Great insights and improvements @elasticsounds, i am working on splitting the PR to make it feature-specific and then i'll work on the improvements separately in their respective branches. I'll then notify you as soon as i am done with the process.

KaReeeeeeeeEM:
Implemented the review in three independently reviewable PRs, each rebased onto the latest develop:

- #709 — GitHub publishing: secure pull paths, serialized publishing, explicit materialization, ref synchronization, remote-advance protection, session credentials, PAT guidance, project links, missing-Git handling, collision-safe versions, and added traversal/race/divergence tests.
- #710 — Speech: durable versioned regeneration, edited-text persistence, page-batch preservation, timestamp invalidation, partial-failure handling, explicit selection mode, target-language safeguards, and Swahili number parsing.
- #711 — Storyboard: explicit responsive canvas mode with per-device transforms and numeric controls, typing-safe deletion, keyboard stability, multi-select rotation, undo/pointer handling, sanitized persistence, Storyboard-only Book design settings, correct font availability, and interaction tests.

All three pass build, typecheck, lint, strict i18n compilation, and their focused test suites. Full-suite parallel timeout observations and successful isolated reruns are documented in each PR.

KaReeeeeeeeEM:
Closing this combined PR in favor of the three focused replacements above, as requested in review.

PR #712, opened 2026-08-04
fix: recover Image Meaningfulness pages after LLM failures

### LINKED ISSUE #685: Retry transient LLM transport failures without pruning pages
## Problem

During Science Standard 5 extraction, Image Meaningfulness failed on individual pages with retryable transport errors such as `Connect Timeout Error` and `other side closed`. A transport failure must not be interpreted as evidence that a page or image is meaningless, and it must never cause the page to disappear from downstream Storyboard output.

## Proposed general fix

- Classify connect timeouts, connection resets/closures, 408, 429, and retryable 5xx responses as transient.
- Retry with bounded exponential backoff and jitter.
- Preserve successful per-page LLM results through the existing LLM cache.
- Resume only failed page IDs rather than restarting the whole step.
- After retries are exhausted, retain the page and surface a visible per-page error/retry action instead of pruning it.
- Record attempts, final error class, page ID, prompt/model, and cache status for transparency.

## Acceptance criteria

- A simulated timeout on one page is retried and does not remove that page from Sectioning or Storyboard.
- Successful pages are not called again when the failed page is resumed.
- Exhausted retries leave an actionable error state and preserve page cardinality.
- Tests cover connect timeout, closed connection, 429, retryable 5xx, and a non-retryable error.

## Evidence

Observed errors included:

- `Image Meaningfulness: pg088 failed: Cannot connect to API: Connect Timeout Error`
- `Image Meaningfulness: pg024 failed: Cannot connect to API: other side closed`

The reusable workflow and quality gates are documented in the Storyboard Book Remediation Guide PR.

ksokolovic:
## Summary

- preserve the existing bounded LLM retry loop for malformed model output and unknown failures while fast-failing known permanent authentication, quota, and model errors
- classify transient transport failures, including temporary DNS failures, and retain exponential backoff with jitter
- inspect structured provider error payloads so permanent quota errors are distinguished from temporary rate limits
- let Studio users retry only the failed Image Meaningfulness page, skip it, or stop after automatic retries are exhausted; successful pages are not rerun
- preserve page IDs and existing image-filtering data when a failed page is skipped so Sectioning and Storyboard retain page cardinality
- expose translated error classes, attempt counts, retryability, cache status, prompt, model, and outcome in Studio diagnostics, including final semantic-validation exhaustion
- return a specific 400 for an ineligible retry while keeping the pending decision available for Skip or Stop

## Behavior clarification

The failure in #685 was a halted pipeline, not a page-deletion path: Image Meaningfulness already kept the existing page result when an LLM call failed. This PR adds the interactive recovery path that allows Studio to continue after the automatic retry budget is exhausted.

`max_retries`, exponential backoff, and jitter already existed. The classifier now fast-fails only known permanent failures and preserves bounded retries for malformed model output and unknown failures. The CLI/headless default remains `pageErrorPolicy: "stop"`, so without an interactive Studio decision the run still stops after retries are exhausted.

## Acceptance criteria

- [x] malformed model output and transient transport failures use the configured bounded retry budget
- [x] known permanent authentication, quota, and model failures fail immediately
- [x] temporary 429 rate limits remain retryable while structured `insufficient_quota` responses fail immediately
- [x] `EAI_AGAIN`, connect/request timeouts, closed connections, 429s, and retryable 5xx failures are classified as transient
- [x] successful pages are not called again when a failed page is resumed
- [x] a skipped exhausted page preserves both page IDs through Sectioning and Storyboard
- [x] dialog copy only advertises Retry when Retry is available
- [x] error-class labels are translated in all supported locales
- [x] retry remains page-scoped and cannot be combined with apply-to-all
- [x] an ineligible retry returns 400 without consuming the pending decision
- [x] exhausted semantic validation is recorded as a final retryable `model-output` failure

## Verification

- Node.js 24.13.0
- `pnpm exec vitest run --no-file-parallelism` — 204 files, 2,545 tests passed
- real AI SDK HTTP integration with `maxRetries: 2`: malformed output makes 3 requests; HTTP 401 and `insufficient_quota` 429 each make 1 request
- `pnpm typecheck`
- `pnpm lint` — 0 errors; 9 pre-existing unused-suppression warnings
- `pnpm --filter @adt/studio extract` — 3,305 messages per locale, 0 missing in `en`, `es`, `fr`, `pt-BR`, and `sq`
- `pnpm build`
- `git diff --check origin/develop...HEAD`

Closes #685


elasticsounds (commented):
Nice work — the diagnostics and the decision-broker plumbing are a real improvement, and the tests around them are thorough. I verified the PR's claims locally and they hold up:

| Check | Result |
|---|---|
| `pnpm exec vitest run --no-file-parallelism` | 195 files / 2390 tests pass |
| `pnpm typecheck` | exit 0 |
| `pnpm lint` | 0 errors; the 9 warnings are all in files this PR doesn't touch |
| i18n catalogs | all 5 locales fully translated, 0 new empty `msgstr` |

One blocking finding, a few smaller ones, and a scoping question below.

---

## 1. Malformed model output is no longer retried — and can't be retried by hand

`packages/llm/src/client.ts:279` changes `if (attempt < maxRetries)` to `if (classification.retryable && attempt < maxRetries)`. Because `classifyLLMError` returns `non-retryable` for anything it doesn't explicitly recognise, that also catches `AI_NoObjectGeneratedError` / `AI_TypeValidationError` — the model returning unparseable or schema-invalid JSON.

I measured this by driving the real `ai` SDK over real HTTP against a local server that injects each failure mode, counting the requests the retry loop actually issues, then repeating with `origin/develop`'s `client.ts`. With `maxRetries: 2`, **3 requests = retried, 1 = failed immediately**:

| Failure mode | Error thrown | `develop` | this PR |
|---|---|---|---|
| Malformed model output | `AI_NoObjectGeneratedError` | **3** | **1** :warning: |
| Socket closed mid-response | `AI_APICallError` (`other side closed`) | 3 | 3 :white_check_mark: |
| HTTP 503 | `AI_APICallError` | 3 | 3 :white_check_mark: |
| HTTP 401 | `AI_APICallError` | 3 | **1** :white_check_mark: |

The 401 row is the win this PR is after, and it's a good one — that used to burn 6 attempts with backoff up to ~60s in production.

But the first row is a regression, and in production it's worse than the table suggests: `DEFAULT_LLM_MAX_RETRIES` is 5 (`packages/types/src/config.ts:8`), so malformed output goes from **6 attempts to 1**. Re-sampling is usually what fixes that class of failure. And because the classification is `retryable: false`, `runMeaningfulnessPass` sets `canRetry: false`, so the new dialog won't offer **Retry this page** either — a page that fails on unparseable JSON becomes a dead end in exactly the workflow this PR built to rescue dead ends.

The root cause is that the classifier's fallback is inverted: unknown → non-retryable, so the blast radius is every class nobody enumerated. `EAI_AGAIN` (transient DNS) lands there too.

Two ways out:

1. Add a retryable `model-output` class for `NoObjectGeneratedError` / `TypeValidationError`. Narrow, fits the existing structure.
2. Invert the default — unknown → retry (today's behaviour), fast-fail only on a known-permanent list (401/403/400/quota). Keeps the 401 win with no regression surface. I'd lean this way.

Either way, please pin `NoObjectGeneratedError`'s classification in `errors.test.ts` so the decision is deliberate and stays put.

<details>
<summary>Reproduction harness</summary>

A local `http.Server` that counts requests and injects one fault per mode; `createLLMModel({ modelId: "custom:gpt-4.1", credentials: { customBaseUrl: "http://127.0.0.1:<port>/v1" } })` drives it through the real SDK. `badjson` mode returns HTTP 200 whose message content is `{ this is definitely not json`; `close` mode calls `res.socket.destroy()` (this reproduces the exact `UND_ERR_SOCKET` / `other side closed` from the issue). Run once on the branch, once with `git show origin/develop:packages/llm/src/client.ts` swapped in. Happy to push it as a test if useful.

</details>

---

## 2. `EAI_AGAIN` isn't classified as transient

A temporary DNS resolution failure is textbook transient but currently classifies as `non-retryable`. One line in `CONNECT_TIMEOUT_CODES` (`packages/llm/src/errors.ts:20`). `ECONNREFUSED` / `ENOTFOUND` are more debatable, but worth a deliberate call.

## 3. The dialog copy promises Retry everywhere, but only one step provides it

`PageErrorDecisionDialog.tsx:63` now always reads *"Retry this page, skip it and continue, or stop the step entirely."* Only `runMeaningfulnessPass` passes `canRetry`; the other `reportPageFailure` call sites (sectioning, captions, rendering, easy-read, TTS) leave it undefined, so no Retry button renders and the description describes an option that isn't there. Scoping the behaviour to Image Meaningfulness matches the issue — the copy just needs gating on `current.canRetry`.

## 4. Test gap on the headline acceptance criterion

"Exhausted retries preserve page cardinality" is only exercised through the happy path — `stage-runner.test.ts` returns `"retry"` and the second call succeeds. There's no case where the page fails permanently (or is skipped) and still lands in `page-sectioning` / `web-rendering`. That's the criterion most worth locking down, since it's the one the issue was actually filed about.

Also untested: the `TimeoutError` branch for `AbortSignal.timeout` (`errors.ts:105`). I confirmed it still classifies as `request-timeout`, so the 90s internal timeout keeps retrying as the comment at `client.ts:271` claims — but it's one regex tweak away from silently breaking, and nothing would catch it.

## 5. Retry on a non-retryable entry answers 409

`page-error-decisions.ts:140` returns `false`, which `stages.ts:169` maps to 409 "Decision already resolved". `use-book-run.ts` has already optimistically dropped the dialog, so the run blocks until the 5-minute safety timer stops the step. Only reachable via direct API use today, but 400 would be more honest than 409.

## Nits

- `errorClass.replaceAll("-", " ")` renders untranslated English in both the dialog and the LLM Logs tab — the only user-visible strings in the PR that aren't in the catalogs.
- `stage-runner.ts:3331`: `while (unprunedImages.length > 0)` never re-evaluates its condition inside the loop, so it reads as bounded but is `for (;;)` with breaks.

---

## Scoping question

The PR never touches a code path that *removes* a page — on failure `runMeaningfulnessPass` keeps `existing` in `results` and simply doesn't write new node data, so nothing gets pruned. The disappearance in #685 looks like it was the step failing and halting the run before Sectioning/Storyboard ever executed, rather than pruning. Could you confirm? If that's right, the title slightly overstates it, and it's worth being explicit in the PR body that the recovery path is the interactive dialog: with the default `pageErrorPolicy: "stop"` (CLI/headless), `handlePageFailure` still returns `"stop"` and the step fails as before.

Related framing worth capturing somewhere: `max_retries` already defaulted to 5 and `backoffDelay` already had exponential backoff with jitter, so transient errors were already getting 6 attempts before this PR. The classifier doesn't add retries — it removes them for anything it can't classify. That's the right trade for 401, which is why finding #1 matters: right now it also removes them for the most common flaky failure there is.

---

Findings 2-5 and the nits are all fine as follow-ups. Finding 1 is the one I'd want resolved — or at least an explicit "yes, we accept losing retries on malformed output" — before this lands.

*Verified locally at 9ef5bd6c: full test suite, typecheck, lint, i18n catalogs, plus the live-SDK A/B above.*

elasticsounds:
In the testing it seems that there are some instances where the model is returning malformed json and it's not allowing it to retry which seems like a regression. Can you see if you can reproduce?

ksokolovic:
Thanks for the detailed review — valid points. I addressed the blocking regression and the remaining findings in focused commits:

- Restored retries for malformed/schema-invalid model output and unknown failures.
- Added `EAI_AGAIN` transient classification.
- Added a real AI SDK HTTP harness proving malformed output gets 3 attempts with `maxRetries: 2`, while HTTP 401 and `insufficient_quota` 429 fail after one.
- Gated Retry-specific dialog copy and localized diagnostic labels.
- Added coverage proving skipped pages retain their IDs through Sectioning and Storyboard.
- Made ineligible retries return 400 without consuming the pending decision.
- Added final semantic-validation failure diagnostics.
- Clarified the PR title and description: the failure halted the pipeline rather than pruning pages, and the recovery path is scoped to Image Meaningfulness in interactive Studio runs.

Final verification:

- 204 test files / 2,545 tests passed
- Typecheck and build passed
- Lint passed with 0 errors
- i18n, test, and Docker CI checks are green

PR #721, opened 2026-08-04
Feature: AI Agnostic

zignaggo:
# Make the AI pipeline provider-agnostic

## Summary

This PR replaces the hardcoded OpenAI integration with a provider registry: every AI capability (structured text, images, speech) is now served by pluggable provider modules described by manifests, with per-request credential resolution, dynamic model discovery, and connection health checks. It ships nine built-in providers — OpenAI, Anthropic, Claude Agent (Claude Code login), Codex, Google, Gemini, Azure, Ollama, and custom OpenAI-compatible endpoints — and makes the Studio UI fully manifest-driven, so adding a provider no longer touches the frontend.

## Architecture

- **`@adt/types`** — new `ai-provider` manifest schema and qualified model-id schema (`provider:model`), the single source of truth for provider capabilities, credential fields, and modality support.
- **`@adt/llm`** — provider ports, registry, and credential resolution; generation, caching, and prompt handling routed through the registry; shared REST adapters (OpenAI-compatible and Anthropic); advisory dynamic model discovery; connection health checks. Cache keys carry per-provider adapter fingerprints so provider behavior changes invalidate stale entries.
- **`@adt/agents`** — agents depend on an `AgentBackend` port instead of importing the AI SDK directly.
- **`apps/api`** — provider catalogue routes, credential middleware (per-provider `X-ADT-Provider-*` headers), per-request credential resolution across all AI routes, and a provider health endpoint.
- **`apps/studio`** — manifest-driven credential and model settings, a model picker fed by live discovery, live provider connection status in Settings, and stage availability derived from provider manifests instead of hardcoded checks. i18n catalogs updated for all five locales.

## Local CLI providers

Two providers run through locally installed CLIs instead of HTTP APIs, reusing existing logins (no API key required):

- **Claude Agent** — drives the Claude Agent SDK with the machine's Claude Code login. Turns run isolated (no tools, no session files, no local settings, scratch cwd) so results stay reproducible, and credential env vars are scrubbed so an ambient token can never be billed by accident. Model discovery goes through the SDK's `supportedModels()` when no API key is set, falling back to the Anthropic REST catalogue when one is.
- **Codex** — same shape via `codex exec`, with secret redaction on CLI error output.

## Fixes along the way

- **Ollama thinking models exhausted the token budget** — reasoning output is now capped so structured answers complete.
- **Timeout floor for local providers** — Ollama, Claude Agent, and Codex get a 600s minimum timeout; pipeline step timeouts tuned for HTTP APIs killed CLI-backed calls mid-generation, and each retry re-spawned the CLI and re-spent the whole prompt.
- **Claude Agent turns overthinking** — the CLI defaults to adaptive extended thinking, which produced 13k–32k output tokens per structured extraction (~3–8 minutes per call, surfacing as "Operation aborted" timeouts). Thinking is now disabled for structured turns: the same call returns in seconds with output proportional to the answer. The adapter version was bumped so cached thinking-era responses are invalidated.

## Config

- `config.yaml` now defaults to a local Ollama model, so a fresh install works without any API key.

## Testing

- Full suite passes (`pnpm test`, `pnpm typecheck`); new hermetic tests cover both CLI providers with injected fakes (no CLI spawned in CI).
- Verified live against a local Claude Code login: model discovery, connection check, and a structured turn (5s end-to-end, 80 output tokens).


ksokolovic:
I reviewed the exact PR head `84ea4e53531033450da535a82df69cf959378435`. I’m posting this while the PR is still in draft because several findings affect the provider architecture and trust boundaries, and are cheaper to address before the design settles.

## Findings

### [P1] Studio corrupts case-sensitive model IDs

The new model-ID contract preserves the case of the model portion, but several Studio paths lowercase the entire qualified ID:

- `apps/studio/src/components/pipeline/stages/book/BookModelSettings.tsx:33`
- `apps/studio/src/components/pipeline/stages/book/GlobalPromptsSettings/promptSettings.ts:16`
- `apps/studio/src/components/settings/DefaultModelSettings.tsx:33`

This also affects dynamically discovered models. For example, `custom:MyDeployment` is persisted as `custom:mydeployment`, which can select a different model or fail against case-sensitive custom deployments. Three deterministic API tests currently expose the same contract mismatch: the API preserves uppercase model IDs while their assertions expect lowercase.

I recommend using the shared model-ID parser/normalizer everywhere, lowercasing only the provider portion, and keeping prompt-variant canonicalization separate from model-ID normalization.

### [P1] The shipped default provider is not usable in a fresh installation and can bill ambient credentials

`config.yaml` currently defaults to `claude-agent:claude-opus-4-6`. The availability check treats Claude Agent as available because its API key is optional, even when the CLI has not been installed or authenticated. A fresh Docker installation therefore presents a usable-looking default that fails when the first text stage runs.

There is also a billing-boundary issue: the Claude Agent provider falls back to ambient `ANTHROPIC_API_KEY`, and the Codex provider falls back to ambient `OPENAI_API_KEY`, then injects the resolved value into the spawned process. That conflicts with the stated guarantee that ambient tokens cannot be billed.

I recommend adding a real readiness/onboarding check, choosing a default that is reliable in the supported installation paths, and accepting only explicit provider-specific CLI billing configuration for these agent providers.

### [P1] Provider defaults remain stale after settings changes

`apps/studio/src/api/client.ts` caches `/providers` in a module-level promise forever, while `use-provider-credentials.ts` gives the corresponding React Query data `staleTime: Infinity`. Because `/providers` includes mutable default-model configuration, changing defaults does not update stage availability or provider-derived state until the application is reloaded. Invalidating the query alone would still reuse the same promise.

I recommend separating immutable provider manifests from mutable defaults, or removing the permanent promise and explicitly invalidating/refetching provider data after settings changes.

### [P1, security] Custom provider discovery enables SSRF through the unauthenticated API

The custom endpoint validator accepts arbitrary HTTP(S) hosts, including loopback, private, and link-local addresses, and model discovery fetches `${baseUrl}/models`. The API server is not explicitly bound to loopback and has no authentication middleware; CORS does not protect against non-browser clients. Any client that can reach the API can therefore supply `X-Custom-Base-URL` and make the server probe internal services.

I recommend defining the deployment trust boundary explicitly: bind the Electron API to loopback, authenticate network-accessible deployments, and apply appropriate endpoint/DNS restrictions before server-side discovery.

### [P2] TTS and STT bypass the new provider registry

The registry defines speech resolver interfaces, but production paths still hardcode concrete providers:

- `apps/api/src/routes/tts.ts` directly selects OpenAI, Azure, and Gemini synthesizers and hardcodes OpenAI Whisper transcription.
- `apps/studio/src/components/pipeline/stages/languages/LanguageSettings.tsx` hardcodes the same speech-provider list.

Adding a speech provider therefore still requires changes across the API, Studio, pipeline/configuration, and registry, while part of the new speech registry is effectively duplicate/dead abstraction. I recommend routing speech through qualified provider IDs and manifest-driven UI, or narrowing this PR’s registry scope to text/image providers until speech is migrated end to end.

## Verification and merge state

- TypeScript build passed.
- Runtime build and API server bundle passed.
- Lint completed with 0 errors and 9 warnings.
- Provider-focused tests passed: 305/305.
- Full suite: 2,511/2,519 passed. Three failures are deterministic model-ID casing mismatches. Five `part-service.test.ts` tests hit the parallel five-second timeout but passed 13/13 when rerun alone, so I consider those load-sensitive rather than a demonstrated regression.
- I did not run live paid-provider calls or build the Docker image.
- GitHub currently reports the PR as `CONFLICTING` / `DIRTY`. A merge simulation against `develop` produced conflicts in 20 files, including central API pipeline routes/services, Studio client/settings code, locale catalogs, and shared types.

I recommend resolving the P1 findings and rebasing/merging `develop` before moving this PR out of draft. The broad provider direction is valuable, but the current gaps affect correctness, fresh-install usability, credential isolation, and the security boundary.

zignaggo:
1. All p1 findings was fixed
2. Gemini is working in entire pipeline
3. Conflicts solved

elasticsounds (commented):
## Review summary (pre-merge)

**Verdict: the provider-agnostic architecture is the right approach and the OpenAI happy path is preserved.** Ports-and-adapters with Zod-validated provider manifests, a frozen registry, and capability-driven strategy selection — call sites describe their schema (`recursiveSchema`/`looseSchema`) instead of hardcoding provider quirks. Credential headers, localStorage keys, env fallbacks, and the `openai:gpt-5.4` default are all unchanged, and the v2 cache key ships with an exact legacy-v1 dual-read + promotion, so existing books keep their cache hits (no re-generation cost after merge).

### 1. Deliberate behavior changes on the generation path

- **Activity answers** (`packages/pipeline/src/render-llm.ts`): marked `looseSchema: true` because the answer `value` is a scalar union (string | boolean | number) that Gemini's strict response schema rejects outright. The strategy selector now routes this call through non-strict JSON mode on all providers (including OpenAI, which previously used strict native schema). Output is still schema-validated client-side, and old cache entries still hit via the legacy dual-probe.
- **Section render `maxTokens`: 16384 → 32768** — headroom for longer renders; cache-compatible (the v1 key never included maxTokens).
- **Page sectioning**: `mode: "json"` → `recursiveSchema: true` — identical wire request on OpenAI, now portable to other providers.
- **Temperature** is stripped for reasoning models (gpt-5.x / o-series) instead of passed through.
- **Stage runs validate credentials before `beforeRun` clears stage data** — a run that can't make a single LLM call can no longer wipe stage output first.
- The per-request `process.env.OPENAI_API_KEY` mutation hack is gone (it was a race hazard that forced serial re-renders); credentials are request-scoped end to end.

### 2. Develop merge + minor bug fixes (d86097a47)

- The develop merge (multiple-voices speech PR) brought in tests still using the pre-refactor `apiKey`/`geminiApiKey` option fields, crashing `runSpeechStep` on `Object.keys(options.credentials)` — 3 red tests. All 10 stale call sites migrated to the `credentials` shape. (Test files are excluded from `tsc --build`, which is why typecheck missed it.)
- The Anthropic manifest defaulted to `claude-opus-4`, which is not a valid API model id (Anthropic ids are suffixed) — now `claude-opus-4-6`, matching the Studio model picker.
- Full suite: 3,126 tests, green after the fixes (5 `part-service` failures under full parallel load pass in isolation — pre-existing flakiness, not this branch).

### ⚠️ Bundled Claude Agent SDK adds ~75 MB to every download — consider lazy loading

The **claude-agent** provider bundles `@anthropic-ai/claude-agent-sdk`, which ships a ~245 MB per-platform `claude` binary. npm only installs the current platform's binary, so the cost is **~245 MB on disk / ~75 MB compressed per installer and per Docker pull** (not 2×250 MB) — but every user pays it, and the vast majority will use OpenAI or Gemini directly and never touch this provider.

Note the **codex** provider already solves this correctly: it deliberately ships **no** SDK and spawns whatever `codex` CLI is on the user's PATH, reporting "Codex CLI not found" otherwise. Suggested follow-up: make claude-agent symmetric — detect an existing Claude Code install on PATH (or lazy-install the SDK on first enable) and drop it from the default bundle. That would remove the entire size increase for everyone who doesn't opt in.

### Deferred to a separate pass

- Security review of the local-CLI providers (they expose the host machine's logged-in CLI credentials to anyone who can reach the API — fine for single-user desktop, needs thought for shared/server deployments).

elasticsounds:
## Follow-ups deferred to a later PR

Items surfaced by the review of this branch that were deliberately not addressed here. None block merge; listed roughly by priority.

### Correctness / UX

1. **`base_prompt_model` is only honored by PromptViewer.** GlobalPromptsSettings, `promptSettings.ts`, PromptModelActionsDialog and DefaultModelSettings still assume the hardcoded base (`openai:gpt-5.4`), so a configured base model shows false "missing prompt" badges and a spurious warning while the runtime treats it as the base. (`apps/studio/.../PromptViewer/promptModel.ts` call sites)
2. **Keyless speech run with `word_highlighting` silently persists empty timestamps.** Now that speech-only runs no longer require an OpenAI key, a run with only Azure/ElevenLabs credentials overwrites the latest `tts-timestamps` version with `entries: {}` and reports the step green — Whisper's key requirement should fail the step or skip persistence. (`apps/api/src/services/stage-runner.ts` ~2862, ~3612)
3. **LanguageView gates speech runs on the global provider, SpeechLandingPage on the book's** — the two screens can disagree, disabling a run that would succeed. (`LanguageView.tsx:316`)
4. **Google discovery leaks non-generative models into pickers.** Catalogue entries whose `supportedGenerationMethods` lack `generateContent` (embeddings, imagen, AQA) collapse to "unknown" and pass every modality filter — return the empty array instead of `undefined`. (`packages/llm/src/providers/google/index.ts:78`)
5. **Render `maxTokens` raised to 32768 with no cap awareness.** Models capped lower (e.g. `gpt-4o` at 16384, the old value) now hard-fail every render and retries resend the same value. The new `StructuredTextCapabilities.maxOutputTokens` field has zero consumers — populate and clamp against it, or revert the raise. (`packages/pipeline/src/render-llm.ts:91`)

### CLI provider robustness (edge cases)

6. **Signal-killed child mapped to exit code 0** (`exitCode ?? 0`): codex can accept a truncated run as a successful turn; claude-agent mislabels the failure. (`codex/cli.ts:142`, `claude-agent/cli.ts`)
7. **Codex abort/timeout misreported as "Codex CLI could not be started"** — missing the AbortError branch its claude-agent twin has. (`codex/cli.ts` `describeSpawnFailure`)
8. **Codex `rm(workDir)` in `finally` can mask the real error** on Windows (EBUSY on a dying child's cwd) — wrap in try/catch as the claude-agent runner now does.
9. **Codex has no CLI version gate.** Old CLIs pass the `codex login status` health check, then every turn fails on unknown flags (`--ephemeral`, `--ignore-user-config`). Consider a version floor mirroring claude-agent's.
10. **`--json-schema` still travels on argv** (the CLI has no file variant) — residual Windows 32k risk for very large native-strategy schemas now that the system prompt goes by file.
11. **Agent-loop turn log drops one transcript message** when a run aborts mid-tool-execution (`transcriptLength` assumes the tool-results message was already pushed). (`packages/llm/src/agent-loop.ts`)

### Cleanup

12. **`elevenLabsApiKey` legacy field accepted but silently dropped** by `toProviderCredentialValues` — map it or delete the field and parameter. (`apps/studio/src/api/client.ts:296`)
13. **Dead middleware helpers**: `hasCredential` and the `credentialValue` re-export have no callers; `hasCredential` is also the header-only presence trap c1db8007 fixed. (`apps/api/src/middleware/provider-credentials.ts`)
14. **Unused new `@adt/llm` exports**: the `validateConfigModels`/`assertConfigModels`/`collectConfigModelChecks` path, `describeCredentialPresence`, `defineAgentTool`, and `assertSupportedModel`/`isSupportedModel`/`qualifyModelId`/`providerIdOf` have no consumers outside tests — delete or adopt.
15. **`PROVIDER_HEADER_NAMESPACE` documents an unenforced invariant** — enforce it in `CredentialHeaderName` or drop the constant. (`packages/types/src/ai-provider.ts:51`)

### Infra / future work

16. **api vitest project timeout flake**: `part-service.test.ts` and `stage-runner.test.ts` exceed the 5s default under full parallel load (they pass in isolation) — raise the api project's `testTimeout` the way this branch did for studio. (`vitest.config.ts`)
17. **Agent system prompts are compiled-in constants** (`packages/agents/src/prompts/*.ts`) — not per-model tunable and invisible to the Prompts UI. Moving them into the prompts directory would let them flow through the model-variant system and match the transparency principle.
18. **The curated codex model list has no discovery source** (the CLI offers no catalogue command) — it will need a manual bump when the next model generation ships. (`packages/llm/src/providers/codex/models.ts`)

PR #734, opened 2026-08-05
Feature: New inicial UI

zignaggo:
### Summary

This PR imports the redesigned Studio interface prototype into the app as a fully functional UI, living under the `/redesign` route tree and enabled by default. Users can switch between the new and classic interfaces at any time via a UI version toggle (persisted in `localStorage`, `new` is the default).

### What's included

**App shell & navigation**
- New app shell with a collapsible sidebar (`AppSidebar`) and solo-screen layout (`RedesignLayout`), built on Tailwind + shadcn primitives
- Command palette and Add Book dialog rebuilt on shadcn `Dialog`, with proper book cover rendering
- View state migrated from search params to type-safe TanStack Router routes (`/redesign`, `/redesign/library`, `/redesign/handoffs`, `/redesign/settings/*`)

**Screens**
- **Home** — first-run welcome hero, feature tour, per-section splits summary, and a top bar; element sizes increased for readability
- **Library** — grid/row book views, book detail dialog, and a first-run empty state, rebuilt on shadcn `Input`/`Select`/`Popover`/`SegmentedControl`
- **Split & merge (Handoffs)** — wired to real data with contribution rows, split summary cards, empty state, corrected split range indexing, and working Preview/Export actions
- **Settings** — rebuilt on shadcn `Tabs`/`Switch`/`Select` and split into route-backed sections: Theme, Language, Models (ported model defaults), Providers (customization + API key dialog), Prompts (global prompts port with editor pane and file tree), Notifications (new preferences system with sound chime), and About; help section links updated

**Design system**
- Brand color ramp, stage color tokens, and badge status variants added to `globals.css`
- New UI primitives: `EmptyState`, `GhostCover`, `StagePill`, `StageDiscs`, `Kbd`, `Pager`, `Eyebrow`, `ScreenFallback`
- `ScrollArea` used in place of native scrolling; dark mode fixes for the prompts viewer and segported model defaults), Providers (customization + API key dialog), Prompts (global prompts port with editor pane and file tree), Notifications (new preferences system with sound chime), and About; help section links updated

**Design system**
- Brand color ramp, stage color tokens, and badge status variants added to `globals.css`
- New UI primitives: `EmptyState`, `GhostCover`, `StagePill`, `StageDiscs`, `Kbd`, `Pager`, `Eyebrow`, `ScreenFallback`
- `ScrollArea` used in place of native scrolling; dark mode fixes for the prompts viewer and segment theme colors
- Legacy `redesign.css` removed — Tailwind-only styling

**Beta branding & updates**
- Beta builds now use a dedicated beta logo and purple accent color (detected from the release channel)
- New auto-update modal flow (`UpdateDialogProvider` + `PostUpdateDialog`) and app toaster

**i18n**
- All new strings extracted and fully translated across `en`, `pt-BR`, `es`, `fr`, and `sq` catalogs (~600 lines per locale)

### How to test

1. `pnpm dev` (or `pnpm dev:desktop`) — the app opens on the new UI by default
2. Use the UI version toggle to switch back to the classic interface and confirm both remain functional
3. Walk through Home → Library → Split & merge → Settings sections, including first-run empty states
4. Verify dark mode across the new


apps/studio/src/components/app/screens/settings/providers/useProviders.ts
```diff
@@ -0,0 +1,169 @@
...
+  if (allOptional && hasSecret) return "cli"
+  if (allOptional && !hasSecret) return "local"
+  return "api-key"
+}
+
+
+export function requiredFieldsFilled(descriptor: ProviderDescriptor, creds: Record<string, string>): boolean {
+  return descriptor.manifest.credentialFields
+    .filter((f) => f.required)
+    .every((f, i) => (creds[f.key]?.trim().length ?? 0) > 0 || descriptor.fieldStatus[i]?.configuredOnServer)
```
elasticsounds: **Latent index-misalignment bug.** `i` here indexes the *required-only filtered* array, but `descriptor.fieldStatus` is parallel to the full `credentialFields` array. With a manifest like `[baseUrl (optional), apiKey (required)]`, `fieldStatus[0]` would read `baseUrl`'s `configuredOnServer` for the `apiKey` check → a key already stored on the server reports `missing-credential`.

It's masked today only because every bundled descriptor lists required fields first and all `configuredOnServer` are `false` — but this file's contract says it's built to swap in real server descriptors, where that ordering isn't guaranteed. Match by key instead:

```ts
.every((f) => (creds[f.key]?.trim().length ?? 0) > 0
  || descriptor.fieldStatus.find((s) => s.key === f.key)?.configuredOnServer)
```

apps/studio/src/components/app/screens/settings/searchIndex.ts
```diff
@@ -0,0 +1,224 @@
...
+]
+
+const PROVIDER_ENTRIES: SettingsSearchEntry[] = PROVIDER_IDS.map((provider) => ({
+  id: `settings-provider-${provider}`,
+  kind: "option",
+  section: "providers",
+  label: PROVIDER_META[provider].name,
+  hint: PROVIDER_META[provider].desc,
+  keywords: PROVIDER_META[provider].hint,
+  anchor: providerAnchor(provider),
```
elasticsounds, line 153: **Dead anchor — provider search results silently no-op.** This emits `settings-provider-<id>` (via `providerAnchor`), but the only id the providers screen renders is `prov-panel-<cardKey>` (`ProvidersList.tsx:72`). `useSettingsAnchor` does `document.getElementById(hash)`, retries ~30 frames, then gives up — so selecting any provider from settings search or the command palette navigates but never scrolls/highlights. The providers screen has no alternate hash handling either.

Fix: render `id={providerAnchor(provider)}` on the card (or point `providerAnchor` at the `prov-panel-*` scheme).

apps/studio/src/components/app/AppLayout.tsx
```diff
@@ -0,0 +1,120 @@
...
+import { AddBookDialog } from "./AddBookDialog"
+import { AppShellContext } from "./AppShellContext"
+import { useAppBooks } from "./use-app-books"
+import { Kbd } from "./ui/Kbd"
+import { APP_PATHS } from "./nav"
+
+const SHORTCUTS: { keys: string[]; label: MessageDescriptor }[] = [
+  { keys: ["⌘", "K"], label: msg`Open command palette` },
+  { keys: ["/"], label: msg`Focus settings search` },
+  { keys: ["G", "H"], label: msg`Go to Home` },
```
elasticsounds: **Advertised shortcuts aren't wired.** The help dialog and sidebar advertise `G H`, `G L`, `N`, and `?`, but the only handlers in the app shell are ⌘K (here) and `/` (`useFocusSearchShortcut`). Pressing `?` — the shortcut that's supposed to open this very dialog — does nothing, as do `N` and the `G` chords.

Either wire them up in the `onKey` handler (with an editable-element guard so typing in inputs doesn't trigger them) or trim the `SHORTCUTS` list to what actually works.

apps/studio/src/components/app/data.ts
```diff
@@ -0,0 +1,129 @@
...
+    hex: s.hex,
+  }))
+}
+
+/** Compact relative time — "2h ago", "yesterday", "3 days ago", else a date. */
+export function formatRelative(iso: string, locale: string): string {
+  const then = new Date(iso).getTime()
+  if (Number.isNaN(then)) return ""
+  const diffMs = Date.now() - then
+  const mins = Math.round(diffMs / 60_000)
```
elasticsounds: **`Math.round` overstates elapsed time** at every unit boundary: 30s shows "1m ago", a 90-minute-old edit shows "2h ago", ~23.6h shows "yesterday", and 6.5 days rounds to 7 and falls through to an absolute date. Relative timestamps conventionally floor — `Math.floor` on lines 81/84/86 fixes all four.

apps/studio/src/components/app/screens/settings/providers/useProviders.ts
```diff
@@ -0,0 +1,169 @@
...
+  }
+  return out
+}
+
+function getSnapshot(): Creds {
+  if (snapshot === null) snapshot = readStorage()
+  return snapshot
+}
+
+function subscribe(listener: () => void): () => void {
```
elasticsounds: **Missing cross-tab sync.** The sibling pref stores (`use-notification-prefs.ts:60`, `use-library-prefs.ts:49`) add a `window.addEventListener("storage", …)` to re-read and notify on external localStorage changes; this store caches a module-level `snapshot` and only mutates it via same-tab `writeCredential`. Editing an API key in one window leaves another window stale until reload. Minor since this module is prototype scaffolding, but worth carrying over the same listener pattern when the real backend wiring lands.

elasticsounds (commented):
Ran a pre-release check + correctness review on this branch (verified each finding against the source, twice — two independent review passes agreed).

**Mechanical gates: all green** ✅
- `pnpm typecheck` ✅  `pnpm lint` ✅  `pnpm test` ✅ (5 timeout flakes in `part-service.test.ts` under CPU contention — backend file untouched by this PR; 14/14 pass in isolation)
- i18n: catalogs in sync after extract, all 5 locales fully translated (0 missing of 3478)

**Findings: 5 confirmed** (inline comments below) — triage:
- 🔴 Live user-facing: provider search anchors are dead links; advertised keyboard shortcuts (`G H`, `G L`, `N`, `?`) aren't wired
- 🟡 Latent trap: `requiredFieldsFilled` indexes `fieldStatus` by the filtered index — masked by current bundled data ordering, breaks when real server descriptors land
- ⚪ Polish: `formatRelative` rounds instead of floors; credential store misses the cross-tab `storage` listener its sibling hooks have

Also note the branch is ~49 commits behind `develop` — worth merging up before this lands.

Eliezir:
Thanks for the review @elasticsounds — all five are fixed. Each one was reproduced in the running app first, so every fix has a before/after rather than just a code change.

### Checklist

- [x] **1. `fieldStatus` index misalignment** — `providers/useProviders.ts` · `ac786d9d1`
  `i` indexed the required-only filtered array while `fieldStatus` is parallel to the full `credentialFields`. Now matched by key. Reproduced in the app with an optional-first manifest, and it failed **both** ways: a key stored on the server reported `missing-credential`, and — worse — an unrelated optional field on the server marked the provider "Authenticated · API key" with no key present at all. 6 tests cover the alignment and pin the two conditions that mask it today.

- [x] **2. Dead provider anchor** — `settings/searchIndex.ts`, `nav.ts`, `ProvidersList.tsx` · `29e7cce34`
  Confirmed dead: searching "OpenAI" navigated to `/settings/providers#settings-provider-openai` and `getElementById` returned nothing, from both settings search and ⌘K. The anchor is now rendered on the provider row, and the card **expands** on match — scrolling to a collapsed row would still have hidden the credential field the search pointed at. `providerFromAnchor` is the inverse of `providerAnchor` over one shared prefix so the emit and render sides can't drift again, with a test asserting every provider search entry resolves to a card the screen renders. Rows also took over the corner clipping from the group card, because the anchor flash ring was being cut off by its `overflow-hidden`.

- [x] **3. Advertised shortcuts aren't wired** — `AppLayout.tsx`, `AppSidebar.tsx`, `CommandPalette.tsx` · `b71d36790`
  Trimmed to ⌘K — see the note below.

- [x] **4. `Math.round` overstates elapsed time** — `components/app/data.ts` · `0746883a2`
  Floored all three steps. Nested flooring is exact, so the chain no longer compounds the way rounding a rounded value did. Confirmed in-app on a populated library ("Pick up where you left off · 5 days ago"). Before → after at the boundaries:

  | elapsed | before | after |
  |---|---|---|
  | 30s | `1m ago` | `just now` |
  | 1h30m | `2h ago` | `1h ago` |
  | 23h36m | `yesterday` | `23h ago` |
  | 6d12h | `Aug 24, 2026` | `6 days ago` |

- [x] **5. Missing cross-tab sync** — `providers/useProviders.ts` · `2d3f69fae`
  Added the `storage` listener. Two differences from the sibling pref stores, because their shape doesn't fit: credentials live under **one key per field**, so the listener matches against the whole set rather than a single `STORAGE_KEY`; and it also re-reads when `event.key === null`, which is what fires when another window calls `localStorage.clear()`. Proved with two tabs sharing one profile — tab A saved a key through the UI and tab B went from "Not configured" to "Authenticated · API key" with no reload.

### On the shortcuts (#3)

We removed them rather than wiring them. `G H`, `G L`, `N` and `?` are gone from the help dialog and the sidebar, leaving **⌘K**, which is the only one the shell actually handles. `/` was dropped from that list too — it works, but only inside Settings, so advertising it in the global dialog was misleading anywhere else. The dialog now carries a "still in development" note naming what's coming, so the gap is stated rather than implied.

The intent is to bring them back on [TanStack Hotkeys](https://tanstack.com/hotkeys/latest) rather than hand-rolling chord handling and editable-element guards — it gives us typed bindings, key sequences for the `G` chords, and a `Mod` key that resolves to Cmd/Ctrl per platform. It's still alpha, which is the other reason we didn't want to half-wire this now.

One thing that came out of #3: ⌘K and Ctrl+K both already worked, but the printed glyph was hardcoded to `⌘` in **four** places (help dialog, sidebar search hint, palette footer, and the `?` chip). Those now derive from the existing `usePlatform()` hook — which prefers Electron's real `process.platform` over the user agent — so Windows and Linux see `Ctrl` and `Ctrl+K` rather than `⌘K`.

### Also in this push

While verifying the above we hit two unrelated problems and fixed them here, since both sit in code this PR already touches:

- **Onboarding could render in the main window** (`e6ceb2a3d`). Completion was tracked twice — `userData/onboarding-state.json` in the main process and a `localStorage` flag in the renderer's route guard. When they disagreed, the main process opened the app window and the guard immediately redirected it to `/onboarding`, so the tour rendered at 1400×900 instead of its own 900×620 window, with no way out but finishing it. The renderer now asks the main process through the `onboarding:get-status` bridge that already existed and was never called.
- **Welcome empty state** (`4ccbc7d84`, `6164de689`). The content was centred in the viewport rather than top-aligned like every other screen, and top-aligning it left the bottom third empty. The copy now occupies a cell of the same grid as the outputs, so nothing is left over. Also a responsiveness pass across 4 viewports × 19 page/state combinations, which caught one clipped CTA at 1024px.

The branch is also merged up to date with `develop` (was 52 behind); the only conflicts were the Lingui catalogs, resolved as a union.

Catalog extraction for the changed strings is still to come before merge.

PR #777, opened 2026-08-11
feat(storyboard): reorder pages in the page list

### LINKED ISSUE #660: [Storyboard] Ability to re-order pages in the page list
### Preliminary Checks

### Problem / motivation

The intention of this feature is to be able to re-order pages in the storyboard stage. This would enable the user to improve the order of some sections that were not in the best position in the original PDF.

### Proposed solution

Add an option to move pages in the page list and in the overview page. This could be seen as a handle in the overview page where the user can move pages around, or a three-dot menu in the page list where the user can select to re-order the pages (as well as other page-specific options, like hide/unhide).

### Alternatives considered

I did not find any alternative to move pages around in the ADT Studio. This could be done (partially) in the original PDF. If the page in the storyboard is a section of a full page in the original PDF, this cannot be done.

Pipeline — Storyboard

### Additional context

_No response_

## Technical design proposal

Treat source-page identity, output identity, and reading position as separate concepts:

- `pageId` / `pageNumber` remain immutable source-PDF identity and provenance. Reordering must never rename source pages, images, nodes, or other page assets.
- Each storyboard section has an immutable `sectionId` and represents one output page.
- Each quiz receives an immutable stored `quizId` and also represents one output page. IDs such as `qz001` must not be derived from the quiz's current array position.
- A versioned, book-level reading-order entity stores the ordered list of output IDs. The position in this list is the output position; it must not be written back to `pageNumber`, `sectionIndex`, or an asset name.

For example:

```ts
{
  items: [
    { kind: "section", id: "pg001_sec001" },
    { kind: "quiz", id: "qz_a83f..." },
    { kind: "section", id: "pg003_sec001" }
  ]
}
```

Store this through the existing versioned `node_data` mechanism (for example, node `reading-order`, item `book`) so every reorder can be rolled back. When no saved order exists, derive the default from source-page and section order.

### Stable identity requirements

Current structural operations renumber section IDs within affected pages. That behavior must change before reading order can safely reference sections:

- Clone: the original retains its ID; the clone receives a fresh, never-reused ID and is inserted after the original in reading order.
- Split: the first half retains the original ID; the second half receives a fresh ID and is inserted after it.
- Merge: the surviving section retains its ID; the removed section's ID is retired and removed from reading order.
- Cross-page merge: the target retains its ID; the source ID is retired.
- Hard delete: retire only the deleted ID; never renumber surviving sections.

`sectionIndex` may remain a transient array/display position, but persistent relationships such as renderings and editable activities should use `sectionId` as their canonical key.

### Reordering and removing sections

- Dragging an item creates a new reading-order version only. It must not move section data between source pages or modify assets, LLM outputs, or source provenance.
- The panel's normal remove action should be reversible pruning/hiding rather than destructive deletion.
- A pruned section remains in the reading-order list but is omitted from preview and export. Unpruning restores it at the same position with all assets intact.
- A truly destructive deletion, if retained, should be a separate explicit action.

### Preview and export behavior

Create one shared reading-order resolver and use it in the storyboard list, overview, live ADT preview, TOC construction, quiz placement, and packaging. Avoid implementing separate ordering logic in each consumer.

Packaging should iterate the resolved reading order, write `content/pages.json` in that order, and keep stable section/quiz-based filenames. WebPub `readingOrder`, EPUB/PNLD spines, next/previous navigation, page-position metadata, and sign-language page mappings should all be derived from the same resolved sequence. Files must not be renamed according to their new position.

### Acceptance criteria

- [ ] Users can reorder storyboard sections and quizzes by drag-and-drop or equivalent accessible controls.
- [ ] Reordering changes preview navigation and all exported reading orders/spines.
- [ ] Source `pageId`, source `pageNumber`, section/quiz IDs, image IDs, filenames, and stored assets are unchanged by reordering.
- [ ] Clone, split, merge, cross-page merge, prune/unprune, and delete reconcile reading order without renumbering surviving IDs.
- [ ] Cloned and split output pages receive unique, stable IDs.
- [ ] Quizzes have stored stable IDs and can be moved independently.
- [ ] Removing from the book is reversible and restores the item in its previous position.
- [ ] Reading-order changes are versioned and can be rolled back.
- [ ] Books without a reading-order entity retain the current source-derived order.
- [ ] ADT, WebPub, EPUB, and PNLD exports agree on the resulting sequence.
- [ ] Tests cover reorder, clone, split, merge, prune/unprune, quiz movement, rollback, and export order.

gbergengruen:
Closes #660.

Lets the user rearrange the pages of a book in the Storyboard stage, so the reading order no longer has to match the order the source PDF happened to use.

> **This PR was split.** It previously carried 11 commits and ~5,000 lines. The four standalone fixes underneath it now live in their own PRs, and **all four have merged**:
>
> | | |
> |---|---|
> | #784 | **stable sectionIds** — structural ops renamed neighbouring sections, reattaching sign-language videos, TOC entries and TTS audio to the wrong content |
> | #785 | **stable quizIds** — `qz001` was an array index, so adding a quiz re-pointed later quizzes at another quiz's translations and audio |
> | #786 | **one reading-order resolver** — the ordering walk existed five times and had drifted |
> | #787 | **id-based output filenames** — `index.html` was whichever page came first, giving one page a position-dependent URL |
>
> This PR now sits directly on `develop` and its diff shows only the feature.

## How it works

Source-page identity, output identity and reading position are kept as three separate things:

- `pageId` / `pageNumber` stay immutable source-PDF provenance. Reordering never renames a page, image, node or any asset.
- Each storyboard section has an immutable `sectionId` and is one output page; each quiz has an immutable stored `quizId` (#784, #785).
- A versioned, book-level `reading-order` entity holds the ordered list of those ids. Position in that list is the output position, and is never written back into `pageNumber`, `sectionIndex` or a filename.

One resolver (#786) turns that into the book's sequence, and everything downstream consumes it — packaging, live preview, both TOC builders, the TOC section picker, quiz placement and the sign-language position map.

Storage needs no migration: `node_data.node` is a free-form column.

## Where you can reorder

**The page list** — drag (behind a `Rearrange` toggle), `Alt+Arrow`, or Move up / Move down in each row's three-dot menu.

**The overview** — a `PDF order` / `Book order` toggle. `PDF order` is the table as it has always been, grouped under each source page, which is what the structural operations reason about. `Book order` is a flat list in reading order, and it is the only view where Move up / Move down are offered, because it is the only one where the row the user moved visibly moves. Both views show a `Book page` and a `PDF page` column.

The step arithmetic is shared between the two screens. The displayed list and the stored order are not the same list — each screen shows only the slots it can resolve — so a step is measured in visible rows and then anchored back onto the row it should land beside, rather than applied to the stored order as a raw offset.

## Decisions worth a look

**Reconciliation is lazy.** The stored order is reconciled against the book's current contents on every read and never written back. That keeps the entity's version history a log of deliberate reorders instead of machine-generated churn, means a reconcile bug can never be persisted, and leaves clone, split, merge, delete, prune, spread-apply, extract and quiz-add with no reading-order code at all.

That last property comes from one rule: an id the book has but the stored order does not is inserted at its **default-order neighbourhood**, after the nearest preceding sibling that survived. A clone's default position is right after its original, so that is where it lands, wherever the user moved the original to. The trade-off — a genuinely new last page also follows its sibling rather than going last — is deliberate and tested: the two are indistinguishable to the reconciler, and keeping a clone beside its original matters more.

**Slots come from sectioning, not from rendering.** Which slots exist is a question about what the book contains; whether a slot produces an output page is a separate one. `order` is every section; `items` is the output pages — unpruned *and* rendered. Conflating the two is what caused the worst bug found in review (below).

**`reading-order` is deliberately absent from every stage's output nodes.** It is user intent, not stage output — listing it in `EXTRA_STAGE_OUTPUT_NODES` would make every storyboard re-run delete it irrecoverably. There is a test asserting no stage clears it, because the mistake is one line and invisible until someone loses work.

**A reorder invalidates the bundle and the accessibility assessment, and nothing else.** Clearing the storyboard dependency chain would destroy the user's generated speech, translations and glossary for a drag-and-drop.

**A rearrangement is a pending change.** Moves accumulate and are committed from the shared floating save bar, with Save / Discard and the navigation guard, the same as every other edit in the app. One arrangement is therefore one version, rather than one per drag. The draft lives in a provider above both surfaces — the sidebar and the overview are mounted together and rearrange the same book, so a draft held inside either would be invisible to the other.

**Dragging is opt-in.** The page list is primarily something you click through, so a stray drag must not rewrite the book. Each row's three-dot menu can always move it; drag lives behind a `Rearrange` toggle. `Alt+Arrow` stays live in both modes since it already needs a modifier.

**Removal is reversible.** "Remove from book" sets `isPruned`: the page keeps its slot in the order and all its content, and is still listed in the sidebar (greyed, showing `–` instead of a book page) so it can be put back exactly where it was. Hard delete is a separate, explicitly permanent action whose confirmation names the reversible alternative.

**Two page numbers, both named.** The sidebar and hover card say "Book page N" and "PDF page N" rather than one ambiguous "Page N" — they are the same kind of thing measured in two documents, and after a reorder they simply differ.

**Quiz batching follows reading order.** Both quiz runners built their input from `storage.getPages()` — source-PDF order — and sliced it into groups of N. With six pages at three per quiz reordered to `F E D C B A`, the batches stayed `(A,B,C)`/`(D,E,F)` and anchored to pages now at reading positions 4 and 1: quizzes covering unrelated content at uneven intervals.

**The row menu builds on #776.** `ActionMenu` portals its content, so an open menu in the virtualized list is no longer painted under the rows that follow it and this PR needs no stacking-context handling of its own.

## Fixed since the first review

An audit of the whole stack against #660's acceptance criteria, plus testing in the app, turned up five defects. All are fixed here with tests.

**Pruned pages disappeared after a storyboard re-run** (`44fb0e5a`) — the worst of them, and it destroyed access to real content. The resolver built its slots from `web-rendering` rather than from the sectioning tree. Rendering skips a section that is pruned or that has nothing renderable in it, so pruning alone looked fine — the old HTML stayed in the node — but re-running the storyboard regenerated `web-rendering` without those sections, their slots vanished, and there was no row left to put them back with. The same hole silently hid any section with no renderable content, pruned or not.

**Re-adding a page left the reading order stale** (`52cec9eb`) — you had to move an unrelated page to make it catch up. `reRenderPage` only submits a task, so the order was invalidated while the section still had no rendering, and nothing invalidated it again when the render landed. The order is now part of `invalidateStoryboardDependents`, which every structural operation and the re-render completion handler already route through — closing the same gap for clone, split, merge, cross-page merge and delete, none of which had ever invalidated it. Stage completion never refreshed it either, and `sectioning`, which owns the slots, had no case at all.

**EPUB and PNLD navigation ignored the reading order** (`ebb6ffe6`) — `toc.json` and both NCX builders followed a reorder, but the two nav documents read the stored `toc-generation` node in whatever order it held. An EPUB shipped a nav that disagreed with its own spine. Directly contradicts the acceptance criterion that all four exports agree.

**Legacy quiz ids could shift** (`ff0ef3d7`) — `generate-one` sorted by reading rank *before* stamping ids, so a quiz predating #785 could take another quiz's id and inherit its translations and audio. Harmless while the sort key matched the derivation order; not once it became reading rank. Also fixes a missing anchor page sorting a quiz to the front rather than the end.

**Three labels claimed "Removed from the book"** from a missing position alone (`14d7a884`), which became wrong once unrendered sections were visible again — one would have rendered the literal string `Book page null`. Same commit bumps two row spans the stack had left at 7 after adding the `Book pg` column.

## Fixed during the browser pass

Testing in the app turned up five more things, all fixed here.

**There was no way back to the original order.** The entity is not written until the first rearrangement, so its v1 is already a rearrangement and the version history held nothing representing the order the book started in. The picker now offers `Original — PDF order`, which recomputes the source-derived order and saves it as a new version — recomputed rather than recovered, because a page added since the first reorder belongs in it.

**The API and the UI disagreed about when reordering is blocked, in both directions.** The server refused while *any* step was running; the sidebar greyed out only for the storyboard. So a reorder during a captions run was rejected by a server with no reason to care, and a reorder during quiz generation — which really does rewrite the quiz slots — was offered as though it were fine. `READING_ORDER_BLOCKING_STEPS` now names the four steps that change what the order is made of, and both sides read it.

**A refused save was silent.** The row slid back with no message, which reads as the drag having missed rather than the save having been refused. Both mutations now surface the server's message, which names the steps in the way.

**The overview's page column showed `pg001`, then showed the wrong number.** It is now named `PDF page` beside `Book page`, matching the sidebar's wording — and it counts *sheets*, not the number printed on the page. Those differ on any book with unnumbered front matter: in `momo-p001-010` the fifth sheet prints "1", so reading the printed number left the first four rows showing a dash and the rest offset by four.

**A reorder made from the sectioning stage could not be undone there.** Rows can be moved from the overview table in either stage — one component, move wired inside it — but the history lives in the storyboard sidebar, which sectioning does not have. That overview now carries its own picker.

## Known limits, not worked around

- **Quiz placement is anchored to a page** (`Quiz.afterPageId`), so if a reorder splits one source page's sections apart, its quiz sits after that page's *first* run of sections. Exact placement needs `afterSectionId`, a schema change not taken on here. Filed as #871.
- **The "Pages per Quiz" setting counts source pages, not sections.** On a multi-section page those differ. Changing the unit is a product decision, tracked in #778.
- **Quizzes have no `isPruned`.** Sections got reversible removal; removing a quiz is still a real deletion, undoable only by rolling `quiz-generation` back. Filed as #872.
- **Quizzes anchored to a page the book no longer has are dropped from the output** rather than placed last, which is what `orderQuizzesForInsert` does with the same case. Pre-existing from #786; this PR puts the two behaviours side by side. Filed as #860.
- **`maxNodeVersionExcluding`, added to `develop` by #851, cannot see a reorder.** Node versions are per-entity, so `reading-order`'s first save writes version 1 and never moves a max that other nodes have already pushed past. Nothing calls it yet; flagged for the publish stack as #861.
- **Deferred from the 2026-09-16 review**, as new surface rather than defects in this PR: Sectioning's Slides / PDF pages sidebar tabs (#868), the minimal `7 / 42` header (#869, blocked on this), and reporting reading position as a `#slide=N` fragment (#870). The review also asked that navigation forwarded from the preview iframe follow the shared sequence; there is no such channel today, so it is filed as #873 rather than invented here.
- **`StoryboardIndex.tsx` is ~1,060 lines and `SectioningOverview.tsx` ~1,300**, and both should be split into folders. Deliberately *not* done here — most of those lines are pre-existing and would render as additions, burying this PR's actual changes. Filed as #782, to be done immediately after this merges.

## Fixed since the second review

A backend audit of the merged stack turned up five more things, all fixed here.

**Both quiz runners shared a twenty-line page-gathering walk, and the test meant to pin it had written it out a third time** — its header said so outright, on the grounds that no shared helper existed. That copy had already drifted, calling `batchPages` with two arguments where both runners pass a third (`quizSectionTypes`), so the section-type filter was untested and a runner reverting to `storage.getPages()` would have left the test green. There is now one `gatherQuizPageInputs`, called by both runners and by the test.

**Two quiz diagnostics were quietly broken by the move to reading order.** A page in the reading order has rendered content by construction, so `missingRendering` could no longer increment; and the denominator became "pages that reach the reader", making the no-quizzes warning read *"0 of 0 pages"* for exactly the book it exists to explain — one whose storyboard has never run.

**The two new routes created a book directory for any label you asked about.** `createBookStorage` mkdirs unconditionally, so `GET /books/typo/reading-order` wrote `books/typo/` to disk and returned a 200 describing nothing. Every comparable route already guards; these did not.

**The `PUT` saved in two writes with no transaction** — the new version, then the dependents it invalidates. A failure between them leaves a reordered book pointing at a bundle built for the old sequence.

**Reconcile rule 3's docstring described behaviour the code does not have**, promising a newcomer with no surviving predecessor lands "before the nearest following sibling, else appended". It goes to the front of the stored order, and "else appended" is unreachable. For a book the user has reversed, the two readings differ.

## Fixed since the third review

[The review of 2026-09-16](https://github.com/unicef/adt-studio/pull/777#issuecomment-5698949571) asked for six things. Three were defects in what this PR already shipped and are fixed here; three are new surface and are filed as #868, #869 and #870.

**A full Sectioning rebuild silently rebound the arrangement to different content.** The rebuild clears `page-sectioning`, which is where the section-id high-water mark is read from, so the next run re-mints ids densely from `_sec001` — for other content. A stored order matches by id string alone, so it reconciled *cleanly* onto whatever now held those ids. Nothing downstream could notice, because no slot was missing. There is now a `sectioning-generation` counter, bumped at the rebuild boundary and stamped onto every order saved after it; an older generation is ignored by the resolver and refused by the restore route. The rows are kept, so the history stays visible and the picker can strike them through and say why. Two deliberate exemptions: an unstamped order predates the stamp and is honoured rather than invalidated on upgrade, and a re-extract wipes the counter along with the reading order itself, so nothing survives to be stale against. `clearsSectionIdHistory` moved into `@adt/types` as `rebuildsSectionIds` so the warning shown before the run cannot drift from what the backend will actually reset.

**Merge acted on the source-order neighbour, not the one on screen.** Every structural op addressed a section as `(pageId, sectionIndex)` plus a direction. With the book arranged `A → C → B`, "Merge with next" on A merged A+B: the row the user was looking at and the row that vanished were different rows. The displayed neighbour can also live on another page, so *which operation applies* is itself a function of what the list showed — something an index cannot express. Merging is now addressable by id (`POST /books/:label/sections/merge`), and both index routes resolve their neighbour and call the same two implementations so they cannot drift. Extracting them removed the adjacency assumption, which was only ever load-bearing in the guards, and surfaced a latent bug: the survivor shifts down when the removed section sat before it, unreachable while every merge was adjacent. The confirmation now names the destination rather than saying "next section", which in a reordered book is neither true nor findable.

**Navigation contradicted the sidebar beside it.** The page list was built from the reading order while the arrows, the keyboard and the header walked `usePages()` — source-PDF order — so "next" went somewhere other than the row highlighted next to it, and on a quiz both arrows were dead. There is now one sequence, draft-aware, read by the sidebar and by navigation alike. The selected slide moved into the URL as `?section=<sectionId>`, so a reload, a Back and a shared link land where the user was; `skipNextResetRef` went with it, because page and section now move in one navigation. Three regressions were caught and fixed in the process: a structural edit names its target by an index into a tree the view has not received yet (so the index is parked until it arrives); the keyboard handler's dependency list predated the sequence, so keys and toolbar disagreed once the order loaded; and stepping onto a quiz replaced a history entry, making Back skip the slide it came from.

Captions and the plain page list keep the numeric section selection on purpose — Captions has only a section *count* to work from, never the section list, so it cannot address one by id.

## Follow-up: test gaps to close after this merges

Everything previously listed here is now covered, bar one. What is left:

- **The in-route quiz page sort** (`quizzes.ts`), whose own comment says it must not disagree with `orderQuizzesForInsert`, is duplicated inline in the handler and cannot be reached without an LLM. `orderQuizzesForInsert` itself is now tested against a reversed reading rank.

## Verification

`pnpm typecheck` clean · `pnpm test` 3734 passing (293 files) · `pnpm lint` 0 errors · `lingui extract` 3477 messages, 0 missing across `en`/`es`/`fr`/`pt-BR`/`sq`.

The load-bearing test is end-to-end: packaging a book whose stored order puts page two first changes the sequence **and renames nothing** — each page keeps the filename its id gives it, its source `page_number` is untouched, `index.html` follows the order, and the dock counter reflects the new positions. That is #660's core requirement.

Also covered: reconcile rules including idempotence and the clone-follows-its-original case; drag/keyboard/menu interaction including that dragging is inert until switched on; a pruned section keeping its slot across a re-run that drops its rendering; and EPUB/PNLD nav order matching the spine.

## Please check in a browser before approving

**Confirmed in the app:** removing a page and adding it back, and the reading order updating as it should when you do.

**Since run in a browser:** the reorder interactions, removal and re-adding, the overview's two views, the version history, and the reset — which is what turned up the five fixes above. Still unexercised at the time of writing: the edge auto-scroll on a long book, quiz batching after a reorder, and packaging order end to end.

**Originally not run:** everything else. Types, tests and translation completeness cannot tell you whether the drop indicator is visible, how the edge auto-scroll feels, whether `[2] PDF 1` reads well at 10px in a narrow column, whether the toolbar is crowded now that it holds both the version picker and `Rearrange`, or whether the overview's `PDF order` / `Book order` toggle reads clearly next to the existing panel filters. Those are the parts most likely to need adjusting.


elasticsounds:
## Review notes: complete the shared reading-order behaviour

The separation between stable content identity and reading position is the right foundation for #660. A normal reorder preserves section IDs, source-page identity, image references, HTML `data-id`s, and TTS mappings. Keep that approach.

The remaining changes should make editing and navigation consistently follow the book's reading order, without renaming assets or replacing the whole Sectioning editor.

### 1. Sectioning sidebar: Slides / PDF pages tabs

Sectioning currently navigates source PDF pages, each of which can contain multiple sections. Storyboard navigates individual sections/output slides. These are different units, so simply sorting the existing Sectioning page list cannot represent the storyboard order when sections from one page have been moved apart.

Add two tabs:

| Tab | Behaviour |
| --- | --- |
| **Slides** | One entry per section, following the shared book order. Default once sections exist. |
| **PDF pages** | Retain the existing page-based navigation in original PDF order, for reviewing a source page and its sections together. |

Keep the existing Sectioning editor for this iteration:

- Selecting a slide opens its source page and focuses the corresponding section.
- Selecting a PDF page opens the existing whole-page sectioning view.
- Moving a slide in either stage updates the shared order, reflected in Sectioning's Slides tab and Storyboard.
- Use source-PDF thumbnails until storyboard renderings are available.
- Preserve the selected section when switching views where possible.

The Slides tab is synchronized with Storyboard. The PDF pages tab intentionally retains source order. Both are views of the same section data; do not maintain a second order or physically relocate sections between source-page storage merely because they were reordered.

A full conversion of the Sectioning editor into a slide-only editor is outside this iteration. Do not lock Sectioning just because the book has been reordered.

### 2. Full Sectioning regeneration resets the storyboard and reading order

A full Sectioning rerun currently clears storyboard rendering but leaves the saved reading order. Regenerated section IDs can name different content, so the old arrangement can silently bind to the wrong sections. This was reproduced using the stage reset and section finalizer.

Use the following lifecycle:

| Operation | Expected behaviour |
| --- | --- |
| Reorder | Save reading order only; do not regenerate content or invalidate speech. |
| Edit, merge, split, or clone sections | Preserve/reconcile reading order using stable section identities. Existing broader regeneration behaviour can remain. |
| Rerun Storyboard rendering | Preserve section identities and reading order. |
| Regenerate Sectioning from extracted content | Reset reading order and rebuild the whole storyboard against the new section structure. |

Show an explicit warning before the full Sectioning rebuild, for example:

> **Rebuild sections and storyboard?** This replaces the current sections and resets storyboard layouts, manual storyboard edits, and page order. Dependent output will need regeneration.

Apply the reset at the backend rebuild boundary, not just in one UI entry point. Retained order history must not be independently restorable onto an incompatible regenerated structure. Associate order versions with their Sectioning generation, or provide an equivalent backend compatibility guard.

Targeted regeneration after individual section edits is deferred. This iteration does not need to make every downstream stage incremental.

### 3. One reading-order navigation path

The sidebar follows reading order, but `StoryboardView.tsx` still advances `sectionIndex` and then navigates source pages. This makes keyboard and toolbar navigation disagree with the displayed sequence.

Use the shared slide sequence for:

- Keyboard and toolbar previous/next actions, including navigation forwarded from the preview iframe.
- Sidebar selection and initial selection.
- Navigation into and out of quizzes.
- Any slide-position display.

Resolve selection by stable section/quiz ID and find its neighbours in the shared sequence. While rearranging, navigation should follow the same pending order shown in the sidebar.

The Studio route must identify the exact selected section, rather than relying solely on a source-page URL plus an in-memory section index. Refresh and browser Back should preserve the selected slide.

### 4. Minimal header: `7 / 42`

Use the existing coloured Storyboard header. Remove the `Page 8` label and the section `1 2 3` shortcut buttons. Show only the slide counter alongside the existing controls:

```text
Storyboard    7 / 42                       ← →
```

Do not introduce a `Book page` label or PDF provenance into this header. The counter follows the book's slide sequence, including quizzes, rather than source PDF pages.

Keep source provenance on thumbnail panels:

```text
PDF page 12    1/2
PDF page 12    2/2
```

The fractions describe sections within the source page; they are not reading-position numbers. They can therefore remain `2/2` and `1/2` when those slides are rearranged.

Use the corresponding slide counter when a slide is selected in Sectioning's Slides tab. Whole-PDF-page browsing must not imply that a PDF page containing several sections is one slide.

This is an authoring-header change. Do not insert an authoring header into the book's content; the exported reader retains its existing navigation controls.

### 5. Stable filenames with a dynamic position fragment

**Do not prefix or rename HTML files with reading-position numbers.** Keep the definitive URL and filename:

```text
pg002_sec001.html
```

For browser-based viewing, display the current position as a fragment:

```text
pg002_sec001.html#slide=1
```

If the section moves to position 7, the viewed address becomes:

```text
pg002_sec001.html#slide=7
```

The filename identifies the content. The fragment only reports its current reading position; it must never select or redirect to a different section based on that number.

Opening an old bookmark ending in `#slide=1` must still open the same section. Once its current position is resolved, update the fragment to `#slide=7` by replacing the current history entry, without adding an extra Back-button step.

Implementation boundaries:

- Derive the fragment from the same reading order as navigation.
- Keep source assets, output filenames, manifest `href`s, section IDs, HTML `data-id`s, and TTS keys unchanged.
- Preserve existing content anchors and glossary links. The runtime already uses `#glossary=...`; position display must not overwrite those destinations. Existing content-targeting fragments take priority.
- Keep export spines and document references unchanged. This is a browser-viewing enhancement, not a new filename or identity scheme.

This supersedes the earlier numbered-filename proposal. Keeping the pathname unchanged also avoids introducing a regression in activities that currently derive saved-answer keys from the filename; migrating those keys is not required for this fragment-only change.

### 6. Merge explicit sections, using the displayed order

With source order `A → B → C` and book order `A → C → B`, “Merge with next” on A currently merges A+B. The Book order overview sends the source section index and direction, and the backend selects a source-order neighbour.

Determine the target from the displayed sequence and send explicit source/destination section IDs. In that example, the operation must merge A+C. Apply the same rule across source-page boundaries.

Both editors should use the same structural operation. Preserve the destination's ID, retire the removed ID, reconcile reading order, and identify the actual destination in the confirmation.

### Acceptance checks

- Moving a slide updates Storyboard and Sectioning's Slides tab; PDF pages remains in source order.
- Sectioning's Slides tab works before storyboard HTML has been generated.
- Keyboard and toolbar navigation follow the displayed sequence across source pages and quizzes.
- The header counter follows slide position; source-page and within-page section labels stay on thumbnails.
- Refresh, browser Back, and stage/view changes retain the intended section selection.
- Merge acts on the displayed neighbour, including when it is not the adjacent source section/page.
- Full Sectioning regeneration resets reading order and rebuilds Storyboard; rendering-only regeneration preserves order.
- Incompatible old reading-order versions cannot be restored onto regenerated sections.
- A stale `#slide=N` bookmark opens the same content and updates its displayed position without extra history entries.
- Glossary/content-anchor navigation continues to work.
- Stable filenames, IDs, image references, TTS mappings, and existing activity-answer persistence remain intact.
- Preview and ADT, WebPub, EPUB, and PNLD exports agree on the reading sequence.

**Scope:** shared slide navigation, the two Sectioning sidebar views, a minimal counter, dynamic URL fragments, explicit merge targets, and full-rebuild handling. Defer per-section incremental regeneration and a complete Sectioning-editor redesign.

gbergengruen:
Thanks — this was a useful split to have drawn. Three of the six were defects in what this PR already ships, so they are fixed here; three are new surface, and putting them in would roughly double a PR that is already 57 files. Those are filed and linked below.

## Landed here

**2 — a full Sectioning rebuild resets the reading order.** You were right that this was reachable, and it is worse than it looks: the rebuild clears `page-sectioning`, which is where the section-id high-water mark is read from, so the re-run re-mints ids densely from `_sec001` for other content. A stored order matches by id string alone, so it reconciled *cleanly* onto whatever now held those ids — nothing downstream could complain, because no slot was missing.

There is now a `sectioning-generation` counter, bumped at the backend rebuild boundary (`makeBeforeRun`, not a UI entry point) and stamped onto every order saved after it. An order from an older generation is ignored by the resolver and refused by the restore route with a 409 — that is the "equivalent backend compatibility guard" you offered as an alternative to tying versions to a generation; it turned out to be the same thing. The rows are kept rather than deleted, so history stays visible and the picker strikes those versions through and says why. The warning before the rebuild is there too, and distinguishes the two cases: a Sectioning rebuild keeps the history, a re-extract deletes it outright, and promising the user their earlier orders survive when they do not seemed worse than saying nothing.

One thing I moved while doing it: `clearsSectionIdHistory` now lives in `@adt/types` as `rebuildsSectionIds`, so the UI warns about exactly the reruns the backend will reset. A rule defined on one side and re-guessed on the other is how the reorder-blocking rule came to disagree with itself earlier in this PR.

**6 — merge acts on the displayed neighbour.** Confirmed exactly as you describe. There is a new id-addressed `POST /books/:label/sections/merge` taking `{ keepSectionId, removeSectionId, direction }`; both index routes now resolve their neighbour and call the same two implementations, so the source-order and displayed-order paths cannot drift. Extracting them removed the adjacency assumption — the node concat, the splice, the rendering shift and the activity remap were all index-agnostic already — which is what lets `A → C → B` merge A+C, across pages or within one. The confirmation names the destination. The cross-page items are hidden in `Book order`, since from the reading order's point of view there is only one kind of neighbour.

That also surfaced a latent bug: the survivor shifts down when the removed section sat before it, unreachable while every merge was adjacent.

**3 — one reading-order navigation path.** The sidebar and navigation now read one draft-aware sequence, so mid-rearrangement the arrows follow what is on screen. Quizzes are slides like any other, so the arrows step into and out of them. The selected slide is in the URL as `?section=<sectionId>` — a search param, not a path segment, so existing links keep working and the stage rail can still carry a page between stages; a link with a section but no page resolves to the page that section belongs to. `skipNextResetRef` is gone: page and section now move in one navigation, which is all it existed to paper over.

Three regressions turned up on the way and are fixed with tests: a structural edit names its target by an index into a tree the view has not received yet; the keyboard handler's dependency list predated the sequence, so keys and toolbar disagreed once the order loaded; and stepping onto a quiz replaced a history entry, so Back skipped the slide it came from.

**One correction to the brief.** Item 3 asks that navigation forwarded from the preview iframe follow the sequence. There is no such channel today — the iframe's `postMessage` handler carries only activity-editor anchors (`link-select` / `link-hover` / `select*` / `text-changed`), and the only other listener is reader-settings capture. I have not invented one; if you want iframe-driven page turns, that is worth its own issue.

## Deferred

- **#868** — Sectioning's Slides / PDF pages sidebar tabs (item 1).
- **#869** — the minimal `7 / 42` header (item 4). Blocked on this PR, which adds the sequence it counts.
- **#870** — reading position as a `#slide=N` fragment (item 5). Your reasoning holds and the issue records two traps found while scoping it: the runtime's only fragment consumer is anchored to the *whole* hash and erases it after reading, and there is no `hashchange` listener anywhere in the runtime.

Agreed on the boundaries you drew: no second order, no physically relocating sections between source-page storage, no locking Sectioning because the book has been reordered, and no numbered filenames.

## State

`pnpm typecheck` clean · `pnpm test` 3788 passing (287 files) · `pnpm lint` 0 errors · all five locales at 0 missing.

Items 2 and 6 have not been through a browser pass yet — worth checking the rebuild warning and a cross-page merge from `Book order` before approving.

gbergengruen:
## Parking this, deliberately

Moving back to draft. Not because anything here is unfinished — every agreed item is done, CI is green, and the branch merges clean — but because of what it touches.

This changes how the book's sequence is decided, and then rewires packaging, preview, both TOC builders, quiz batching, the sign-language map, the storyboard and sectioning UIs, and now navigation. That is a lot of load-bearing surface for one change, and the priority right now is stabilising the core rather than widening it. Merging something this broad while that is the goal would work against it.

So this waits. Nothing is abandoned and nothing is blocked on a decision.

### State at the time of parking

| | |
|---|---|
| Head | `9d9a9c1b` |
| CI | `test`, `docker`, `i18n` all green |
| Merge | clean against `develop` |
| Local | `pnpm typecheck` clean · `pnpm test` 3788 passing (287 files) · `pnpm lint` 0 errors · 5 locales at 0 missing |

The [2026-09-16 review](https://github.com/unicef/adt-studio/pull/777#issuecomment-5698949571) is answered in full: items 2, 3 and 6 are implemented here, and 1, 4 and 5 are filed. Details in [the reply](https://github.com/unicef/adt-studio/pull/777#issuecomment-5715721190).

### What has not been done

**A browser pass over the three items landed since the review.** T1–T20 passed before that work; items 2, 3 and 6 have been verified by tests and typechecking only. Whoever picks this up should exercise, at minimum:

- the Sectioning-rebuild warning, and that old page orders are then struck through in the picker and refused on restore;
- a cross-page "Merge with next" from `Book order`, and that the dialog names the real destination;
- arrowing through a book with quizzes, including mid-rearrangement with an unsaved draft, and reload/Back preserving the open slide.

### Issues carrying the remaining work

Deferred from the review, all reachable from this branch's foundation:

- #868 — Sectioning sidebar: Slides / PDF pages tabs
- #869 — Storyboard header: `7 / 42` slide counter (blocked on this PR)
- #870 — reading position as a `#slide=N` fragment
- #873 — preview iframe cannot drive the authoring view (no navigation message exists)

Known limits this PR documented and did not work around:

- #871 — quiz placement anchored to a source page, not a section
- #872 — quizzes have no reversible removal
- #860 — quizzes anchored to a missing page are dropped from the output
- #861 — `maxNodeVersionExcluding` cannot detect a reorder
- #782 — split `StoryboardIndex.tsx` / `SectioningOverview.tsx`

Found while working here but independent of it:

- #874 — `eslint-plugin-react-hooks` is not enabled in `apps/studio`
- #875 — Docker builds download an Electron binary no image uses

### Before resuming

Rebase or merge `develop` first and re-run the suite — the longer this sits, the more `page-sectioning`, `quiz-generation` and the storyboard UI will have moved underneath it. The branch touches all three.

PR #780, opened 2026-08-11
speech: add selectable second narrator voice per language

### LINKED ISSUE #779: feat(speech): selectable second narrator voice per language in the Speech step
## Summary

Allow each language in the **Speech** step to have, in addition to the primary voice, an **optional, selectable second narrator voice**. The two voices are complete alternatives: they are generated, cached and exported independently, and the reader/export lets the end user choose which one to listen to.

## Motivation

Today TTS generates a single voice per language. We want to offer an alternative voice (e.g. a different gender/tone, or a different provider) without losing the primary one, and to carry that choice all the way through to the exported material.

## Functional scope

- **Voices (global):** one primary voice per provider/language. For ElevenLabs the **human-readable name** (e.g. "Tomás") is shown instead of the opaque ID.
- **Speech Settings (per book):** a **+** control per language adds a second row with **independent provider / model / voice**, plus a remove button. The second voice is stored per book and per **exact locale** (`speech.secondary_voices`).
  - The second voice may use a **different provider and model** than the primary.
  - Only the **Save & Re-run** flow is offered, and it is disabled until a valid voice is selected (incomplete profiles are never persisted).
- **Speech view:** when a second voice exists, a *Narrator voice* selector appears with the correct display name for both.
- **Reader/Export:** a narrator selector in the reader Settings; the end user switches between the primary and secondary voice.

## Technical design

- **Slot identity:** `primary` / `secondary`; composite identity = language + slot + textId.
- **File names:** primary `textId.ext`; secondary `textId--secondary.ext`.
- **Voice resolution:** `resolveSpeechVoice()` returns a full provider/model/voice/label profile per slot; primary uses global routing + `voices.yaml`; secondary uses `speech.secondary_voices` with **exact-locale matching** (an es-UY secondary never applies to es-AR). The ElevenLabs shipped-name fallback is symmetric for both slots.
- **Generation/caching:** primary always; secondary only when configured. Per-slot caching, retries, manual uploads and timestamps. Reusing audio (including manual uploads) refreshes the `voiceLabel` without regenerating.
- **Packaging/reader:** `audio_voices.json` and `timecode_voices.json` manifests. **HTML/PNLD** expose both voices; **EPUB** stays primary-only (secondary assets are stripped before building the OPF). PNLD relocation also handles `.wav`/`.ogg`.
- **Backward compatibility:** legacy mappings are still read (scalar and legacy global slot objects); global writes are **normalized to primary-only**.
- **i18n:** new strings translated across `en, es, pt-BR, fr, sq`.

## Out of scope

- A second voice in EPUB (by contract, EPUB stays primary-only).
- More than two voices per language.

## Testing

- Targeted suites green (schema, resolver, stage-runner, TTS routes, packaging web/PNLD/EPUB, floating-save): 400+ tests.
- `pnpm typecheck`, `pnpm lint` (studio) and `@adt/runtime typecheck` all green.
- Lingui extraction with no missing translations.

## Acceptance criteria

- [ ] Voices shows a single voice per provider/language (ElevenLabs with a human-readable name).
- [ ] The **+** in Speech Settings adds a second voice per book/locale; it can be removed.
- [ ] The second voice can use a different provider/model/voice than the primary.
- [ ] Save & Re-run is blocked until the second voice is complete.
- [ ] The Speech view and the exported reader let the user choose between both voices with the correct name.
- [ ] HTML/PNLD include both voices; EPUB only the primary.
- [ ] Existing projects (legacy mappings) keep working.

### LINKED ISSUE #824: fix(pnld): wav/ogg audio is not relocated to resources/audios, so PNLD export fails validation
## What happened?

PNLD export leaves `.wav` and `.ogg` audio inside the ADT sidecar instead of moving it to the edital's `resources/audios/` folder, so the exported package fails PNLD validation. The audio files are present in the `.zip`, just in a location the validator does not accept.

`relocateMedia()` in [`packages/pipeline/src/packaging/pnld.ts`](https://github.com/unicef/adt-studio/blob/develop/packages/pipeline/src/packaging/pnld.ts) keys on file extension, and `MEDIA_DEST` only listed:

```ts
".mp3":  { folder: "audios", map: "audios.json" },
".flac": { folder: "audios", map: "audios.json" },
".mp4":  { folder: "videos", map: "videos.json" },
```

Anything else is skipped by `if (!dest) continue`, so it stays under `resources/data/content/i18n/<lang>/audio/` and its entry in `audios.json` keeps the un-prefixed filename.

## Who is affected

Any book whose audio is `.wav` or `.ogg` at export time:

- **Gemini-narrated books — always.** `resolveSpeechFormat()` returns `wav` unconditionally for the `gemini` provider (`packages/pipeline/src/speech.ts`), so every Gemini book is affected.
- **Books with a `.wav`/`.ogg` manual upload**, even when the rest of the book is mp3 — one hand-recorded line is enough to strand a single file.

Books narrated by OpenAI, Azure or ElevenLabs produce `.mp3` and are unaffected — their audio has always been relocated correctly.

## Expected behavior

`.wav` and `.ogg` audio is moved to `resources/audios/<lang>__<file>` and `audios.json` is rewritten to match, exactly as `.mp3` already is, so the package passes the PNLD validator.

## Actual behavior

`.wav`/`.ogg` files remain in `resources/data/content/i18n/<lang>/audio/` with `audios.json` still pointing at the original names, and validation fails.

## Steps to reproduce

1. Create a book and route its Speech step to the **Gemini** provider (or upload a `.wav` manually for one entry).
2. Run the Speech stage so audio is generated as `.wav`.
3. Export the book to **PNLD (.zip)**.
4. Inspect the archive: the `.wav` files are under `resources/data/content/i18n/<lang>/audio/`, not `resources/audios/`.
5. Run the package through the PNLD validator — it rejects the audio location.

## Notes

Present since PNLD export was introduced (#555); not a regression from any recent change.

A fix is already implemented and covered by a regression test in PR #780, which adds `.wav`/`.ogg` to `MEDIA_DEST`. It is independent of that PR's dual-narrator feature and would be needed regardless — filing separately so it is tracked and gets its own release-note line, since it changes the exported layout of existing books.

gaguerre-iugo:
Closes #779
Closes #824

## What & why

Adds an optional **second, selectable narrator voice** per language to the Speech step. The two voices are complete alternatives: generated, cached and exported independently, and the end user can pick between them in the reader/export. Previously TTS produced a single voice per language.

## Behavior

- **Voices (global):** one primary voice per provider/language. ElevenLabs now shows the **human-readable name** (e.g. "Tomás") instead of the opaque voice ID.
- **Speech Settings (per book):** a **+** control per language adds a second row with **independent provider / model / voice**, plus a remove button. The secondary voice is stored per book and per **exact locale** under `speech.secondary_voices`.
  - The secondary voice may use a **different provider and model** than the primary.
  - Only the **Save & Re-run** action is offered, and it is disabled until a valid voice is chosen (incomplete profiles are never persisted).
- **Speech view:** when a second voice exists, a *Narrator voice* selector appears with the correct display name for both.
- **Reader/Export:** the reader Settings expose a narrator selector so the end user switches between primary and secondary.

## Technical notes

- **Slot identity:** `primary` / `secondary`; composite identity = language + slot + textId. Filenames: primary `textId.ext`, secondary `textId--secondary.ext`.
- **Resolution:** `resolveSpeechVoice()` returns a full provider/model/voice/label profile per slot. Primary uses global routing + `voices.yaml`; secondary uses `speech.secondary_voices` with **exact-locale matching** (an es-UY secondary never applies to es-AR). ElevenLabs shipped-name fallback is symmetric for both slots.
- **Generation/caching:** primary always; secondary only when configured. Per-slot caching, retries, manual uploads and timestamps. Reusing audio (including manual uploads) refreshes `voiceLabel` without regenerating.
- **Packaging/runtime:** adds `audio_voices.json` and `timecode_voices.json` manifests. **HTML/PNLD** expose both voices; **EPUB** stays primary-only (secondary assets/manifests are stripped before OPF enumeration). PNLD relocation also handles `.wav`/`.ogg`.
- **Backward compatibility:** legacy scalar and legacy global slot mappings remain readable; global writes are normalized to primary-only.
- **i18n:** new strings translated across `en, es, pt-BR, fr, sq`.

## Testing

- Targeted suites green (types schema, resolver, stage-runner, TTS routes, packaging web/PNLD/EPUB, floating-save): 400+ tests.
- `pnpm typecheck`, `pnpm lint` (studio), `@adt/runtime typecheck`, and Lingui extraction (no missing translations) all green.

## Reviewer notes

- Out of scope by design: a second voice in EPUB (primary-only export), and more than two voices per language.
- Please sanity-check the packaging split (HTML/PNLD both vs EPUB primary-only) and the exact-locale semantics for `speech.secondary_voices`.

---

## Review follow-up (2026-08-27)

`develop` merged in (two conflicts from #797's TTS log-builder refactor), plus one commit per review finding.

**Correctness**

- **Narrator labels resolved through one shared helper.** The packager and the Studio preview each derived the voice name their own way, so the same book showed "Tomás" in an export and the raw ElevenLabs ID in the preview. Both now use `resolveNarratorLabel`. This also fixes the label being reassigned per entry (one manual upload renamed the whole narrator to "uploaded") and stops the ElevenLabs picker persisting `describeVoice`'s decorated `"Tomás (premade, spanish)"`, which was reaching end users on the reader's narrator button.
- **Deterministic entry order.** Reused entries are collected in the scan pass and generated ones after it, so a mixed run could persist the secondary first — and Studio reads the first entry to describe a language. `mergeSpeechEntry`'s comparator is now shared as `sortSpeechEntries` and used by both writers. The Studio-side primary filter is *required*, not defensive: a run in flight is served from the live snapshot in completion order, which the persist-time sort never touches.
- **`narrator-voice-label` was in 6 of 69 runtime locales.** `useTranslation` returns the key for a missing entry, so the `|| "Narrator voice"` guard was unreachable — 63 locales would have rendered the raw id. `t` now takes an optional fallback (used across `SettingsTab`, whose ~20 other fallbacks were equally dead), and the key is filled into all 69 locales. **Oromo (`cus`) and Dzongkha (`dz`) are best-effort and want a native reviewer.**
- **Empty-slot manifest.** When every clip for a slot fails, the manifest still listed it — the reader offered a selectable option under an untranslated "Primary"/"Secondary" placeholder that played silence (`{}` is not nullish, so the fallback never fired). The picker now requires both slots to have audio.

**Deliberate, worth a reviewer's eye**

- **PNLD `.wav`/`.ogg` relocation is an independent bug fix — confirmed with the team, tracked as #824.** `relocateMedia` keys on file extension and previously handled only `.mp3`/`.flac`/`.mp4`, so wav/ogg audio stayed in `resources/data/content/i18n/<lang>/audio/` instead of moving to `resources/audios/` — where the PNLD validator requires it. That affects Gemini-narrated books (`resolveSpeechFormat` returns wav unconditionally for Gemini) and any book with a `.wav`/`.ogg` manual upload; mp3 books were already relocated and export byte-identically. So this fixes **every** wav/ogg PNLD export, not just dual-voice ones, and would be needed even without this feature. Pinned by a single-voice regression test. **Worth its own release-note line**, since it changes the output of existing books.
- **`voices.yaml` gains a forward-only shape.** ElevenLabs rows with a label persist as `{ primary: { voice, label } }`; an older build reading that file gets `[object Object]` as the voice ID. Accepted — the label is needed for cloned voices the packager can't resolve offline.
- **Secondary voices have no cross-provider fallback.** The old lookup was unreachable (the voices endpoint normalises `voices.yaml` to primary-only), so it now says so at the call site: retrying a secondary elsewhere would narrate in a voice the user didn't choose.

**Also**: dropped the write-only `timecodeMapAtom`; added a cost warning to the `+` control (a second voice doubles synthesis *and* Whisper alignment); restored the dropped page-batch diagnostic.

**Coverage**: 219 files / 2836 tests (was 218 / 2815). `adt-preview`'s `audio_voices.json` and `useTranslation` had no tests at all and now do. Typecheck, lint and Lingui extraction all clean.


gbergengruen (approved):
This looks good to me.

PR #784, opened 2026-08-11
fix(storyboard): make sectionIds stable across structural ops

gbergengruen:
First of five PRs splitting #777 (#660, reorder storyboard pages) into reviewable pieces. **This one stands alone and fixes live data corruption on `develop` today** — it is worth reading on its own merits, not as groundwork.

## The bug

Every structural operation on a page renumbered that page's `sectionId`s. So editing one section silently renamed its neighbours — and only `editable-activity` was migrated to match. Everything else keyed by `sectionId` was left pointing at the wrong section:

- `sign_language_videos.section_id` — a sign-language video reattached to unrelated content
- `toc-generation` entries
- text-catalog keys `${sectionId}_ans_*`, and so the translations and generated TTS audio derived from them

Cross-page merge renumbered both pages wholesale.

## The fix

`sectionId` becomes immutable. Clone, split, merge, cross-page merge and delete now retire ids rather than renumber survivors, via a new `createSectionIdFactory` / `retireSectionIds` pair in `apps/api/src/routes/pages.ts`.

**Ids are never reused.** The factory reads its high-water mark from *all* stored versions of a page's sectioning, not just the current one — hence the new `Storage.getAllNodeVersions`. Without that, `delete _sec003` → `split _sec002` reissues `_sec003` and adopts the deleted section's TOC entry, video and catalog keys.

This also rules out storing the counter on the entity itself: it would live inside the thing it counts, so a version restore would roll it back and reissue everything allocated since.

## Public surface added

- `formatSectionId` / `parseSectionId` / `MAX_SECTION_SEQ` in `packages/types/src/page-sectioning.ts`
- `Storage.getAllNodeVersions(node, itemId)` in `packages/storage`

No schema change, no migration.

## Not addressed here

Books already edited on `develop` carry dangling `sign_language_videos.section_id` and `toc-generation` sectionIds from historical renumbering. This stops the bleeding but repairs nothing. Surfacing those as a validation warning is follow-up work — they must never be auto-remapped, since a wrong guess silently attaches a sign-language video to unrelated content.

## Verification

`pnpm typecheck` clean · `pnpm test` 2745 passing · `pnpm lint` 0 errors.

Load-bearing tests: `sectionId stability across structural ops` in `apps/api/src/routes/pages.test.ts` (drives the real HTTP routes, one case per operation), no-reuse-after-delete, and sign-language survival across a merge.

---

**Review stack** (merge bottom-up): **#1 this** → stable quizIds → one reading-order resolver → id-based output filenames → #777 the feature.

### DIFF
--- a/apps/api/src/routes/adt-preview.test.ts
+++ b/apps/api/src/routes/adt-preview.test.ts
@@ -191,6 +191,48 @@ describe("ADT preview routes", () => {
     expect(res.status).toBe(404)
   })
 
+  it("resolves a stored legacy `_sN` section id", async () => {
+    // Books upgraded from a version whose agent tools minted `_sN` ids still
+    // store them, and the UI builds the preview URL from the stored id. The
+    // section is still matched exactly — only the owner-page parse is widened.
+    const storage = createBookStorage(label, tmpDir)
+    try {
+      storage.putNodeData("page-sectioning", `${label}_p1`, {
+        reasoning: "ok",
+        sections: [
+          {
+            sectionId: `${label}_p1_sec001`,
+            sectionType: "content",
+            backgroundColor: "#fff",
+            textColor: "#000",
+            pageNumber: 1,
+            isPruned: false,
+            nodes: [],
+          },
+          {
+            sectionId: `${label}_p1_s2`,
+            sectionType: "content",
+            backgroundColor: "#fff",
+            textColor: "#000",
+            pageNumber: 1,
+            isPruned: false,
+            nodes: [],
+          },
+        ],
+      })
+    } finally {
+      storage.close()
+    }
+
+    const app = createAdtPreviewRoutes(tmpDir, webAssetsDir, path.resolve(process.cwd(), "config.yaml"))
+    const res = await app.request(`/books/${label}/adt-preview/${label}_p1_s2.html`)
+
+    expect(res.status).toBe(200)
+    const html = await res.text()
+    expect(html).toContain("Second section body")
+    expect(html).not.toContain("First section body")
+  })
+
 
   it("applies shared section-role cleanup and image alt fallbacks in preview output", async () => {
     const app = createAdtPreviewRoutes(tmpDir, webAssetsDir, path.resolve(process.cwd(), "config.yaml"))
--- a/apps/api/src/routes/adt-preview.ts
+++ b/apps/api/src/routes/adt-preview.ts
@@ -4,7 +4,7 @@ import { createHash } from "node:crypto"
 import { pathToFileURL } from "node:url"
 import { Hono } from "hono"
 import { HTTPException } from "hono/http-exception"
-import { isHeadingRole, isTtsExcluded, parseBookLabel } from "@adt/types"
+import { isHeadingRole, isTtsExcluded, parseBookLabel, parseSectionId } from "@adt/types"
 import {
   WebRenderingOutput,
   type SpeechConfig,
@@ -304,10 +304,9 @@ function buildSectionIdToPageIndex(storage: Storage): Map<string, number> {
         const sections = [...parsed.data.sections].sort((a, b) => a.sectionIndex - b.sectionIndex)
         for (const rs of sections) {
           const sectionMeta = sectioning?.sections?.[rs.sectionIndex]
-          if (sectionMeta?.isPruned) continue
+          if (!sectionMeta || sectionMeta.isPruned) continue
           index++
-          const sectionId = sectionMeta?.sectionId ?? `${page.pageId}_sec${String(rs.sectionIndex + 1).padStart(3, "0")}`
-          map.set(sectionId, index)
+          map.set(sectionMeta.sectionId, index)
         }
       }
     }
@@ -348,9 +347,11 @@ function buildPagesManifest(storage: Storage): Array<{ section_id: string; href:
         const sections = [...parsed.data.sections].sort((a, b) => a.sectionIndex - b.sectionIndex)
         for (const rs of sections) {
           const sectionMeta = sectioning?.sections?.[rs.sectionIndex]
-          // Skip sections that are pruned in the sectioning data
-          if (sectionMeta?.isPruned) continue
-          const sectionId = sectionMeta?.sectionId ?? `${page.pageId}_sec${String(rs.sectionIndex + 1).padStart(3, "0")}`
+          // Skip sections that are pruned, and orphan rendering entries with no
+          // sectioning row — their real sectionId is unknowable and a positional
+          // guess could collide with another section's id.
+          if (!sectionMeta || sectionMeta.isPruned) continue
+          const sectionId = sectionMeta.sectionId
           const entry: { section_id: string; href: string; page_number?: number } = {
             section_id: sectionId,
             href: `${sectionId}.html`,
@@ -435,7 +436,7 @@ function buildHeadingBasedToc(storage: Storage): Array<{ section_id: string; hre
       const sectionMeta = sectioning?.sections?.[rs.sectionIndex]
       if (!sectionMeta || sectionMeta.isPruned) continue
 
-      const sectionId = sectionMeta.sectionId ?? `${page.pageId}_sec${String(rs.sectionIndex + 1).padStart(3, "0")}`
+      const sectionId = sectionMeta.sectionId
 
       const heading = findFirstHeadingLeaf(sectionMeta.nodes)
       if (heading) {
@@ -1090,12 +1091,11 @@ export function createAdtPreviewRoutes(
 
       // Content page — require sectionId format (e.g. pg001_sec001).
       // This prevents ambiguous fallback to unrelated sections.
-      const sectionIdMatch = pageId.match(/^(.+)_sec(\d{3})$/)
-      if (!sectionIdMatch) {
+      const parsedSectionId = parseSectionId(pageId)
+      if (!parsedSectionId) {
         throw new HTTPException(404, { message: `Section not found: ${pageId}` })
       }
-      const ownerPageId = sectionIdMatch[1]
-      const fallbackSectionIndex = parseInt(sectionIdMatch[2], 10) - 1
+      const ownerPageId = parsedSectionId.pageId
 
       const renderRow = storage.getLatestNodeData("web-rendering", ownerPageId)
       if (!renderRow) {
@@ -1107,12 +1107,14 @@ export function createAdtPreviewRoutes(
         throw new HTTPException(500, { message: "Invalid rendering data" })
       }
 
-      // Get sectioning to look up sectionId → sectionIndex mapping
+      // Get sectioning to look up sectionId → sectionIndex mapping. The id's
+      // sequence number is NOT its array position (ids are allocated once and
+      // never reused), so there is no positional fallback — an unresolvable id
+      // is a 404 rather than a guess that would serve a different section.
       const sectioning = getRenderSectioning(storage, ownerPageId)
-      const resolvedIndex = sectioning
+      const targetSectionIndex = sectioning
         ? sectioning.sections.findIndex((s) => s.sectionId === pageId)
         : -1
-      const targetSectionIndex = resolvedIndex >= 0 ? resolvedIndex : fallbackSectionIndex
       const renderedSection = parsed.data.sections.find((s) => s.sectionIndex === targetSectionIndex)
 
       if (!renderedSection || targetSectionIndex < 0) {
--- a/apps/api/src/routes/adt-preview.ts
+++ b/apps/api/src/routes/adt-preview.ts
@@ -4,7 +4,13 @@ import { createHash } from "node:crypto"
 import { pathToFileURL } from "node:url"
 import { Hono } from "hono"
 import { HTTPException } from "hono/http-exception"
-import { isHeadingRole, isTtsExcluded, parseBookLabel, parseSectionId, resolveEntryVoiceSlot } from "@adt/types"
+import {
+  isHeadingRole,
+  isTtsExcluded,
+  parseBookLabel,
+  parseAnySectionId,
+  resolveEntryVoiceSlot,
+} from "@adt/types"
 import {
   WebRenderingOutput,
   type SpeechConfig,
@@ -1158,9 +1164,12 @@ export function createAdtPreviewRoutes(
         return c.body(html)
       }
 
-      // Content page — require sectionId format (e.g. pg001_sec001).
-      // This prevents ambiguous fallback to unrelated sections.
-      const parsedSectionId = parseSectionId(pageId)
+      // Content page — the id must have a section shape (e.g. pg001_sec001) so
+      // this cannot fall back to an unrelated section. The legacy `_sN` shape is
+      // accepted too: books upgraded from a version whose agent tools minted
+      // those still store them, and the section is matched exactly below either
+      // way, so widening the parse resolves them without guessing anything.
+      const parsedSectionId = parseAnySectionId(pageId)
       if (!parsedSectionId) {
         throw new HTTPException(404, { message: `Section not found: ${pageId}` })
       }
--- a/apps/api/src/routes/adt-preview.ts
+++ b/apps/api/src/routes/adt-preview.ts
@@ -4,7 +4,7 @@ import { createHash } from "node:crypto"
 import { pathToFileURL } from "node:url"
 import { Hono } from "hono"
 import { HTTPException } from "hono/http-exception"
-import { isHeadingRole, isTtsExcluded, parseBookLabel, resolveEntryVoiceSlot } from "@adt/types"
+import { isHeadingRole, isTtsExcluded, parseBookLabel, parseSectionId, resolveEntryVoiceSlot } from "@adt/types"
 import {
   WebRenderingOutput,
   type SpeechConfig,
@@ -310,10 +310,9 @@ function buildSectionIdToPageIndex(storage: Storage): Map<string, number> {
         const sections = [...parsed.data.sections].sort((a, b) => a.sectionIndex - b.sectionIndex)
         for (const rs of sections) {
           const sectionMeta = sectioning?.sections?.[rs.sectionIndex]
-          if (sectionMeta?.isPruned) continue
+          if (!sectionMeta || sectionMeta.isPruned) continue
           index++
-          const sectionId = sectionMeta?.sectionId ?? `${page.pageId}_sec${String(rs.sectionIndex + 1).padStart(3, "0")}`
-          map.set(sectionId, index)
+          map.set(sectionMeta.sectionId, index)
         }
       }
     }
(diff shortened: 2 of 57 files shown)


ksokolovic:
I found three issues that should be addressed before merging:

### [P1] Generated activities become unpreviewable and can receive duplicate IDs

The preview route now resolves only canonical `pageId_secNNN` IDs through `parseSectionId()`, but the agent activity tools still create IDs in the legacy `pageId_sN` form. For example, both `createTemplatedActivity` and `createCustomSection` derive the ID from the current section count.

That has two consequences:

- Newly generated activities return 404 in classic/stepper preview because `parseSectionId("pg001_s2")` returns `null`.
- After a deletion leaves a stable-ID gap, length-based allocation can mint an ID that already exists. I reproduced `[pg001_sec002, pg001_s2]` becoming `[pg001_sec002, pg001_s2, pg001_s2]` after creating another custom activity. Duplicate IDs make manifest references ambiguous and can cause package HTML writes to overwrite one another.

Relevant code: [preview resolution](https://github.com/unicef/adt-studio/blob/501b9c2425d110022f73dc4c769bfdd887a026eb/apps/api/src/routes/adt-preview.ts#L1167) and [activity ID allocation](https://github.com/unicef/adt-studio/blob/501b9c2425d110022f73dc4c769bfdd887a026eb/packages/agents/src/tools/book-tools.ts#L454).

I recommend using one shared allocator across structural API routes, pipeline code, and agent tools, and retaining exact legacy `_sN` lookup or migration without restoring positional guessing.

### [P1] Sectioning and Extract reruns still reuse retired IDs

Stage reruns clear the versioned node history, while page sectioning recreates IDs densely with `formatSectionId(pageId, sIdx + 1)`. External sign-language assignments survive those clears. This means a previously retired ID can be assigned to unrelated regenerated content.

I reproduced this with old sections `[pg001_sec001, pg001_sec003]` and a video assigned to `pg001_sec003`. After clearing page-sectioning history and finalizing a new three-section result, the regenerated IDs were `[pg001_sec001, pg001_sec002, pg001_sec003]`; the surviving video assignment now resolved to the unrelated new third section. This violates the PR's never-reuse contract and preserves the wrong-content attachment failure the change is intended to prevent.

Relevant allocation: [page-sectioning.ts](https://github.com/unicef/adt-studio/blob/501b9c2425d110022f73dc4c769bfdd887a026eb/packages/pipeline/src/page-sectioning.ts#L652).

I recommend persisting a high-water or retirement registry outside the history cleared by reruns, allocating fresh IDs for regenerated entities, and retiring/unassigning external references during both Sectioning and full Extract reruns.

### [P2] The orphan-page warning is invisible in normal packaging/export flows

Packaging emits the new orphan warning only through the optional `Progress` callback. `packageAdtWeb` defaults that callback to `nullProgress`, and the Studio packaging and export paths call it without a progress implementation. Even in progress-aware execution, the warning is immediately replaced by later status messages such as `Writing manifests...`.

As a result, the primary user-facing flows can complete successfully while silently omitting orphaned pages.

Relevant code: [warning emission](https://github.com/unicef/adt-studio/blob/501b9c2425d110022f73dc4c769bfdd887a026eb/packages/pipeline/src/packaging/web.ts#L551) and [Studio packaging call](https://github.com/unicef/adt-studio/blob/501b9c2425d110022f73dc4c769bfdd887a026eb/apps/api/src/routes/package.ts#L260).

I recommend returning structured warnings from packaging and exposing them in task/export results, or failing closed when pages would be omitted.

Verification: I reviewed the exact PR head and current generated merge result. `pnpm build`, `pnpm typecheck`, `pnpm lint`, the focused affected suites, and the full test suite all passed (3,364 tests). I also reproduced both P1 behaviors with targeted diagnostic tests. The green suite therefore appears to reflect missing regression coverage for these paths rather than test or environment noise.

gbergengruen:
Thanks — verified all three against the PR head. Two are regressions this PR caused and are fixed in 304f71c; two are pre-existing and are follow-ups. Detail on each, including one correction:

### [P1] Generated activities — split verdict

**Duplicate ids: confirmed, and newly reachable here. Fixed.** This depended on `renumberSectionIds`, which this PR removed — it had been rewriting every id on a page after each structural op, so it was incidentally *healing* legacy `_sN` ids. Your repro is exact: `[sec001]` → create → delete section 0 → create mints the same id again, because `nextIndex = sections.length` is a position, not a high-water mark. Blast radius is worse than ambiguous manifest refs — packaging writes `${sectionId}.html`, so one section silently overwrites the other in the bundle.

Fixed as you recommended, with one shared allocator: `createSectionIdFactory` / `collectSpentSectionIds` move from `pages.ts` into `@adt/pipeline`, and the agent tools now allocate through it. The scan pattern also admits legacy `_sN` so those burn a sequence number instead of being invisible to the high-water mark. `updateSection` and `mirrorLayout`'s `?? ${pageId}_s${sectionIndex}` fallbacks now throw rather than guess.

**The 404, though, is pre-existing — not caused by this PR.** `adt-preview` has hard-required `_sec\d{3}` since 35e2fc5 (#113), and `parseSectionId`'s `_sec\d+` is strictly *looser*. `pg001_s2` 404s identically on `develop` and at this head; what this PR removed is only the positional fallback for ids that already matched `_secNNN`. Once the tools mint canonical ids the symptom goes away for new content, and there is no legacy `_sN` in real data — zero non-`_sec` sectionIds across all 11 book DBs in `books/`. I skipped the exact-match lookup for now (a 3-line follow-up) since nothing needs it; happy to add it if you'd rather belt-and-braces it.

### [P1] Rerun reuse — confirmed, follow-up

Reachable as described. One refinement worth recording: after a rerun there are no retired ids left *inside* the sectioning node, since `clearNodesByType` / `clearExtractedData` delete every version. `toc-generation` and `text-catalog` also live in `node_data` and are wiped by the same clear, so they self-heal. The single surviving reference is `sign_language_videos.section_id` — a separate table nothing in `makeBeforeRun` reconciles. So the collision is that row against a freshly re-minted dense id space.

It's also pre-existing in kind: dense re-minting already reattached a pinned video to whatever landed at that position on `develop`. What this PR changes is that it now *states* the never-reuse invariant this path violates, and makes the sparse-gap variant possible for the first time.

Deliberately not fixed here to keep this PR reviewable — it's the base of a 5-PR stack. Agreed approach for the follow-up is to unassign videos for the pages being re-sectioned, reusing the "unassign, never delete" decision already in `retireSectionIds`. I've documented the hole at the `stages.ts` asymmetry, since the page-reconcile route already handles the identical hazard around `deletePage`.

### [P2] Orphan warning — confirmed, and understated. Fixed.

You're right, and it's worse than `nullProgress`: the CLI *does* pass a real `Progress`, but `createCliProgress` drops any `step-progress` without `page`/`totalPages`, which the orphan warning has none of. So no live sink ever received it — the "overwritten by `Writing manifests...`" path would only bite a last-write-wins sink.

Also broader than per-section drift: `getRenderSectioning` returns `undefined` both when no row exists *and* when the row fails `safeParse`, so an unparseable sectioning row drops the **whole page** from `pages.json`, where `develop` would have packaged it.

Fixed via structured warnings, as you suggested: `packageAdtWeb` returns `PackagingWarning[]` (data rather than a formatted string, since the wording needs Studio's Lingui catalogs), riding the existing task-result channel to a toast in the preview and export flows. One thing your repro didn't reach that the tests now cover — the warnings are cached next to the build hash, because a packaging cache hit short-circuits and would otherwise report a clean run while the bundle on disk is still short. There were two such short-circuits; a test caught the second.

### On the green suite

Your read was right — it was missing coverage, not noise. New tests: the duplicate-id sequence after a delete, canonical id shape, legacy `_sN` counted from history, returned packaging warnings (per-section, whole-page, and the clean case), and warnings surviving a cache hit. `pnpm typecheck` clean, `pnpm lint` 0 errors, `pnpm test` 3371 passing, all five locales translated.

ksokolovic:
Thanks for the detailed response. I rechecked the exact updated head (`304f71c`) against the original findings and your explanation. The shared allocator fixes the duplicate-id sequence and makes newly generated activities canonical, and structured warnings reach newly executed asynchronous Preview/export tasks. I still see three residual issues:

### [P1] Reruns still silently reattach sign-language videos to unrelated content

This remains reachable exactly as described. The new comment in `stages.ts` accurately documents the gap, but both rerun branches still delete all `page-sectioning` history, `finalizePageSectioning` still densely remints from `_sec001`, and `sign_language_videos.section_id` still survives. A video can therefore remain pinned to an ID now naming unrelated regenerated content.

Relevant code: [stage clearing](https://github.com/unicef/adt-studio/blob/304f71cca24640cd5ae117346e9ec0ce48ad3089/apps/api/src/routes/stages.ts#L28) and [dense reminting](https://github.com/unicef/adt-studio/blob/304f71cca24640cd5ae117346e9ec0ce48ad3089/packages/pipeline/src/page-sectioning.ts#L652).

I agree with your refinement that TOC/text-catalog rows are cleared and the surviving reference is the video table. That narrows the blast radius but does not reduce the silent wrong-content attachment. Given this PR's stated “IDs are never reused” invariant and its purpose of stopping that corruption, I still consider this a blocker rather than a follow-up.

### [P2] Existing `_sN` activities lose the preview path that worked on `develop`

I agree that a direct `_sN` URL already returned 404 on `develop`. The user-visible behavior was nevertheless different: the `develop` UI generated a canonical positional `_secNNN` URL, and the server fallback resolved that URL to the activity's array index. This PR changes the UI to request the stored `_sN` ID while also removing that fallback, so an existing agent-created `_sN` activity now has no working preview URL.

I reproduced the upgrade state with a stored `_s1` activity: both the stored URL and the former canonical URL returned `[404, 404]` at this head. The scan of 11 local DBs lowers the observed likelihood, but shipped code could create these IDs in other desktop/shareable books. An exact legacy-ID lookup or migration would preserve compatibility without restoring positional guessing.

Relevant code: [stored-ID preview URL](https://github.com/unicef/adt-studio/blob/304f71cca24640cd5ae117346e9ec0ce48ad3089/apps/studio/src/components/pipeline/stages/storyboard/components/SectioningOverview.tsx#L892) and [canonical-only preview parsing](https://github.com/unicef/adt-studio/blob/304f71cca24640cd5ae117346e9ec0ce48ad3089/apps/api/src/routes/adt-preview.ts#L1167).

### [P2] Packaging omissions can still be reported as clean

The new cache test covers a cache created by the new code, where `.build-warnings` already exists. It does not cover an existing valid package cache created before that sidecar was introduced. `readBuildWarnings()` returns `[]` when the file is absent, and `PACKAGING_FORMAT_VERSION` remains `5`, so the upgrade can accept the old build hash and return a clean result without rebuilding.

I reproduced this by removing the warning sidecar from a valid cached build: the cache hit returned `[]` instead of the known orphan warning.

There are also UI consumption gaps after the API returns warnings:

- Preview shows warnings for task completion, but its synchronous cache-hit branch ignores `result.warnings`.
- `api.packageAdt`'s response type does not include warnings.
- Validation ignores warnings in both synchronous and task-completion paths.
- The synchronous export fallback likewise downloads without consuming returned warnings.

Relevant code: [missing-sidecar behavior and cache fast path](https://github.com/unicef/adt-studio/blob/304f71cca24640cd5ae117346e9ec0ce48ad3089/apps/api/src/routes/package.ts#L63), [unchanged packaging format version](https://github.com/unicef/adt-studio/blob/304f71cca24640cd5ae117346e9ec0ce48ad3089/packages/pipeline/src/packaging/web.ts#L182), and [Preview cache-hit branch](https://github.com/unicef/adt-studio/blob/304f71cca24640cd5ae117346e9ec0ce48ad3089/apps/studio/src/components/pipeline/stages/PreviewView.tsx#L203).

I recommend invalidating pre-sidecar caches (format-version bump or requiring a valid warning sidecar) and surfacing warnings in every synchronous/asynchronous Preview, Validation, and export completion path.

Verification at `304f71c`: build, typecheck, and lint passed; 247 focused tests passed. The full run completed all 261 files / 3,371 tests successfully but exited nonzero because five `window is not defined` errors appeared after the runtime language-hook tests; that file passed 5/5 in isolation, matching a load-sensitive suite issue rather than these changes. The two additional regression diagnostics above reproduced their failures and were removed afterward.

gbergengruen:
Thanks — all three fixed, in `bf07bb4d` (plus `79998358`, a UI follow-on explained at the end).

### [P1] Rerun reuse — fixed here, agreed it belongs in this PR

`makeBeforeRun` now unassigns pinned sign-language videos before the clear, reusing the "unassign, never delete" decision already in `retireSectionIds` — which moves from `pages.ts` into `@adt/pipeline` beside the allocator, for the reason the allocator moved there: retirement is the other half of allocation, and the structural edit routes and the rerun path have to agree on it.

Worth being precise about why this is a fix and not just a narrowing. After the unassign, *no* reference to a retired id survives the clear, so the dense re-mint stops being observable reuse:

- `toc-generation`, `text-catalog` and `editable-activity` are deleted by the same clear, as you agreed.
- `sign_language_videos.section_id` is now reconciled. I checked the schema: it is genuinely the only sectionId reference outside `node_data` — no other table holds one.
- Audio was the case I wanted to rule out, since filenames embed `${sectionId}_ans_*` and live on disk outside the DB. It is safe: packaging copies files named by *node entries* (`web.ts:755`), that node is cleared by the rerun, and regeneration is keyed on a content hash of the text (`speech.ts:990-1010`) — so a filename that repeats is rewritten from the new text rather than served stale.

Ordering is load-bearing twice over: the retirement reads both the sectioning history the ids come from and the `pages` table, and the extract branch deletes both.

One trap worth recording, because the obvious implementation is wrong and I shipped it briefly before a test caught it. Scoping this to "any cleared sectioning node" over-retires: a storyboard rerun clears `fixed-layout-sectioning`, but `sectionFixedLayoutPage` derives that page's single id from the pageId rather than allocating it, so it regenerates identically and nothing pinned to it was ever at risk. Retiring those would detach every video on each storyboard rerun of a fixed-layout book — and on a reflowable book carrying a stale fixed-layout row it would retire a `_sec001` that the live `page-sectioning` still owns. Scoped to `page-sectioning`, with tests both ways.

**Not fixed, and I would rather flag it than let it pass:** the unassignment only reaches the server log. Telling the user *which* videos to reattach needs a notice channel the stage-run SSE does not have. Happy to add it here if you would rather not merge without it.

### [P2] Legacy `_sN` previews — you are right, and my earlier argument was wrong

Conceded on both counts. It is a regression this PR causes, not merely a pre-existing 404: `develop`'s UI sent a positional `_secNNN` URL that the server fallback resolved by index, and this branch changed the UI to send the stored id *and* removed the fallback. And "no such ids in the local books" was an argument about local data, not about upgrade safety — shipped code minted them, which is the only thing that matters.

`parseAnySectionId` accepts both shapes and is used *only* to derive the owner pageId; the section itself was already matched by exact string compare, so nothing positional comes back. `spentSeq` drops its own legacy regex and uses it too, so id shapes live in one place.

Verified against a book seeded with a stored `_s1`: the legacy id resolves `200`, the canonical id it replaced returns `404` (no positional fallback), and unknown ids still `404`.

### [P2] Packaging omissions — confirmed, including the pre-sidecar case

You are right that the cache test only covered a cache the new code created. Cache validity now requires the sidecar: an entry that cannot say what its build left out is not a usable entry. I went with that rather than bumping `PACKAGING_FORMAT_VERSION`, since the version is a pipeline-wide hash input shared with the other export formats and both approaches invalidate exactly the same entries.

Verified end to end against a real book with an orphaned rendering: first package writes the sidecar and reports the omission; the second returns it from cache; deleting the sidecar makes the third **rebuild** instead of returning clean.

All four consumption gaps closed through one `toastPackagingWarnings` helper — Preview's cache-hit branch, Validation (which had neither branch wired), and the export watcher's synchronous fallback — with both client response types now declaring `warnings` so the type system pushes on any new call site. `usePackageAdt` has no callers; I left it alone rather than delete it in this PR.

### On the full-suite errors

Your read that it was load-sensitive rather than environmental looks right. I reproduced the five `window is not defined` errors once, then got two clean full runs, and a run at `304f71c` was also clean here — an intermittent teardown race in `apps/adt-runtime/.../useTranslation.test.tsx`, unrelated to these changes.

### One thing you did not ask for

While testing the rerun path by hand I hit a gap this PR creates. The sign-language assignment UI names sections positionally (`Page 3 — Section 2`), which no longer tracks the id now that ids are sparse — and it never displayed the sectionId anywhere. So two sections on a page were distinguishable only by an ordinal that names neither, and a video whose section was retired left no on-screen record of what it had been attached to, which is exactly the state the rerun fix now produces on purpose. `79998358` shows the id beside the label in the options, the section rows and the "next missing" slot. That also fixed an accessible-name bug the id span introduced. Easy to drop if you would rather keep this PR to the three findings.

`pnpm typecheck` clean · `pnpm lint` 0 errors · `pnpm test` 3390 passing.

ksokolovic:
Thanks for the thorough follow-through — this is very close now. I checked `79998358` against the previous findings and your latest explanation. The legacy preview fix, pre-sidecar cache invalidation, warning delivery, and sign-language video retirement all check out. The fixed-layout distinction also makes sense.

Based on this re-review, **manual-audio retirement is the last correctness item I still have open**. I think this is a narrow final fix, rather than another round of broad changes.

### [P1] Preserved manual recordings can still follow a retired section ID

Your audio analysis is correct for content-hashed generated audio, but manual uploads take a different path:

- [`getStageRerunClearNodes`](https://github.com/unicef/adt-studio/blob/79998358ac2a1caaf2736f95082d97dc891cdfa4/packages/types/src/pipeline-effects.ts#L241) preserves the `tts` manifest when the rerun includes Speech, including Sectioning → Speech.
- [`canReuseSpeechEntry`](https://github.com/unicef/adt-studio/blob/79998358ac2a1caaf2736f95082d97dc891cdfa4/apps/api/src/services/stage-runner.ts#L679) accepts a manual entry when its file exists, without checking the new text.

I reproduced this with a targeted test using the route's retirement/clear operations, a regenerated activity answer, the real catalog builder, and the translation/speech runner:

1. Seed a manual recording for `pg001_sec001_ans_a` with the old answer.
2. Retire and clear Sectioning → Speech state; section history disappears, but the TTS manifest survives.
3. Give a different activity answer the re-minted section ID and rebuild its catalog.
4. Run translation/speech: the catalog contains the new answer, but the runner reports **0 generated / 1 reused** and retains the old manual recording.

This is an existing reuse path left exposed by the new cleanup, not a newly introduced audio regression. It does mean the claim that no references survive retirement is not quite true yet.

Please invalidate the manual-audio associations belonging to retired section-answer IDs before those IDs can be reused, while preserving the uploaded files. A regression test should cover changed answer content under a re-minted ID, alongside a control showing ordinary reruns still retain valid manual recordings.

The user-facing video-unassignment notice you flagged remains a worthwhile UX follow-up; I would keep that separate from this remaining correctness item.

Verification: build, typecheck, and 141 focused tests passed; lint had no errors. The additional targeted audio test fails as described. I did not rerun the full suite or a live LLM pipeline. Once this path is covered, I expect the next pass to be a focused closure check. Thanks again — the outstanding scope is substantially smaller now.

PR #785, opened 2026-08-11
fix(quizzes): give quizzes stable ids instead of deriving them from position

gbergengruen:
Adding, deleting, or replacing a quiz previously changed the positional IDs used by later quizzes, which could attach another question's translations or audio. This PR gives quizzes persistent book-local identities across Studio, catalogs, preview, and exported output.

## Identity and recovery

- Existing quizzes retain their IDs through edits and repositioning. Legacy books resolve IDs from each original stored array without writing a backfill version.
- Every application write allocates and saves atomically against all retained quiz history, including versions newer than a rollback. Generate-one replacement and full API/CLI regeneration allocate fresh identities, even for cached model output.
- IDs use at least three digits: `qz001` through `qz999`, then `qz1000` onward, bounded by JavaScript's maximum safe integer. Existing IDs and filenames remain unchanged. Deletion, regeneration, and rollback never release retained identities.
- Generation preflights capacity before model calls and rechecks during transactional persistence. Errors report requested and remaining counts without a partial save.
- Full quiz reruns retain prior output until success. Upstream invalidation and extraction retain history behind an inactive version. The API exposes the selected `historyVersion` separately from active output, so Studio keeps restore available and labels inactive versions as **Invalidated** in every supported locale.

## Validation and compatibility

Canonical format and uniqueness checks reject unsafe paths, duplicate IDs, alternate padding, whitespace, and unsafe integers. Catalog generation, preview, and packaging validate stored/imported data too. Packaging validates before replacing an existing export and checks destination containment.

Legacy ID-less PUT retries, deletions, reorders, and appends preserve identity when unambiguous. Ambiguous content edits return HTTP 400 with instructions to reload and retain IDs. Writes during quiz generation return HTTP 409, with a second check after generate-one's model call.

Historical comparison resolves IDs within each original version, avoiding false additions/removals after a legacy quiz's first save. Raw debug history remains unchanged. Studio consumes API-resolved identities and parses Storyboard URLs locally; legacy numeric links remain compatible.

The guarantee covers retained history and subsequent application writes. Previously erased history or externally replaced archives cannot be reconstructed. Quiz restore keeps the existing downstream invalidation contract; it is not a full pipeline snapshot restore. See [the identity contract](docs/QUIZ_IDENTITY.md).

## Verification

- Final commit `57338f59`: [GitHub CI passed](https://github.com/unicef/adt-studio/actions/runs/34583208928), including **3,548 tests across 270 files**, the Docker API build, and i18n/lint. The local full regression suite also passed.
- Build/typecheck, Studio production build, runtime typecheck, locale extraction, and strict compilation for all five locales passed.
- Lint: zero errors and the same eight existing warnings.

Regression coverage exercises real routes, SQLite, generation/reset/rollback, catalogs, preview, web/WebPub/EPUB exports, and Speech file writing. Boundary cases include `qz999` to `qz1000` to `qz1001`, unchanged audio references and recording bytes, failed/empty generation, concurrent allocation, corrupt imports, and export containment. Studio component tests exercise the real quiz view and version picker through successful restore, failed restore, and selecting an invalidation version.

Remote model responses and synthesized audio bytes are stubbed. Capacity-failure tests inject the preflight error, with safe-integer capacity arithmetic tested directly. No live provider calls or signed desktop packaging were performed locally.

This is the stable-quiz-ID prerequisite for #660. Reading-order editing remains in the later stack. The stable-section-ID prerequisite #784 is already merged into `develop`.

### DIFF
--- a/apps/api/src/routes/adt-preview.ts
+++ b/apps/api/src/routes/adt-preview.ts
@@ -4,7 +4,14 @@ import { createHash } from "node:crypto"
 import { pathToFileURL } from "node:url"
 import { Hono } from "hono"
 import { HTTPException } from "hono/http-exception"
-import { isHeadingRole, isTtsExcluded, parseBookLabel, parseSectionId } from "@adt/types"
+import {
+  isHeadingRole,
+  isTtsExcluded,
+  parseBookLabel,
+  parseSectionId,
+  parseQuizId,
+  resolveQuizId,
+} from "@adt/types"
 import {
   WebRenderingOutput,
   type SpeechConfig,
@@ -34,7 +41,6 @@ import {
   buildQuizAnswers,
   buildTextCatalog,
   flattenEasyReadEntries,
-  pad3,
   loadBookConfig,
   buildPreferredImageAltMap,
   buildDecorativeImageIdSet,
@@ -368,8 +374,7 @@ function buildPagesManifest(storage: Storage): Array<{ section_id: string; href:
     const quizzes = quizzesByAfterPageId.get(page.pageId)
     if (quizzes) {
       for (const quiz of quizzes) {
-        const quizIndex = quizData!.quizzes.indexOf(quiz)
-        const quizId = `qz${pad3(quizIndex + 1)}`
+        const quizId = resolveQuizId(quiz, quizData!.quizzes.indexOf(quiz))
         list.push({ section_id: quizId, href: `${quizId}.html` })
       }
     }
@@ -751,7 +756,7 @@ export function createAdtPreviewRoutes(
         const catalog = await getTextCatalog(storage)
         if (quizData?.quizzes) {
           for (let i = 0; i < quizData.quizzes.length; i++) {
-            const quizId = `qz${pad3(i + 1)}`
+            const quizId = resolveQuizId(quizData.quizzes[i], i)
             allHtml += renderQuizHtml(quizData.quizzes[i], quizId, catalog) + "\n"
           }
         }
@@ -1048,15 +1053,17 @@ export function createAdtPreviewRoutes(
         reflowableFont: config.reflowable_font,
       })
 
-      // Check if this is a quiz page (qzNNN)
-      const quizMatch = pageId.match(/^qz(\d{3})$/)
-      if (quizMatch) {
-        const quizIndex = parseInt(quizMatch[1], 10) - 1
+      // Check if this is a quiz page (qzNNN). The id's number is NOT the quiz's
+      // array position — ids are allocated once and never reused — so resolve by
+      // id rather than indexing into the array.
+      if (parseQuizId(pageId) !== null) {
         const quizData = getQuizData(storage)
-        if (!quizData?.quizzes || quizIndex < 0 || quizIndex >= quizData.quizzes.length) {
+        const quiz = quizData?.quizzes?.find(
+          (q, i) => resolveQuizId(q, i) === pageId
+        )
+        if (!quiz) {
           throw new HTTPException(404, { message: `No quiz data for: ${pageId}` })
         }
-        const quiz = quizData.quizzes[quizIndex]
         const catalog = await getTextCatalog(storage)
 
         const quizPalette = (config.quiz_generation?.match_book_style ?? true)
--- a/apps/api/src/routes/adt-preview.ts
+++ b/apps/api/src/routes/adt-preview.ts
@@ -10,6 +10,8 @@ import {
   parseBookLabel,
   parseAnySectionId,
   resolveEntryVoiceSlot,
+  parseQuizId,
+  resolveQuizId,
 } from "@adt/types"
 import {
   WebRenderingOutput,
@@ -42,7 +44,6 @@ import {
   buildQuizAnswers,
   buildTextCatalog,
   flattenEasyReadEntries,
-  pad3,
   loadBookConfig,
   buildPreferredImageAltMap,
   buildDecorativeImageIdSet,
@@ -380,8 +381,7 @@ function buildPagesManifest(storage: Storage): Array<{ section_id: string; href:
     const quizzes = quizzesByAfterPageId.get(page.pageId)
     if (quizzes) {
       for (const quiz of quizzes) {
-        const quizIndex = quizData!.quizzes.indexOf(quiz)
-        const quizId = `qz${pad3(quizIndex + 1)}`
+        const quizId = resolveQuizId(quiz, quizData!.quizzes.indexOf(quiz))
         list.push({ section_id: quizId, href: `${quizId}.html` })
       }
     }
@@ -763,7 +763,7 @@ export function createAdtPreviewRoutes(
         const catalog = await getTextCatalog(storage)
         if (quizData?.quizzes) {
           for (let i = 0; i < quizData.quizzes.length; i++) {
-            const quizId = `qz${pad3(i + 1)}`
+            const quizId = resolveQuizId(quizData.quizzes[i], i)
             allHtml += renderQuizHtml(quizData.quizzes[i], quizId, catalog) + "\n"
           }
         }
@@ -1123,15 +1123,17 @@ export function createAdtPreviewRoutes(
         reflowableFont: config.reflowable_font,
       })
 
-      // Check if this is a quiz page (qzNNN)
-      const quizMatch = pageId.match(/^qz(\d{3})$/)
-      if (quizMatch) {
-        const quizIndex = parseInt(quizMatch[1], 10) - 1
+      // Check if this is a quiz page (qzNNN). The id's number is NOT the quiz's
+      // array position — ids are allocated once and never reused — so resolve by
+      // id rather than indexing into the array.
+      if (parseQuizId(pageId) !== null) {
         const quizData = getQuizData(storage)
-        if (!quizData?.quizzes || quizIndex < 0 || quizIndex >= quizData.quizzes.length) {
+        const quiz = quizData?.quizzes?.find(
+          (q, i) => resolveQuizId(q, i) === pageId
+        )
+        if (!quiz) {
           throw new HTTPException(404, { message: `No quiz data for: ${pageId}` })
         }
-        const quiz = quizData.quizzes[quizIndex]
         const catalog = await getTextCatalog(storage)
 
         const quizPalette = (config.quiz_generation?.match_book_style ?? true)
--- a/apps/api/src/routes/adt-preview.ts
+++ b/apps/api/src/routes/adt-preview.ts
@@ -12,6 +12,7 @@ import {
   resolveEntryVoiceSlot,
   parseQuizId,
   resolveQuizId,
+  withResolvedQuizIds,
 } from "@adt/types"
 import {
   WebRenderingOutput,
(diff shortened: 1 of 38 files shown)


gbergengruen:
## Review + follow-up fix (2 commits pushed)

Reviewed this on top of the merged #784. The approach is right — stored `quizId`, `resolveQuizId` on reads, `ensureQuizIds` on writes, `usedQuizIds` reserving every id across stored versions — and all the identity-bearing positional sites in `text-catalog`, `packaging/web`, `adt-preview` and the storyboard were converted correctly.

But one case survived, and it's the same corruption this PR exists to prevent.

### The gap

`ensureQuizIds` back-fills legacy ids from the positions of **the array it is handed**, and both mutating routes handed it the array *after* the edit. So the first structural quiz edit on a book that predates `quizId` — that is, every book that exists today — stamped the wrong ids.

Legacy book `[A, B, C]`, catalog keys `qz001/qz002/qz003`. `QuizzesView.deleteQuiz` filters `A` out and PUTs `[B, C]`:

| quiz | catalog key in use | id stamped | result |
|---|---|---|---|
| B | `qz002` | `qz001` | inherits A's translations + audio |
| C | `qz003` | `qz002` | inherits B's translations + audio |

A mid-book insert did the same to the newcomer. Only appending at the tail was safe.

The existing tests missed it because every delete/insert case operates on an array a prior write had already stamped — `const existing = storedQuizzes()` in `quizzes.test.ts`, `const existing = ensureQuizIds(...).output` in `quiz.test.ts`.

### The fix — 201f4fbe

`withResolvedQuizIds` fills in every id from `resolveQuizId` without allocating anything, so the result is precisely what the read paths derive from the same stored array.

- **GET** returns it, so the studio always holds ids and sends them back on a PUT.
- **`generate-one`** applies it to the stored set *before* inserting, since it reads storage rather than the request body.
- **PUT** keeps `ensureQuizIds` as the safety net for direct API callers.

GET stays side-effect free — resolving on read rather than persisting a back-fill means the ids the UI shows can never diverge from the ids packaging and the text catalog use, including after a rollback to an unstamped version.

Four smaller findings in the same commit:

- **`parseQuizRouteId`** replaces the route parsing duplicated between `StoryboardView` and `StoryboardIndex`. Only the former carried the legacy `quiz-{index}` fallback, so an old link rendered the right quiz but highlighted no sidebar row.
- **`MAX_QUIZ_SEQ` is now enforced** — it was exported but never read, so the allocator could mint `qz1000` and silently widen the `qz\d{3}` shape. `ensureQuizIds` throws `QuizIdExhaustedError`, the routes map it to a 400. Mirrors `SectionIdExhaustedError` from #784.
- **The AGENTS.md sample quiz** no longer hardcodes `qz001`, which would document catalog keys and `.mp3` filenames absent from the bundle once ids are sparse.
- **`QuizItem` gains `quizId`**, and the version diff keys on it instead of the question text — an edited question now reads as one edit rather than remove+add.

`resolveQuizId` also now documents that it is only valid on a wholly stamped or wholly unstamped array; on a half-stamped one its positional fallback can collide with a stored id.

### Tests — b53f680b

`apps/api/src/routes/quiz-id-lifecycle.test.ts` drives the real routes against a real sqlite book and asserts on the **text catalog**, not on `quizId`. That distinction is the point: `${quizId}_que` is what the translation step writes against and what `speech.ts` names each mp3 after, so an id that survives a delete but no longer keys the same question is exactly the corruption being tested for — and an ids-to-ids comparison would call it a pass. Covers a legacy book left alone, its first delete / insert / reorder, a book that already carries ids, never-reuse across versions, and rollback.

`package-web.test.ts` gains the export half: a book whose ids are sparse and out of order (`qz004` ahead of `qz003`, `qz001` and `qz002` long deleted) must produce those filenames in `pages.json`, in the answer keys and in the SCORM activity list — and must not produce a `qz001.html`.

Both were checked against the unfixed code rather than assumed to work: on 9b323a4c the lifecycle tests fail 4 of 9, and the packaging test reports `qz001/qz002` where `qz004/qz003` belong.

### Verified against a real book

I also ran the whole QA plan against a real book (4 quizzes, 28 quiz mp3s on disk) through the live API — real export, real `generate-one` LLM call. 17/17 checks pass on this branch.

The same run against 9b323a4c reports the failure in the terms that matter:

```
qz001_que: "Why couldn't Hyena fly?" -> "What came off in Hyena's hand?"
qz004_que.mp3 already exists — it would belong to a different quiz
```

`qz001_que.mp3` reads *"Why couldn't Hyena fly?"* aloud, but after the delete that key names a different question. That harness makes real LLM calls, so it isn't committed.

### Checks

`pnpm typecheck` clean · `pnpm test` 3455 passing across 267 files (+10) · `pnpm lint` 0 errors.

### Cleared as non-issues

- **Layer rule** — studio importing `resolveQuizId` / `parseQuizRouteId` from `@adt/types` matches existing precedent (`isTtsExcluded`, `safeParseModelId`, `isHeadingRole`, `getStageClearOrder`).
- **`adt-runtime`** — only `id.startsWith("qz")`; digit-count agnostic. `activity-quiz.ts` sniffs `data-section-type`, never the id.
- **EPUB / PNLD / WebPub / SMIL / SCORM** — all derive from the single `pages.json` that `packaging/web.ts` builds, so they inherit the fix. No sign-language or TOC entity references a quiz id.
- **Full-stage reruns restarting at `qz001`** — intentional and matches pre-existing behaviour; a stage run replaces the whole set, and speech is content-hash cached.

gbergengruen:
## Reviewed, plus verified end-to-end against a real book with real TTS

Coverage of quiz-id *producers* is complete — I traced every site that mints or resolves a `qz*` id (stage run, text catalog, web packaging, AGENTS.md sample, all three adt-preview paths, the studio sidebar/detail/`keyOf` sites, EPUB, adt-runtime) and there's no positional derivation left anywhere. EPUB correctly inherits from the web bundle's `pages.json`; `buildSectionIdToPageIndex`, `AddQuizDialog` and `QuizJumper` only count or anchor, so they're fine as-is.

I did find one gap, and rather than argue about it from the code I ran the whole chain on a real book.

### The gap

`usedQuizIds` reserves only *explicit* `quiz.quizId` fields. A pre-`quizId` version stores none — its ids existed solely as array positions — so **the ids a legacy version spent are invisible to the reservation**. Delete those quizzes on a book's *first* edit and the ids retire without ever being written down; the next quiz added can be handed one.

The PR's own tests miss it because `"never reissues the id of a deleted quiz"` does a stamping PUT *before* deleting (so the deleted id is explicit), and the `"legacy book's first edit"` tests never add a quiz after a delete.

Worth noting this is exactly what #784's sibling helper already warns about, in `collectSpentSectionIds`:

> A version that no longer satisfies today's schema (a legacy row with no `sectionId` field, a shape from before a migration) still burned the ids it contains, and skipping it because `safeParse` failed would let them be reissued.

That also flags a second, narrower instance here: `if (!parsed.success) continue` drops a whole version's ids, and `options: z.array(QuizOption).length(3)` is strict enough for that to bite.

### The real-book end-to-end

Fixture: `books/raven` — real extraction, real rendering/sectioning, real LLM-generated quizzes. Copied to a temp dir, stored `quizId`s stripped and **pre-baked audio deleted**, so it's a faithful pre-merge book and every mp3 is synthesized fresh during the run.

- **Phase A** — real OpenAI TTS (`gpt-4o-mini-tts`) over all 28 `qz*` catalog entries
- **Phase B** — delete the first quiz, then add one via `generate-one` with a real quiz-generation LLM call
- **Phase C** — re-run TTS

Phase C is the interesting one: `generateSpeechFile` always writes `${catalogId}.mp3` and keys its cache on the *text*, so it never skips an existing file — it overwrites from cache or from fresh synthesis. That makes "does a TTS re-run heal this?" measurable.

**With the fix:**

```
PHASE B  deleted qz001, newcomer = qz005
         mp3s voicing different text than their id names: 0
PHASE C  re-synthesized: 7  (qz005_que, qz005_o0, …)
         Phase-A files whose BYTES changed: 0
         deleted quiz's audio still on disk: 7 (intact)
```

**On this branch as it stands:**

```
PHASE B  deleted qz001, newcomer = qz001
         mp3s voicing different text than their id names: 7
   qz001_o1_exp.mp3 says "✅ Correct! Hyena said he would love to…"
                    id now keys "❌ Not quite! The story says Hyena and…"
PHASE C  re-synthesized: 7  (qz001_que, qz001_o0, …)
         Phase-A files whose BYTES changed: 7
```

### What the TTS re-run settles

The corruption is **not usefully self-healing**, and it fails in two different ways:

1. **Without a TTS re-run** — the normal case, since TTS is the expensive stage and nobody re-runs it after a small quiz edit — 7 files ship in the bundle voicing the wrong content. The `qz001_o1_exp.mp3` line above is the concrete harm: a file that says *"Correct!"* out loud lands on an option whose real explanation is *"Not quite!"*. A learner using read-aloud is told a wrong answer is right.

2. **With a TTS re-run** — the current version is repaired, but the re-run **overwrites the deleted quiz's audio** (7 files, bytes changed). Roll back to the version that still contains that quiz and its audio is now the newcomer's voice — a state no further TTS run fixes without another full re-synthesis. That breaks the rollback guarantee in core principle 2.

So the fix isn't only about wasted synthesis; it's the difference between "orphaned audio nobody reads" and "wrong audio that either ships or destroys a recoverable version."

### Proposed fix

Superseded versions now contribute their positional ids too, and the version scan is lenient instead of `safeParse`-and-skip.

The current version is deliberately **exempt** from the positional reservation — that part is load-bearing. It's the base an incoming body is presumed to descend from, so a caller sending quizzes without ids is still *claiming* its positional ids rather than asking for fresh ones. Reserving them unconditionally would make the first PUT on a legacy book stamp `qz004/qz005/qz006` instead of the `qz001/qz002/qz003` its catalog is already written against — i.e. it would break this PR's own back-compat test. `getLatestNodeData` and `readCurrentNodeRow` use identical SQL, so that exemption tracks exactly what `GET` serves, including after a rollback.

Four tests added, each verified to fail against the current branch and pass with the fix:

- `does not reissue an id the legacy version spent but never stored` — the core repro
- `reserves ids held by a version today's schema no longer accepts` — the lenient-read gap
- `does not later hand a newcomer the audio of the quiz that first edit deleted` — the same case stated as catalog entries and `.mp3` filenames, in the lifecycle file's idiom
- `still stamps positional ids when an unstamped body re-saves the current version` — passes either way, but pins the current-version exemption so a future tightening can't silently break back-compat

`pnpm typecheck` clean · `pnpm test` 3459 passing · `pnpm lint` 0 errors.

**Not pushed yet** — the fix and tests are a local commit awaiting review, so the diff above isn't on this PR. The E2E harness is deliberately not committed either: `books/` is gitignored and it needs a local real book plus a `claude` CLI and a TTS key, so it can't run in CI.

### Minor, non-blocking

1. `generateAllQuizzes` passes no `reservedIds`, so a full stage rerun restarts at `qz001` and can hand fresh content a retired id. Deliberate, and a rerun regenerates every quiz anyway — but `Quiz.quizId`'s doc says "allocated once and never reused" unqualified. Worth a caveat there.
2. `parseQuizId` widened from `/^qz(\d{3})$/` to `/^qz(\d+)$/`, so `qz1`/`qz1000` now enter adt-preview's quiz branch. 404 either way — cosmetic.
3. Legacy `quiz-{n}` deep links resolve positionally, so on an already-stamped book an old bookmark can land on "Quiz not found" instead of the quiz it named. Graceful, no crash — the back-compat only really helps unstamped books.
4. `usedQuizIds` JSON-parses every stored version on every PUT. Fine at current scale, same shape as #784.

The test files here are genuinely strong, by the way — asserting on the catalog rather than on `quizId` alone is the right instinct, and it's what made the missing case easy to spot by contrast.

gbergengruen:
Pushed as `c523cda7` — the fix and its four tests are now on this PR, so the diff described above is live. The "not pushed yet" caveat in my previous comment no longer applies.

Still not committed, deliberately: the real-book E2E harness. `books/` is gitignored and it needs a local book plus a `claude` CLI and a TTS key, so it can't run in CI.

PR #791, opened 2026-08-11
feat(desktop): show onboarding once per installed version

Eliezir:
## What

Keys first-run onboarding completion on the **app version** instead of a boolean flag, so onboarding runs once per installed version.

`apps/desktop/src/main/services/onboarding-state.ts`:
- `hasCompletedOnboarding()` → `readState().completedVersion === app.getVersion()`
- `markOnboardingCompleted()` → writes `{ completedVersion: <app version> }`

## Why

Completion is persisted in `onboarding-state.json` under the app's `userData`, and **macOS keeps `userData` across uninstall/reinstall**. A plain boolean flag therefore suppresses onboarding forever after the first run — even on a brand-new install of a newer build. Keying on the version fixes that: a fresh install (no state) and every install/update run the flow once.

## Behavior

| State | Shows onboarding? |
|---|---|
| No state file (fresh install) | ✅ yes |
| Legacy `{"completed":true}` (older builds) | ✅ yes, once (migration) |
| `completedVersion` == current version | ❌ no (already onboarded) |
| `completedVersion` != current version (update / new install) | ✅ yes |

Since every release bumps the version, each distributed install/update shows onboarding once.

**Caveat:** reinstalling the *exact same version* won't re-show it (the persisted version still matches, and there's no reliable macOS "fresh install" signal). Normal releases always bump the version.

## Testing

Verified in a local unpacked build (`build:unpack`) by driving `onboarding-state.json` at the real userData path (`~/Library/Application Support/@adt/desktop/`):
- no file → onboarding ✅
- legacy `{"completed":true}` → onboarding ✅
- matching version → main window, no onboarding ✅
- mismatched version → onboarding ✅

Only the main-process gate changes; the renderer's `/` localStorage guard is unaffected (the onboarding window loads `/onboarding` directly).

### DIFF
--- a/apps/desktop/src/main/services/onboarding-state.ts
+++ b/apps/desktop/src/main/services/onboarding-state.ts
@@ -3,6 +3,9 @@ import { readFileSync, writeFileSync } from "node:fs";
 import { join } from "node:path";
 
 interface OnboardingState {
+  /** App version for which the user last finished onboarding. */
+  completedVersion?: string;
+  /** Legacy flag from builds that tracked completion as a boolean. Ignored. */
   completed?: boolean;
 }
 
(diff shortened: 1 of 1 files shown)


Eliezir:
Closing: keying onboarding on the app version re-shows the first-run flow after every auto-update, which we don't want. The existing first-run boolean gate on `develop` already shows onboarding only on a first install (the `completed` flag persists in userData across updates), so no change is needed.

PR #794, opened 2026-08-11
feat(desktop): periodic update checks

Eliezir:
**Stack 3/3** (on top of the update-UI redesign #792). Polls for updates every 6h so a release surfaces mid-session, not only at launch. Skips while a check/download/install is already in flight; idempotent and no-op when unpacked/dev. Pairs with the ambient card, which then appears on the fly.


Eliezir:
Folded into #792 (single PR stacked on the new-UI branch).

PR #799, opened 2026-08-12
fix(tts): log CLI synthesis calls

### LINKED ISSUE #715: CLI/runFullPipeline TTS runs produce no debug log rows
### Preliminary Checks
- [x] I am filing a bug report.

### What happened?

A book whose Speech step was run through the CLI (`runFullPipeline`) has **zero `tts` rows** in `llm_log`. The debug panel's Log tab is empty for that step, so there is no record of which provider, model, voice or settings produced the audio.

### Expected behavior

CLI and API runs produce equivalent log rows, so the debug panel is equally useful regardless of how the step was run.

### Root cause

`runTtsItem` in [`packages/pipeline/src/pipeline-dag.ts`](../blob/develop/packages/pipeline/src/pipeline-dag.ts) calls `generateSpeechFile` and emits only a `step-progress` event. It never writes a log entry.

The DAG *does* have an `onLlmLog` callback wired — but it is only passed into `createLLMModel`, so it covers regular LLM calls and not TTS.

By contrast the API's `stage-runner.ts` writes a row per item (success and failure), which is why this only shows up for CLI-run books.

### Proposed solution

Have the DAG's TTS executor write log rows through the existing `onLlmLog`, matching what `stage-runner.ts` records for the same item.

Best done together with the companion issue on consolidating the four hand-rolled TTS log constructions: if a shared builder lands in `@adt/pipeline`, the DAG can adopt it directly rather than becoming a fifth copy.

### Steps to reproduce

1. Run a book's Speech step via the CLI rather than the Studio.
2. Open the book in the Studio → debug panel → Log tab, filter to the `tts` step.
3. No rows.

### What area is this issue related to?

Pipeline — Speech

### Additional context

Found while adding request-parameter logging for ElevenLabs TTS (PR #641); predates it and affects all providers.

gbergengruen:
## Summary

Closes #715.

- Records each CLI/DAG TTS synthesis outcome in the book debug log using the shared TTS log builder.
- Captures success, cache hits, provider failures, retry attempt counts, and ElevenLabs request parameters.
- Logs intentional no-audio outcomes (such as punctuation-only text) for complete traceability.

## Verification

- 
- 
 RUN  v4.1.9 /tmp/adt-studio-715

 Test Files  2 passed (2)
      Tests  117 passed (117)
   Start at  15:27:56
   Duration  1.59s (transform 1.52s, setup 0ms, import 2.13s, tests 306ms, environment 0ms) (117 passing)
- Manual CLI run confirmed a cached TTS entry appears in the ADT Studio Debug panel.

## Notes

The change preserves existing synthesis behavior, provider routing, ElevenLabs retry/backoff behavior, cache behavior, and CLI progress reporting.

### DIFF
--- a/packages/pipeline/src/pipeline-dag.ts
+++ b/packages/pipeline/src/pipeline-dag.ts
@@ -90,6 +90,8 @@ import {
   resolveSpeechModel,
   resolveSpeechFormat,
   generateSpeechFile,
+  buildTtsLogEntry,
+  buildElevenLabsTtsLogParams,
   findAdjacentSpeechText,
   elevenLabsVoiceSettingsFromConfig,
   classifyElevenLabsTtsError,
@@ -1112,6 +1114,7 @@ export async function runFullPipeline(
       const elevenLabsVoiceSettings = elevenLabsVoiceSettingsFromConfig(config.speech)
 
       const runTtsItem = async (item: TTSWorkItem) => {
+        const startedAt = Date.now()
         const provider = resolveProviderForLanguage(item.language, routing)
         const providerModel = resolveSpeechModel(provider, providerConfigs, speechModel)
         const outputFormat = resolveSpeechFormat(provider, config.speech?.format)
@@ -1123,9 +1126,25 @@ export async function runFullPipeline(
           provider === "openai" || provider === "gemini"
             ? resolveInstructions(item.language, instructionsMap)
             : ""
-        const ttsSynthesizer = getSynthesizer(provider)
-        const generate = () =>
-          generateSpeechFile({
+        const logParams = provider === "elevenlabs"
+          ? buildElevenLabsTtsLogParams({
+              model: providerModel,
+              voice,
+              language: item.language,
+              format: outputFormat,
+              sampleRate: config.speech?.sample_rate,
+              bitRate: config.speech?.bit_rate,
+              applyTextNormalization: config.speech?.elevenlabs_apply_text_normalization,
+              previousText: item.previousText,
+              nextText: item.nextText,
+              ...elevenLabsVoiceSettings,
+            })
+          : undefined
+        let attemptCount = 0
+        const generate = () => {
+          attemptCount++
+          const ttsSynthesizer = getSynthesizer(provider)
+          return generateSpeechFile({
             textId: item.textId,
             text: item.text,
             language: item.language,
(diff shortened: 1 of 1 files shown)


gbergengruen:
## Manual test pass — results

Tested on `codex/715-cli-tts-debug-log` @ `fd439eb8` against a fresh 2-page book (`tts-log-799`, pages 1–2 of `momo-p001-010.pdf`), OpenAI provider (`gpt-4o-mini-tts`), verifying rows directly out of `llm_log`.

### Passed

| Check | Result |
|---|---|
| `tts` rows appear at all | ✅ from a **zero-row baseline** — the pre-run book had `metadata`/`book-summary`/`book-outline`/`page-sectioning`/`image-*` rows and no `tts` row, so every row below is attributable to this PR |
| Row field correctness | ✅ `promptName=tts-openai`, `modelId=openai/gpt-4o-mini-tts`, `pageId=<textId>`, `attempt=1`, `cacheHit=false`, ~1.9 s durations |
| Message content | ✅ `[en] voice=alloy\n<text>`, complete and untruncated |
| Cache hits | ✅ second run logged `cacheHit=true` at **2 ms** vs ~1900 ms fresh; earlier rows retained (`llm_log` is genuinely append-only) |
| Row count vs real work | ✅ within each run: 2 rows = 2 audio files = 2 ready Core TTS entries (run 1), 4 = 4 = 4 (run 2) |
| No spurious `params` | ✅ OpenAI rows carry no Request-settings block (ElevenLabs-only, as intended) |

A nice incidental confirmation that the log records what was **actually sent**: `pg001_n0003` logged as `"Class one, Book eleven"` where the source text is `"Class I Book 11"` — i.e. the post-normalization string.

### Not covered

Provider-failure logging, the missing-key case, the ElevenLabs `params` grid, the punctuation-only no-audio case, and a Gemini run all need provider keys in the runner's shell — deferred to whoever picks this up.

**Retry attribution** (`attempt: 3` after two 429s, logged as one row) remains the only genuinely unverified behavior, and it is not manually verifiable: `ELEVENLABS_TTS_MAX_CONCURRENCY` deliberately keeps bursts below the limits that would trigger 429s. If we want it covered, the honest route is a unit test with a stub synthesizer that throws 429 twice then succeeds. Happy to add it here if reviewers want it before merge.

### Upstream issues this logging exposed

**None of these are caused by this PR** — the new rows are what made them visible, which is a decent argument for merging it. Each deserves its own issue; none should block.

1. **Two identical runs synthesized different work.** Run 1 produced audio for 2 entries, run 2 for 4 — the glossary term and its definition (`gl001`, `gl001_def`) got audio **only in run 2**, from the same PDF and config. In run 1 the glossary was silently left with no audio. Lead worth checking (not proven from post-hoc state): `text-catalog` is written twice per run — by the `text-catalog` step (`pipeline-dag.ts:848`, 260 b) and again by `package-web` (`packaging/web.ts:325`, 404 b) — while `core-tts-catalog` reads it *before* the Package stage rewrites it.
2. **Normalized speech text drifts, defeating the TTS cache.** `pg001_n0003` normalized to `"Class one, Book eleven"` in run 1 and `"Class one Book eleven"` in run 2 (comma dropped) from identical source text. `computeSpeechCacheKey` hashes the text, so the entry re-synthesized at full cost and orphaned its old cache file (5 cache files for 4 entries).
3. **The TTS-normalization LLM call never hits its cache** — `cacheHit=false` on both runs (2.4 s, 3.9 s). Direct cause of 1 and 2, and the reason no CLI re-run is free.
4. **A CLI re-run wipes all node data.** `extract` calls `clearExtractedData()` (`pdf-extraction.ts:104` → `book-storage.ts:587`), deleting every `node_data` row except `font-registry`/`font-assignment`, plus `images`/`pages`/`step_runs`. Entity versions restart at v1 each run. Worth knowing for anyone reasoning about run history — `llm_log` survives, node data does not.
5. **The CLI reports skipped steps as complete.** `cli.ts` fills the progress bar on `step-skip` identically to `step-complete`, so a *skipped* Speech stage rendered `100/100` while the run had failed upstream. Only the "Failed stages" footer gave it away.

### Two blockers hit while setting up (both unrelated, both pre-existing)

- **Template rendering crashes on a cover with a subtitle.** A fresh label inherits the global `default_render_strategy: two_column`, and the storyboard stage then fails with `Heading role "subheading" (pg001_n0003) must use <h3 class="adt-h3">`. Cause: the front cover has one heading *container* with two differing-role children (`chapter_title` + `subheading`); the template's heading branch (`_render_node.liquid:42-47`) takes its tag from `children[0]` only and renders both children inline inside it, so the subheading lands in the `<h1>` and `validate-typography-hierarchy.ts:176` rightly rejects it. Title + subtitle is a very common cover pattern, so this likely breaks any book rendered with a template strategy. **Worth its own issue.** Worked around here by setting `default_render_strategy: llm`.
- Stale git worktree registrations under `/tmp/adt-studio-*` blocked `git checkout` of this branch; `git worktree prune` clears them. Local-environment only.

### Recommendation

The PR does what it says and the evidence is clean. I'd merge on the strength of the above — the remaining manual checks exercise error paths that #797 already shipped, plus provider plumbing.

Two notes for the merge itself: this is the base of a stack (**#800** → **#801**), so **please don't squash** or the stacked branches will conflict; and `develop` has moved on by 3 commits (#805 + two staging-CI commits, none touching speech), so a rebase before merge would mean merging what was actually tested.

gbergengruen:
## Review continued — rebased, re-verified, ready to merge

Picking up from the manual test pass above.

### Rebase

Rebased onto `develop` @ `52974df6` (the 3 commits noted as missing) and force-pushed: `fd439eb8` → `15f51092`. Clean rebase, no conflicts. What CI now runs is what lands.

### Automated verification on the rebased tree

| Check | Result |
|---|---|
| `pnpm typecheck` | ✅ clean (`tsc --build`, strict) |
| `pnpm test` | ✅ **214 files / 2737 tests passed**, 0 failed |
| `pnpm lint` | ✅ 0 errors (17 pre-existing warnings, all in generated `src/paraglide/*`) |

### Code re-read — two things worth recording

**1. `getSynthesizer` moved inside `generate()` is safe.** It now runs per attempt rather than once per item, but it memoizes into the `synthesizers` map (`pipeline-dag.ts:1023-1054`), so no synthesizer is constructed twice.

**2. It also means credential errors are now logged — and still fail fast.** Because `getSynthesizer` is inside the `try`, a missing key throws *inside* the logged region, so `ELEVENLABS_API_KEY is required…` now produces a `success:false` row instead of escaping unlogged. It does **not** get caught in the retry loop: `classifyElevenLabsTtsError` finds no HTTP status and no transport-error match, so it returns `"permanent"` and the loop rethrows on the first pass (`speech.ts:347-353`). This closes part of the "missing-key case" gap listed as *Not covered* above, by inspection rather than by run.

Retry/backoff arithmetic is unchanged — the loop counter was only renamed `attemptCount` → `retryCount` and `elevenLabsTtsRetryDelayMs(retryCount)` still receives the same 1-based value; the new `attemptCount` is incremented inside `generate()` purely for log attribution.

### Retry attribution (`attempt: 3`) — left uncovered, deliberately

Still the one unverified behavior. Confirmed there is no existing test harness to hang it on: nothing under `packages/pipeline/src/__tests__/` exercises `runFullPipeline`, and `runTtsItem` is a closure inside it, so a stub-synthesizer test would require extracting the retry+log wrapper into `speech.ts` first. That refactor is worth doing but it is not this PR — it would also invalidate the manual evidence gathered above. **Filing separately rather than expanding scope here.**

### Merging

Merging with a **merge commit** (not squash), as flagged above, so the stack stays intact. #800 will auto-retarget to `develop` and needs a rebase before it can be read as a clean diff; #801 follows behind it.

The 5 upstream issues and the 2 setup blockers from the manual pass are untouched by this and still need their own issues.

PR #800, opened 2026-08-12
fix(speech): fail fast when a TTS provider credential is missing

gbergengruen:
## Summary

A missing TTS provider credential produced **one logged failure per item** instead of one clear "key is missing" error.

Both TTS execution paths construct their synthesizer inside the per-item `try`, so an absent key is rediscovered once per item and recorded as a per-item synthesis failure:

- the API stage-runner logs a failure row for **every** entry and still completes the step with gaps;
- the CLI/DAG path logs one row per in-flight item before aborting.

Either way the real cause — a single missing credential — is buried under N identical rows that each look like a synthesis attempt.

## Changes

- Build every synthesizer the run needs **before** admitting any item, in both paths (`stage-runner.ts`, `pipeline-dag.ts`).
- Reuses the existing memoized `getSynthesizer` rather than adding a shared credential helper, so each path keeps its own error wording (the API points at the API Keys dialog, the CLI at env vars).
- The provider set is derived from the **resolved work items** (`item.provider`), plus `"gemini"` when any page groups exist. Not from the configured providers: a run whose entries are all reused needs no synthesizer and must not be failed here.

## Rebased onto `develop` — one behavioural change from the original

This PR predates the multi-voice work (#780 / #825). That work matters here:

**A secondary narrator carries its own provider**, straight from `speech.secondary_voices[<locale>].provider` (`speech.ts:399`), independent of the language's routing. The original pre-flight derived its set via `resolveProviderForLanguage(item.language, routing)` — which on today's `develop` **misses the secondary voice entirely**, i.e. exactly the case the PR exists to catch.

Deriving from `item.provider` fixes that and is simpler, since both `TTSWorkItem` types now carry a resolved provider. New test `"fails fast when only the secondary narrator's provider credential is missing"` pins it; with the old form that test **resolves instead of rejecting**.

> Note for reviewers: the `pipeline-dag.ts` hunk auto-merged cleanly but did **not** compile — `develop` deleted `resolveProviderForLanguage` and `routing` from that scope. Caught by `pnpm typecheck`.

## Behavior change

For a missing credential the API speech stage now **fails immediately** instead of reporting N failed items and completing with gaps. Every item would have failed anyway, and no `llm_log` noise is written — but this does change the error shape the Speech view receives, so it's worth a look.

## Verification

`pnpm build`, `pnpm typecheck`, `pnpm lint` (0 errors), full suite **2889 passing**.

Four tests in `stage-runner.test.ts`, each confirmed failing without the fix:

| Test | Guards |
|---|---|
| ElevenLabs routing, no key | run rejects with the provider-named message, `generateSpeechFile` never called, **zero** `tts` rows, `step-error` carries the message |
| Gemini `pageGroups` path | `pageGroups` skips the per-entry reuse check, so it can carry work the per-entry path never sees |
| Every entry reused | the pre-flight must **not** over-fire — a run needing no synthesizer still succeeds with no credential |
| Secondary narrator on its own provider | **new** — the multi-voice hole described above |

### Live run on `books/tts-log-799` (free — no key, no spend)

| Test | Result |
|---|---|
| **A** — `default_provider: elevenlabs`, no key | 4 work items; `creating TTS synthesizer for provider="elevenlabs"` logged **once**; run failed with the single clear error; `tts` rows still **6** |
| **B** — `openai`, no key | `generating TTS for 0 entries and reusing 4`; routing line empty; **no** `creating TTS synthesizer` line at all; run succeeded; rows still **6** |
| **A′** — primary `openai` (reused), secondary `elevenlabs` (no key) | routing `en:secondary→elevenlabs` ×4; synthesizer built **once**; one clear error; rows still **6**. The primary needs nothing, yet the secondary's missing key is still caught up front |

## Known gap

No automated test for the **CLI-side** pre-flight. `runFullPipeline` has no test harness and exercising it means a full paid run. #807 proposes extracting a testable helper from that same closure — the pre-flight should ride along with it. The CLI side is covered by typecheck and the shared `item.provider` derivation.

### DIFF
--- a/apps/api/src/services/stage-runner.test.ts
+++ b/apps/api/src/services/stage-runner.test.ts
@@ -1505,6 +1505,145 @@ speech:
     } finally {
       db.close()
     }
+
+    // The Speech view is driven by step-error, not by the rejection: assert the
+    // step goes red carrying the provider message, so the user sees one
+    // actionable error rather than a bare run failure with no step highlighted.
+    const ttsErrors = events.filter(
+      (event) => event.type === "step-error" && event.step === "tts"
+    )
+    expect(ttsErrors).toHaveLength(1)
+    expect((ttsErrors[0] as { error: string }).error).toMatch(
+      /ElevenLabs API key is required/
+    )
+  })
+
+  it("fails before any page-batched synthesis when the Gemini credential is missing", async () => {
+    tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), "stage-runner-tts-"))
+    const booksDir = path.join(tmpDir, "books")
+    const promptsDir = path.join(tmpDir, "prompts")
+    const configPath = path.join(tmpDir, "config.yaml")
+    fs.mkdirSync(promptsDir, { recursive: true })
+    // batch_by_page routes page-scoped Gemini entries into pageGroups, which
+    // deliberately skip the per-entry reuse check — so this is the path that
+    // reaches the pre-flight via the pageGroups half of its provider set.
+    fs.writeFileSync(
+      configPath,
+      `role_types:
+  section_text: Main body text
+structure_types:
+  paragraph: Paragraph
+speech:
+  default_provider: gemini
+  batch_by_page: true
+  providers:
+    gemini:
+      languages:
+        - en
+`
+    )
+    seedTextAndSpeechBook(booksDir, "gemini-batched-missing-key")
+    vi.stubEnv("GEMINI_API_KEY", "")
+
+    const events: ProgressEvent[] = []
+    const runner = createStageRunner()
+    await expect(
+      runner.run(
+        "gemini-batched-missing-key",
+        {
+          booksDir,
+          apiKey: "sk-test",
+          promptsDir,
+          configPath,
+          fromStage: "translate",
+          toStage: "speech",
+        },
+        { emit: (event) => events.push(event) }
+      )
+    ).rejects.toThrow(/Gemini API key is required/)
+
+    const db = openBookDb(
+      path.join(booksDir, "gemini-batched-missing-key", "gemini-batched-missing-key.db")
+    )
+    try {
+      expect(db.all("SELECT request_id FROM llm_log WHERE step = 'tts'")).toHaveLength(0)
+    } finally {
+      db.close()
+    }
+    expect(transcribeWithWhisperMock).not.toHaveBeenCalled()
+  })
+
+  it("does not require a credential when every entry is reused and there is no work", async () => {
+    tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), "stage-runner-tts-"))
+    const booksDir = path.join(tmpDir, "books")
+    const promptsDir = path.join(tmpDir, "prompts")
+    const configPath = path.join(tmpDir, "config.yaml")
+    fs.mkdirSync(promptsDir, { recursive: true })
+    fs.writeFileSync(
+      configPath,
+      `role_types:
+  section_text: Main body text
+structure_types:
+  paragraph: Paragraph
+speech:
+  default_provider: elevenlabs
+  providers:
+    elevenlabs:
+      languages:
+        - en
+`
+    )
+    const label = "elevenlabs-tts-all-reused"
+    seedTextAndSpeechBook(booksDir, label)
+
+    // A manual recording with its audio file present is reusable outright, so
+    // the entry never becomes a work item. With nothing to generate, the
+    // pre-flight's provider set is empty and no synthesizer is built — a run
+    // that needs no credential must not be failed by the fail-fast check.
+    const bookDir = path.join(booksDir, label)
+    const audioDir = path.join(bookDir, "audio", "en")
+    fs.mkdirSync(audioDir, { recursive: true })
+    fs.writeFileSync(path.join(audioDir, "pg001_t001.mp3"), Buffer.from("fake-audio"))
+    const storage = createBookStorage(label, booksDir)
+    try {
+      storage.putNodeData("tts", "en", {
+        entries: [
+          {
+            textId: "pg001_t001",
+            language: "en",
+            fileName: "pg001_t001.mp3",
+            voice: "manual",
+            model: "manual",
+            cached: true,
+            provider: "manual",
+          },
+        ],
+        generatedAt: "2026-01-01T00:00:00.000Z",
+      })
+    } finally {
+      storage.close()
+    }
+    vi.stubEnv("ELEVENLABS_API_KEY", "")
+
+    const events: ProgressEvent[] = []
+    const runner = createStageRunner()
+    await runner.run(
+      label,
+      {
+        booksDir,
+        apiKey: "sk-test",
+        promptsDir,
+        configPath,
+        fromStage: "translate",
+        toStage: "speech",
+      },
+      { emit: (event) => events.push(event) }
+    )
+
+    expect(generateSpeechFileMock).not.toHaveBeenCalled()
+    expect(
+      events.filter((event) => event.type === "step-error" && event.step === "tts")
+    ).toHaveLength(0)
   })
 
   it("stops admitting TTS items and unwinds without step errors when the run is cancelled", async () => {
--- a/apps/api/src/services/stage-runner.test.ts
+++ b/apps/api/src/services/stage-runner.test.ts
@@ -4,7 +4,7 @@ import path from "node:path"
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest"
 import { PIPELINE, type AppConfig, type ProgressEvent } from "@adt/types"
 import { computeSpeechCacheKey, stripEmojis } from "@adt/pipeline"
-import { createBookStorage } from "@adt/storage"
+import { createBookStorage, openBookDb } from "@adt/storage"
 import {
   buildStageRunnerImageClassifyConfig,
   createStageRunner,
@@ -1167,6 +1167,7 @@ describe("createStageRunner speech Gemini partial failures", () => {
   })
 
   afterEach(() => {
+    vi.unstubAllEnvs()
     if (tmpDir) {
       fs.rmSync(tmpDir, { recursive: true, force: true })
       tmpDir = ""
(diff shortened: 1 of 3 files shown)


gbergengruen:
## Retargeted to `develop` — picking up here tomorrow

#799 merged as `ba720b13`. I've **retargeted this PR from `codex/715-cli-tts-debug-log` to `develop`**, but *not* rebased yet.

⚠️ **The diff above is currently misleading.** This branch's head still sits on `fd439eb8`, the pre-rebase commit of #799, so the merge-base against `develop` is `7d151d71` and the diff replays #799's already-merged changes. **Don't review the `pipeline-dag.ts` hunks as if they were new** — only ~11 of those lines belong to this PR. Rebase first (step 1).

```
d975d4c3 fix(speech): fail fast when a TTS provider credential is missing   ← this PR
fd439eb8 fix(tts): log CLI synthesis calls                                 ← superseded by 15f51092 on develop
7d151d71 Merge pull request #797 …
```

---

## Step 1 — Rebase (do this first)

```bash
git checkout fix/speech-fail-fast-missing-credentials
git rebase --onto develop fd439eb8        # drops the superseded #799 commit
pnpm typecheck && pnpm test
git push --force-with-lease
```

Then **#801 needs the same treatment**, since it's stacked on this branch:

```bash
git checkout fix/tts-mark-no-audio-entries
git rebase --onto fix/speech-fail-fast-missing-credentials d975d4c3
git push --force-with-lease
```

Sanity check before reviewing: the diff should collapse to **~+81/−1** across `stage-runner.ts` (+12), `stage-runner.test.ts` (+57) and `pipeline-dag.ts` (+11). If #799's hunks are still showing, the rebase didn't take.

---

## Two risks I chased down and cleared — no need to re-litigate

**1. The pre-flight cannot newly break fully-cached runs.** This was my main worry: `pageGroups` deliberately skips the per-entry reuse check ([`stage-runner.ts:2937-2955`](https://github.com/unicef/adt-studio/blob/ba720b13/apps/api/src/services/stage-runner.ts#L2937-L2955)), so it can be populated when every page's audio is already cached — meaning a pre-flight could fail a run that previously succeeded entirely from cache.

It can't. `generateSpeechFile` early-returns on a cache hit *before* touching the synthesizer ([`speech.ts:745-759`](https://github.com/unicef/adt-studio/blob/ba720b13/packages/pipeline/src/speech.ts#L745-L759)) — but the old code already built the synthesizer first, at [`stage-runner.ts:3217`](https://github.com/unicef/adt-studio/blob/ba720b13/apps/api/src/services/stage-runner.ts#L3217) per item, and at [`3102`](https://github.com/unicef/adt-studio/blob/ba720b13/apps/api/src/services/stage-runner.ts#L3102) as an **argument** evaluated inside the group `try`, before `generatePageSpeechFiles` consults its page cache. Every provider the pre-flight touches would have thrown anyway. The change only moves an unavoidable error earlier and deduplicates it.

**2. The Speech view still marks the step red.** The "changes the error shape" caveat in the PR body is fine. `step-start` for `tts` fires at [`stage-runner.ts:2811`](https://github.com/unicef/adt-studio/blob/ba720b13/apps/api/src/services/stage-runner.ts#L2811), well before the pre-flight, so `runningSteps` contains `tts` and the outer catch at [`914-923`](https://github.com/unicef/adt-studio/blob/ba720b13/apps/api/src/services/stage-runner.ts#L914-L923) converts the throw into a proper `step-error`. [`stage-runner.test.ts:1146`](https://github.com/unicef/adt-studio/blob/ba720b13/apps/api/src/services/stage-runner.test.ts#L1146) already covers that mechanism generically. The user gets one clear toast on a red Speech step instead of N per-item failures — strictly better.

---

## Step 2 — Prove the new test has teeth

Highest value per minute, and currently unverified. Temporarily revert the 12-line pre-flight in `stage-runner.ts` and confirm the new test goes **red**:

```bash
pnpm --filter @adt/api test -- stage-runner.test.ts -t "fails the speech step before any synthesis"
```

A pre-flight test that passes without the pre-flight is worthless. Restore afterwards.

## Step 3 — Close two real gaps in `stage-runner.test.ts`

The harness already exists (`seedTextAndSpeechBook` at [L270](https://github.com/unicef/adt-studio/blob/ba720b13/apps/api/src/services/stage-runner.test.ts#L270); model on the Gemini block at [L1122](https://github.com/unicef/adt-studio/blob/ba720b13/apps/api/src/services/stage-runner.test.ts#L1122)). No new abstraction required.

- **Gemini / `pageGroups` half of the pre-flight.** The new test routes ElevenLabs only, so the `pageGroups` half of the provider `Set` — the part the PR body specifically calls out as the motivation — is untested. Add: page-batched Gemini language, no key ⇒ rejects with the Gemini message, `generatePageSpeechFiles` never called, zero `tts` rows.
- **Zero-work non-regression.** A book where every entry is reused and no key is set must **not** fail (`ttsWorkItems` and `pageGroups` both empty ⇒ empty `Set` ⇒ no synthesizer built). This is the one way an over-eager pre-flight could regress a working setup — assert it explicitly.

Optionally tighten the existing test to assert the emitted `step-error` carries the provider message, not just that the promise rejects.

## Step 4 — Manual UI check (costs nothing)

The run fails before any synthesis, so this is free:

1. `pnpm dev`, open a book with upstream stages already done (`books/tts-log-799` from the #799 pass).
2. Clear the ElevenLabs key in the API Keys dialog; route the book's language to `elevenlabs`.
3. Run **only** the Speech stage.
4. Expect exactly one toast — `ElevenLabs API key is required for ElevenLabs TTS provider. Set it in the API Keys dialog (gear icon).` — the Speech step **red** rather than "completed with gaps", and **zero** new `tts` rows in `llm_log`.
5. Restore the key, re-run, confirm the happy path is unaffected and each provider logs `creating TTS synthesizer for provider=` exactly **once** ([L2835](https://github.com/unicef/adt-studio/blob/ba720b13/apps/api/src/services/stage-runner.ts#L2835)) — the memoization is what makes the pre-flight free.

## Accepted gap

The **CLI/DAG pre-flight stays untested**, as the PR proposes — agreed. `runFullPipeline` has no harness and `pnpm pipeline` takes only `<label> <pdf>` with no stage range ([`cli-args.ts:47-87`](https://github.com/unicef/adt-studio/blob/ba720b13/packages/pipeline/src/cli-args.ts#L47-L87)), so exercising it means a full paid run. It's covered by typecheck and by the symmetry of the two implementations. #807 already proposes extracting a testable helper from this same closure — the CLI pre-flight should ride along with that extraction rather than growing its own abstraction here.

---

## Unrelated: the #799 review findings are now filed

#808 (root cause) · #809 · #810 · #811 · #812. None block this PR.

gbergengruen:
## Manual test plan

Rebased onto `develop` (`cd53b75d`), CI green, diff now clean at **+219/−1** across 3 files. Two test gaps closed (see the commit) — below is what still wants a human.

Everything here is **free**: Tests A and B fail or short-circuit before any synthesis, so no LLM spend. Test C costs a few cents and is optional.

---

### ⚠️ Read this first — the one way to get a false negative

A Speech-only re-run **preserves** the `tts` node so cached and manual entries can be merged into the new version ([`pipeline-effects.ts:241-246`](https://github.com/unicef/adt-studio/blob/cd53b75d/packages/types/src/pipeline-effects.ts#L241-L246)).

`books/tts-log-799` currently holds 4 entries, all `provider=openai voice=alloy model=gpt-4o-mini-tts`, with all 4 audio files present. So **if you leave the provider at `openai` and just clear a key, all 4 entries are reused, there are zero work items, the pre-flight builds nothing, and the run succeeds.** That is correct behaviour (it's Test B), but if you were expecting Test A's error you'd wrongly conclude the fix is broken.

Test A therefore **must** switch the provider to `elevenlabs` — that makes the existing entries non-reusable on a provider mismatch (`canReuseSpeechEntry`), which is what creates the work items the pre-flight acts on.

### Setup

The runtime config is the **repo-root `config.yaml`**, not the copy inside the book directory (`process.env.CONFIG_PATH ?? projectRoot/config.yaml`, [`app.ts:48`](https://github.com/unicef/adt-studio/blob/cd53b75d/apps/api/src/app.ts#L48)). Its `speech:` block is currently commented out, so the provider defaults to `openai`.

```bash
git checkout fix/speech-fail-fast-missing-credentials   # cd53b75d
pnpm build && pnpm dev
```

Record the baseline row count — **it should be 6** — and re-run this after each test:

```bash
python3 - <<'EOF'
import sqlite3, json
db = sqlite3.connect('books/tts-log-799/tts-log-799.db')
rows = db.execute("SELECT timestamp,item_id,success,data FROM llm_log WHERE step='tts' ORDER BY id").fetchall()
print(f"{len(rows)} tts rows")
for ts,item,ok,data in rows:
    d = json.loads(data)
    print(f"  {ts[:19]} {item:16} ok={ok} cacheHit={d.get('cacheHit')} attempt={d.get('attempt')} {d.get('durationMs')}ms {d.get('modelId')}")
db.close()
EOF
```

(Or use the Debug panel's LLM Logs tab if you'd rather not touch the DB.)

---

## Test A — missing credential fails fast (the actual behaviour change)

**Setup.** Add to the root `config.yaml` (the commented example around line 390 is a template):

```yaml
speech:
  default_provider: elevenlabs
  providers:
    elevenlabs:
      languages: [en]
```

Make sure **no** ElevenLabs key is set — clear it in the API Keys dialog (gear icon) *and* confirm `ELEVENLABS_API_KEY` isn't exported in the shell running the API, since the synthesizer factory falls back to the ambient env var.

**Run.** Open `tts-log-799` and run **only the Speech stage**.

**Expect:**

| Check | Expected |
|---|---|
| Error toast | Exactly **one**, reading `ElevenLabs API key is required for ElevenLabs TTS provider. Set it in the API Keys dialog (gear icon).` |
| Speech step | **Red / error** — *not* "completed with gaps" |
| `tts` log rows | Still **6**. No new rows at all |
| API console | `creating TTS synthesizer for provider="elevenlabs"` appears **once**, then the error. No per-item synthesis lines |

The row count is the sharpest signal. Before this PR you'd get one failure row per in-flight item and the step would complete with gaps; now the credential is checked once, before anything is admitted.

**Also worth eyeballing:** the Speech view should render normally with the step marked failed, rather than stranding you on the landing page.

## Test B — a run with no work must NOT be failed by the pre-flight

This is the regression the pre-flight could plausibly have introduced, and the reason it derives its provider set from the work list rather than from the configured providers. Worth doing by hand even though it's now covered by a test.

**Setup.** Revert the root `config.yaml` speech block to `default_provider: openai` (or comment it out again). Leave the OpenAI key **cleared**.

**Run.** Run only the Speech stage on `tts-log-799`.

**Expect:**

| Check | Expected |
|---|---|
| Outcome | Run **succeeds**. Speech step green |
| `tts` log rows | Still **6** — every entry was reused, nothing was synthesized |
| API console | **No** `creating TTS synthesizer` line at all |

If this one fails with a missing-key error, the pre-flight is over-firing — that's a real bug and blocks the PR.

## Test C — happy path still works (optional, costs a little)

Restore a working key for whichever provider you prefer and run Speech again.

**Expect:** audio generated as before; `creating TTS synthesizer for provider="…"` logged **exactly once** per provider even though the pre-flight and the per-item loop both call `getSynthesizer` — that memoization is what makes the pre-flight free. New `tts` rows appear with `attempt=1`.

## Test D — Gemini page-batched path (optional)

Now covered by an automated test, so only worth doing if you want the end-to-end confirmation:

```yaml
speech:
  default_provider: gemini
  batch_by_page: true
  providers:
    gemini:
      languages: [en]
```

with no Gemini key. Expect the same single clear error and **6** rows. This path matters separately because `pageGroups` deliberately skips the per-entry reuse check, so it can carry work the per-entry path never sees.

---

## What I verified so you don't have to

- **The existing test has teeth.** Disabling the 12-line pre-flight makes it fail with `promise resolved "undefined" instead of rejecting` — i.e. the old behaviour of completing the step with the item marked failed. Confirmed, restored.
- **Both new tests have teeth.** Same check: both fail without the pre-flight. The Gemini one matters because without the pre-flight the missing key throws from [`stage-runner.ts:3115`](https://github.com/unicef/adt-studio/blob/cd53b75d/apps/api/src/services/stage-runner.ts#L3115) with the *identical message* — what distinguishes them is that the in-loop throw logs a failure row and completes with gaps instead of rejecting.
- **The no-work test isn't vacuous.** Breaking the reuse (wrong audio filename) makes it fail with the missing-key error, so it genuinely gates on both reuse working and the pre-flight not over-firing.
- **`pnpm typecheck` clean, `pnpm test` 2740 passed, `pnpm lint` 0 errors** (17 pre-existing warnings in generated `src/paraglide/*`).
- Stack: **#801 rebased** onto this branch and pushed; its diff is clean too (+231/−11).

## Accepted gap

The **CLI/DAG pre-flight is still untested**, as the PR proposes. `runFullPipeline` has no harness and `pnpm pipeline` takes only `<label> <pdf>` with no stage range, so exercising it means a full paid run. #807 already proposes extracting a testable helper from that same closure — the pre-flight should ride along with it.

If you'd rather not leave it uncovered, say so and I'll do the extraction here instead.

PR #801, opened 2026-08-12
fix(tts): mark TTS entries that produced no audio on purpose

gbergengruen:
## Summary

`generateSpeechFile` returns `null` for text with nothing speakable in it (an `—` or `…` entry) — the only `null` path it has. Both TTS execution paths logged that as `success: true, cacheHit: false`, so:

- the log could not answer "why is there no audio for this item?" — the row looks like a synthesis that happened;
- the stats query counted it as a **call** and a **cache miss**, and averaged its ~0 ms into step latency, dragging the reported `tts` cache-hit rate and average duration away from the truth.

The row is kept rather than dropped — Principle 4 (Maximum Transparency) is exactly why an unexplained gap is worse than a marked one — but it's now marked.

## Changes

- `LlmLogEntry` gains an optional `skippedReason`. **No DB migration**: the whole entry is already stored as JSON in `llm_log.data`.
- `buildTtsLogEntry` passes it through, so both paths inherit it from the one shared builder (#797's consolidation). `NO_SPEAKABLE_TEXT_REASON` keeps the value identical across them and the log filterable.
- Debug log: a 4th row status — slate dot, **No audio** — with a neutral explanation block, separate from the red Error block. An error still outranks a skip. `getStatus` is exported so that precedence is pinned by a test.
- The Cache detail field shows `—` instead of a misleading **Miss** for a lookup that never happened.
- `/debug/stats` excludes skipped rows from `calls`, both cache counts, and `avgDurationMs`, and reports them in a new `skipped` column surfaced in the Stats tab — so a step whose only rows are skipped isn't silently "0 calls".

## Un-stacked and rebased onto `develop`

Previously stacked on #800 (and #799 before it). #799 merged; this PR's own hunks apply cleanly to `develop` on their own, so it has been **rebased directly onto `develop` and retargeted** — it no longer depends on #800 and can be reviewed and merged independently. The diff is now just this change: **+321/−14**.

### One addition the rebase made necessary

`develop`'s multi-voice work (#780) added a **third** `buildTtsLogEntry` call site this PR predates: `emitPageLog` in the page-batched Gemini path. `generatePageSpeechFiles` filters non-speakable entries and returns `[]` when none remain (`speech.ts`, `usable.length === 0`), so an all-punctuation page group logs the exact `success: true, cacheHit: false` row this PR exists to eliminate. Now marked too, keyed off the empty return, with a covering test.

## Scope notes

- The **single-item TTS route** needs no marker: it already returns 422 for a non-speakable entry *before* writing a log row.
- `StatsTab.tsx` derives `cacheHitRate` from `totals.cacheHits / totals.calls`, so excluding skipped rows from `calls` corrects the headline rate for free.

## Verification

`pnpm typecheck` clean, `pnpm lint` 0 errors, full suite **2892 passing**.

4 tests, all confirmed failing first:
- `buildTtsLogEntry` round-trip + absence when not passed (`speech.test.ts`)
- `getStatus` precedence (`LlmLogsTab.test.tsx`)
- stats exclusion while staying visible in `/debug/llm-logs` (`debug.test.ts`)
- **new** — a page-batched group with no speakable text is marked (`stage-runner.test.ts`), driven end-to-end through the real `generatePageSpeechFiles`, which returns early without reaching the provider

`pnpm --filter @adt/studio extract` ⇒ **0 missing** across `es`, `fr`, `pt-BR`, `sq`. Only one new string; "No audio" already existed and was already translated.

## Still to do before merge

End-to-end run on a real book with an `—` section (needs a paid API key): confirm the marker fires on a genuine punctuation-only entry through the **per-entry** path and compare `/debug/stats` before/after. The automated tests prove the marker, the UI status, the stats math, and the page-batched path end-to-end — not that the per-entry wiring triggers in a live paid run.

Adjacent, worth a look before this lands: **#809** (Core TTS normalization stability).

### DIFF
--- a/apps/api/src/routes/debug.test.ts
+++ b/apps/api/src/routes/debug.test.ts
@@ -261,6 +261,49 @@ describe("Debug routes", () => {
       expect(body.totals.errorCount).toBe(1)
     })
 
+    it("does not count no-audio TTS rows as calls, cache misses, or latency samples", async () => {
+      // A punctuation-only entry never reaches the provider, so counting it
+      // would understate the tts cache-hit rate and skew average latency.
+      const skippedLabel = "tts-skipped"
+      const storage = createBookStorage(skippedLabel, tmpDir)
+      storage.close()
+      const db = openBookDb(path.join(tmpDir, skippedLabel, `${skippedLabel}.db`))
+      try {
+        const rows = [
+          { promptName: "tts-azure", modelId: "azure/azure-tts", cacheHit: false, success: true, durationMs: 900 },
+          { promptName: "tts-azure", modelId: "azure/azure-tts", cacheHit: false, success: true, durationMs: 0, skippedReason: "no-speakable-text" },
+        ]
+        for (const data of rows) {
+          db.run(
+            "INSERT INTO llm_log (timestamp, step, item_id, data) VALUES (?, ?, ?, ?)",
+            [new Date().toISOString(), "tts", "pg001_t001", JSON.stringify(data)]
+          )
+        }
+      } finally {
+        db.close()
+      }
+
+      const res = await app.request(`/api/books/${skippedLabel}/debug/stats`)
+      expect(res.status).toBe(200)
+      const body = await res.json()
+      const ttsStep = body.steps.find((s: { step: string }) => s.step === "tts")
+      expect(ttsStep.calls).toBe(1)
+      expect(ttsStep.cacheMisses).toBe(1)
+      expect(ttsStep.avgDurationMs).toBe(900)
+      expect(body.totals.calls).toBe(1)
+
+      // The row still has to be inspectable — it explains the missing audio.
+      const logsRes = await app.request(`/api/books/${skippedLabel}/debug/llm-logs`)
+      const logs = await logsRes.json()
+      expect(logs.logs).toHaveLength(2)
+      expect(
+        logs.logs.some(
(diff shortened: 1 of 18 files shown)


gbergengruen:
## Blocked on #800 — rebase needed tomorrow

#799 merged as `ba720b13` and #800 is now retargeted to `develop`. This PR sits at the top of that stack and needs a rebase once #800 is rebased.

**Order matters — do #800 first**, then:

```bash
git checkout fix/tts-mark-no-audio-entries
git rebase --onto fix/speech-fail-fast-missing-credentials d975d4c3
pnpm typecheck && pnpm test
git push --force-with-lease
```

Base stays `fix/speech-fail-fast-missing-credentials`, so no retarget call is needed here — only the rebase.

⚠️ Until then this PR's diff replays both #799's (already merged) and #800's changes, because its head sits on `fd439eb8`. Not worth reviewing in that state.

Full step-by-step for the stack, plus the test plan and two risks I already cleared, is in the [comment on #800](https://github.com/unicef/adt-studio/pull/800#issuecomment-5319992605).

---

Unrelated: the findings from the #799 manual test pass are now filed as #808 (root cause) · #809 · #810 · #811 · #812. Worth a look at **#809** in particular before this one lands — it concerns Core TTS normalization stability, which is adjacent to the no-audio marking this PR does.

PR #816, opened 2026-08-19
refactor(studio): light FlowTopBar for wizard, import & project menu

Eliezir:
Replaces the last classic dark chrome (`StudioTopBar`, `bg-gray-700`) with a light, redesign-consistent home affordance across the three flows that still used it.

### The one design
`FlowTopBar` — the ADT logo **is** the home button: a back arrow slides in on hover (200ms ease-out) + subtle highlight; click returns to the Library (`/`). The page title follows. It owns the frameless-window drag region, macOS traffic-light spacing, and the Windows/Linux window controls.

Named `FlowTopBar` (not `TopBar`) to avoid colliding with the shell's existing `title-bar/TopBar`, which is a transparent window-controls overlay for the sidebar screens — a different job.

### Surfaces updated
- **Add-book wizard** (`/books/new`) — upload, preset, and step phases
- **Import** (`/books/import`)
- **Project menu** — the book-editor sidebar-top strip now uses the same logo-home affordance in light tokens (was `bg-gray-700`)

### Notes
- `StudioTopBar` removed (−127 lines).
- Settings/Update icons were dropped from these focused create/import flows (always reachable from the shell); the locale switcher stays.
- i18n: new `Back to library` string translated in all locales (en, pt-BR, es, fr, sq).
- typecheck ✓ · lint ✓ (0 errors) · i18n 0 missing.

Stacked on `eliezir/import-prototype-sidebar` (#734).


elasticsounds:
**Dark-mode follow-up (out of scope for this PR):** The new `FlowTopBar` was tested in dark mode and renders correctly — it uses theme tokens (`bg-background` / `text-foreground` / `border-border`) and adapts cleanly, so the top bar itself is good.

However, dark-mode testing showed that the **add-book wizard body and many modals throughout the app are still hardcoded to light backgrounds** (`bg-white` / `bg-[#f5f5f5]`) and need to be matched to the theme. Tracking as follow-up work.

elasticsounds:
This is great. I agree with the path to remote the Settings and update tracker from the top bar to prevent confusion during the wizard. This is tested and approved. Note that in another update - the dark mode must be reflected throughout the rest of the interface.

elasticsounds (approved):
Approved.

PR #821, opened 2026-08-24
fix(runtime): keep the reader chrome alive across page turns when served

### LINKED ISSUE #820: fix(runtime): reader chrome tears down and rebuilds on every page turn
## What happens

In an exported book, every page turn tears the reader UI down and rebuilds it. The dock,
pagination, TOC, glossary and audio controls all disappear and come back on each navigation,
so the reader sees a distinct three-stage flash:

1. blank page (content is emitted with `opacity-0`)
2. skeleton dock, no content
3. content appears, dock populates

This happens on every prev/next click, every TOC jump, every keyboard page turn, and every
"Next" from a completed activity.

## Why

Each book page is its own HTML document (`pg001_sec001.html`, `pg002_sec001.html`, …), and a
page turn is a real `window.location.href` change. The reader chrome is mounted by the runtime
into `#interface-container` and `#nav-container` on each document, so it cannot survive the
navigation.

The usual fix for this is a shell that holds the chrome while only the content changes. What
rules that out is the requirement that the bundle work when `index.html` is opened **straight
from disk**, with no server — the double-click case. That requirement makes every page
independently bootable by design, so there is nowhere for a persistent shell to live.

That design was formalised in #184 (*Add SCORM 1.2 + offline `file://` support*), which is why
the packaged output looks the way it does today:

- `base.bundle.local.js` is an **IIFE** build rather than ESM, because
  `<script type="module">` is blocked for `file://` origins. Only the `?embed=1` preview, which
  is always served over HTTP, still gets the ESM bundle.
- `offline-preloader.js` inlines every manifest **and every page's HTML** into a generated JS
  file and monkey-patches `window.fetch`, because `fetch()` cannot reach sibling files under
  `file://`.
- Each page loads that trio (`offline-preloader.js`, `scorm.js`, `base.bundle.local.js`)
  independently, with no shared parent document.

To be precise about the history: #184 did not introduce the flash. Before it, each page loaded
`base.bundle.min.js` as a module — still once per document, still rebuilding the chrome every
time. What #184 did was lock in the direct-from-disk guarantee, so any fix has to *add* a
served-mode path rather than replace the current one.

The multi-document structure itself also cannot go away: EPUB needs a spine of separate XHTML
documents, PNLD/VALIDE mirrors the same layout, and SCORM reports per page. The problem is not
that the pages are separate files — it is that all of the per-document setup runs from scratch
on every turn.

## Cost per page turn

Measured against a packaged book (25 pages, fixed-layout):

| Work repeated on every page turn | Cost |
|---|---|
| Re-parse + execute the runtime bundle | 751 KB raw / 236 KB gzipped |
| Re-mount two React roots | full tree, twice |
| Re-fetch config, interface translations, `pages.json`, `toc.json`, glossary, timecodes | 6 requests, all awaited |
| **Actual new page content** | **~1.3 KB** |

`#content` stays hidden until all six of those fetches resolve, which is what makes the blank
stage visible rather than instantaneous.

Median time from click to new content visible, driven through the dock in Chromium:

| Condition | Time |
|---|---|
| localhost, unthrottled, warm cache | 11 ms |
| 4× CPU throttle, 5 Mbps, 40 ms latency | 74 ms |

The gap between those two rows is the point. The flash is barely perceptible on a fast local
connection and clearly visible on the low-bandwidth, low-spec devices this project is meant to
reach — the cost scales with exactly the conditions we care most about.

## Other consequences

- **Read-aloud** cannot continue across a page turn. `isPlaying` is persisted to storage purely
  so the next document can resume playback from item 0, which is a workaround for the reload
  rather than real continuity.
- **Scroll position and transient UI state** are lost by definition.
- The runtime already carries mitigation for this — `PagePrefetcher` emits `<link rel="prefetch">`
  and a `speculationrules` prerender for the next page — which helps forward navigation in
  Chrome/Edge only, and not at all under `file://`.

## Scope

Affects the targets that embed the reader runtime: the standalone **adt (web)** bundle and
**SCORM**. EPUB, webpub and PNLD call `stripRuntimeBundle()` and are rendered by a host reader,
so they are unaffected.

## Expected

Turning a page should not visibly rebuild the reader UI. The chrome should stay put while only
the page content changes — without giving up the ability to open `index.html` directly from
disk.

Eliezir:
Closes #820.

Includes two follow-on fixes found while reviewing this change against the
running app — see the commits for detail:

- **Studio's preview stopped tracking the current page.** `PreviewView` read it
  only from the iframe's `load` event, which no longer fires once navigation
  happens in place. The accessibility card froze on the entry page, and reviewer
  validation records were persisted under that stale `page_id`.
- **The page controls drifted in the dock** — the counter widened by a digit at
  page 10, an unmatched `pl-1` made the flanks 4px uneven, and `min-w-64`
  overflowed below ~700px.

## What this does

Served over HTTP, the reader now keeps its document alive across a page turn.
`navigateToPage()` fetches the target page and swaps only what differs — `<main>`,
the page-scoped `<head>` nodes, and the `<body>` presentation attributes — then
re-binds the runtime through `initializePageContent()`, extracted from
`bootRuntime` so a page reached by soft navigation and the same page reached by a
full load run through one implementation.

The chrome, both React roots, every atom and the audio element stay put.

## What this deliberately does not do

- **`file://` is untouched.** Double-clicking `index.html` still hard-navigates.
  `canSoftNavigate()` requires an `http(s):` origin, so the direct-from-disk
  guarantee from #184 is unaffected.
- **The packaged output format is unchanged.** Pages are still standalone HTML
  documents. EPUB, webpub and PNLD call `stripRuntimeBundle()` and are not
  touched by any of this.
- **No new dependencies.**

Every failure path falls back to a full document load: an unreachable page, a
response that is not a book page, or a thrown error. There is no state where a
failed swap strands the reader.

## Measurements

Median of 8 page turns driven through the real dock in Chromium, against two
packaged books (one fixed-layout, one with 15 activity sections). Measured click
→ new content committed.

| Condition | After | Before |
|---|---|---|
| localhost, unthrottled, warm cache | **3 ms** | 11 ms |
| 4× CPU throttle, 5 Mbps, 40 ms latency | **5 ms** | 74 ms |

The soft-nav figure barely moves between rows — the swap re-fetches nothing —
which is the property that matters for low-bandwidth deployments.

With the View Transition crossfade enabled the same turn measures ~30 ms. That
is animation, not latency; the swap work itself is the 3–5 ms above. The
transition is skipped under `prefers-reduced-motion`.

## The parts worth reviewing closely

A document load used to reset everything for free. Several assumptions the
codebase never had to state are now load-bearing:

**`pageEpochAtom`** — the invalidation signal for anything caching nodes out of
`#content`. The concrete case: `useAudioPlayer` builds its TTS queue in a
`useMemo` keyed on `[audioFiles, translations, speechTexts, easyReadMode,
describeImagesMode]`, none of which change on a page turn.

Without the epoch, read-aloud keeps narrating the page you just left — and word
highlighting stops appearing entirely, because it is applied to nodes that are
no longer in the document. Demonstrated by reverting the fix and tracing which
audio file plays: on the real book, sitting on `pg002003` while still playing
`pg001_n0005.mp3` and advancing further through page 1's queue.

**Activity disposers** — all 12 `initializeXActivity()` functions return one and
`bootRuntime` discarded all 12, which was harmless when the document was about
to die. `word-bank`, `quiz` and `stepper` bind `keydown` / drag handlers to
`document`. Now collected and run before re-initializing.

**Script neutralization** — whether an imported `<script>` auto-executes on
insertion is an ambiguous corner of the spec and engines disagree: Chrome leaves
`DOMParser`-sourced scripts inert, jsdom runs them on attach. Scripts are marked
`type="application/adt-inert"` before insertion and revived explicitly, so
execution is exactly-once by construction rather than by engine behaviour. They
are revived *in place* — a custom activity finds its own section with
`document.currentScript.closest('section')`.

**Custom-activity registrar ordering** — after a swap the runtime has already
booted, so `window.adtRegisterCustomActivity` still points at the *previous*
page's closure. The stub is reinstalled before the new DOM lands, reproducing
document-load ordering. Getting this wrong is quiet: the activity renders fine
but its Submit button never enables.

**Page changes are reported, not inferred** — `scorm.js` read
`meta[name="title-id"]` once and committed on `beforeunload`, so it would have
reported only the entry page. It now listens for `adt:page-changed`. Matomo
likewise gets an explicit `trackSpaPageView()` with `setCustomUrl`, or every
page after the first is attributed to the entry page.

**Accessibility** — focus moves to `<main>` (`tabindex="-1"`, non-interactive so
`:focus-visible` does not match and mouse users see no ring) and the heading goes
to the existing `#sr-announcement` region. WCAG 2.4.3 and 4.1.3.

## Verification

| Property | How | Result |
|---|---|---|
| No document reload | `window` sentinel + marker on a live dock node, 10 turns | both survive |
| Activities bind to new DOM | soft-nav into a multiple-choice page vs hard-loading it, then click an option | identical state; Submit enables |
| No listener accumulation | `document.addEventListener` instrumented, net delta over 10 turns | 0 |
| Read-aloud follows the page | start playback on pg002, turn, poll for the highlighted node | highlights pg005; node connected, inside live `#content` |
| Screen-reader announcement | read `#sr-announcement` and `activeElement` after a turn | heading announced, focus on `<main>` |
| Exactly-once script execution | inert scripts left behind after a swap | 0 |
| Back / forward | `goBack()`, assert section and sentinel | restores without reloading |

**218 test files / 2776 tests passing**, including 16 new ones for the swap.
`pnpm typecheck` clean. `pnpm lint` shows only the 9 pre-existing `apps/studio`
warnings.

## Known limits

- **jsdom cannot verify script execution timing.** It evaluates inserted scripts
  through an async resource queue, so side effects are not visible on the turn a
  swap resolves. The unit tests assert this module's contract — every inline
  script revived exactly once, in its authored position — and leave actual
  execution to the browser run.
- **`pushState` under `file://` was not exercised.** Mode A is selected before it
  would matter. If soft navigation is ever extended to `file://`, confirm current
  Chrome behaviour first; the offline preloader's `fetch` patch is
  protocol-agnostic, so the fetch half already works there.
- **Read-aloud panel dismissal on page turn is preserved.** A swap no longer
  needs that workaround, but keeping it means soft and hard navigation stay
  behaviourally identical. Worth changing as a UX decision, not as a side effect
  of this PR.


elasticsounds:
**Review notes for `eliezir/reader-ui-flash-navigation`**

**Overall.** The approach is sound. Over HTTP the document stays alive and only `<main>`, the page-scoped head nodes and the body attributes are swapped before the runtime re-binds. Chrome, audio continuity, SCORM, Matomo, back/forward and the Studio preview are all handled. All 57 new and touched tests pass. The cross-fade itself is fine for our readers. It is opacity only and about 250 ms, so keep it. Four fixes are requested before merge, one limitation is recorded, and one thing should be checked.

**1. The in-app "Reduce motion" toggle does not stop the page transition.** `swapToPage` only consults the OS media query at `page-swap.ts:339`. The toggle sets a `reduce-motion` attribute on `<body>` and the stylesheet zeroes animations under body, but the view-transition pseudo-elements hang off `<html>`, so that CSS cannot reach them. Read the atom as well:

```ts
import { getDefaultStore } from "jotai"
import { reduceMotionAtom } from "@/shared/state/ui.atoms"

const reduceMotion =
  (getDefaultStore().get(reduceMotionAtom) as boolean) ||
  window.matchMedia("(prefers-reduced-motion: reduce)").matches
```

Please add a test for both signals. The current tests stub `matchMedia` to false and never define `startViewTransition`, so neither the animated path nor the reduced-motion path is covered.

**2. Suppress the focus ring on `<main>` after a keyboard page turn.** Moving focus to `<main>` is the right behaviour for screen reader and keyboard users and should stay. But after an arrow-key turn the browser treats the programmatic focus as keyboard-driven and draws its default ring around the whole content column. Verified in Chrome: click the page and press Right Arrow, the ring appears; click Next with the mouse and it does not. `<main>` is not a control, so the ring carries no information. One line next to the tabindex at `page-swap.ts:245`:

```ts
main.style.outline = "none"
```

**3. A throw inside the animated commit is swallowed.** With a view transition, `commit` runs a frame later, after `swapToPage` has already returned "ok". If it throws, the browser captures it in the transition's promise, nobody reads it, and the hard-load fallback never fires. The reader is left with a pushed URL, a half-initialised page and no error toast. The non-animated path catches the same throw and falls back. Reproduced with `window.scrollTo = () => { throw new Error("boom") }` in the console: the animated path logs "Uncaught (in promise)", the reduced-motion path logs the fallback warning and reloads. Await the callback promise so both paths behave the same:

```ts
const viewTransition = (
  document as Document & {
    startViewTransition?: (cb: () => void) => {
      updateCallbackDone: Promise<void>
      finished: Promise<void>
    }
  }
).startViewTransition
const reduceMotion = window.matchMedia("(prefers-reduced-motion: reduce)").matches
if (viewTransition && !reduceMotion) {
  // Wait for `commit` itself, not the animation. A throw inside it rejects
  // this promise, reaches the catch below, and the caller falls back to a
  // hard load, identical to the non-animated path.
  await viewTransition.call(document, commit).updateCallbackDone
} else {
  commit()
}

return "ok"
```

Regression test for `page-swap.test.ts`, inside the `swapToPage` block:

```ts
it("reports failure when the commit throws inside a view transition", async () => {
  Object.defineProperty(document, "startViewTransition", {
    configurable: true,
    value: (cb: () => void) => {
      const done = Promise.resolve().then(cb)
      return { updateCallbackDone: done, ready: done, finished: done }
    },
  })
  mockFetchOnce(pageHtml({ sectionId: "pg002_sec001", title: "Page two" }))
  initializePageContent.mockImplementationOnce(() => {
    throw new Error("boom")
  })
  try {
    await expect(swapToPage("http://localhost/book/pg002_sec001.html")).resolves.toBe("failed")
  } finally {
    delete (document as { startViewTransition?: unknown }).startViewTransition
  }
})
```

**4. Leaving a stepper page wipes the next page's background colour.** `commit` swaps the body attributes first and disposes the previous page's activities later, inside `initializePageContent`. The stepper's effect cleanup at `activity-stepper.tsx:468` restores the body colour it found on the page being left, which overwrites the incoming page's `data-background-color`. Dispose the previous activities at the top of `commit`, before `swapBodyAttributes`. Exporting `disposeActivityInitializers` from lifecycle and calling it first is enough. The call inside `initializePageContent` can stay, since the list is empty by then. The lifecycle mock in `page-swap.test.ts` needs the new export.

**5. Offline bundles do not benefit.** Not a blocker for this version, recorded so nobody expects otherwise. `canSoftNavigate` returns false for anything but http and https, so a double-clicked `index.html` still flashes on every page turn. There is a viable fix: track the page in the URL fragment instead of the path, since browsers allow `pushState` on `file:` URLs to change only the query or fragment, and the offline preloader already serves every page's HTML from memory. That is a separate issue for later, not this PR.

**6. Something to check: hosts that follow the iframe.** WebPub in Thorium and the EPUB export both behaved normally in testing. Still worth a look: any host that tracks progress by watching the iframe load, or its URL, will stop seeing turns. The runtime now fires `adt:page-changed` on its own document, which only same-origin hosts can hear. If cross-origin hosts matter, a `postMessage` to the parent from `commit`, mirroring the preview-settings pattern, covers it in about ten lines. Not requested here, just flagged.

**Still being verified on device.** The viewport meta is now swapped live between reflowable and fixed-layout pages, so zoom state on iOS and Android needs a look. `scrollbar-gutter: stable` reserves a permanent gutter on Windows and Linux, so check fixed-layout pages for a blank strip on the right.

PR #822, opened 2026-08-25
feat: notify users when a pipeline stage finishes and the window is unfocused (#280)

### LINKED ISSUE #280: [Electron] Notifications for when a pipeline step is running and the window is not focused
## Summary

Pipeline runs can take a long time. The app already plays a notification sound when a pipeline finishes, but that alone doesn't cover two common cases: the user has no audio, or they've stepped away from the computer entirely. There's also no feedback for users who are in the app but navigated away from the book that is currently processing.

## Proposed behavior

Two types of notifications:

1. **System notification** *(the main ask)* — When a pipeline step finishes and the ADT window is **not focused at all**, fire an OS-level system notification with the step name and result (completed or failed). This works even if the user is in another app or away from the PC.

2. **In-app notification** *(suggestion)* — When a pipeline step finishes and the user **is** focused on the app but has navigated to a different book or section, show a subtle in-app toast/banner so they know something finished without having to go back to check.

## Why this matters

- Users with audio off get no feedback today
- Users away from the PC miss the sound entirely
- Users multitasking within the app across multiple books have no passive feedback

RaphaelPhillipe07:
Closes #280 

## Goals

- Fire a native OS notification with the stage name and result (`completed` / `failed`) when the Electron window is not focused.
- Show a subtle in-app toast when the window is focused but the user is in another book or section.
- Avoid duplicate feedback when the user is already looking at the stage that just finished.
- Keep cancellation behavior unchanged: cancelling a run is a deliberate action and must not beep, toast, or notify.

---

## Behavior

| Scenario | Result |
|---|---|
| Stage completes/fails, window **not focused** | Native OS notification with localized stage name + result |
| Stage completes/fails, window focused, user in another book | In-app toast with `View` action |
| Stage completes, window focused, user in the same book but another section | In-app success toast with `View` action |
| Stage fails, window focused, user in the same book | Existing per-book error toast (no duplicate) |
| Stage completes/fails while user is already on that stage | Existing chime/announce/error toast (no duplicate) |
| Run cancelled | No notification, no toast, no sound (unchanged) |

---

## Design decisions

- **Stage granularity, not atomic step granularity.** The existing completion chime fires when a stage completes, and multi-step stages (e.g. Extract has 8 steps) would otherwise generate notification spam.
- **Backend emits stage-level terminal events.** The frontend used to compute `stageJustCompleted` locally from step events. That computation cannot be reused by a global listener without duplicating per-book state, so `stage-complete` and `stage-error` are now first-class `ProgressEvent` variants emitted by the stage runner.
- **Global SSE stream instead of per-book subscriptions.** A root-level `EventSource` for `/books/events` keeps one connection open regardless of route changes and emits events for every book, with the book label included.
- **Electron main process owns focus detection and native notifications.** The renderer asks `window:is-focused` and calls `notifications:show`, so the OS notification uses the native `Notification` API and the actual `BrowserWindow.isFocused()` state.
- **Duplicate suppression is route-aware.** The global hook compares the event label/stage against the current route and lets the existing per-book hook handle feedback when the user is already on that stage.

---

### Backend

- `packages/types/src/progress.ts`
  - Added `stage-complete` and `stage-error` to the shared `ProgressEvent` discriminated union.
- `apps/api/src/services/stage-runner.ts`
  - Emits `stage-complete` after each stage runner resolves.
  - Emits `stage-error` when a stage throws, after recording/emitting the underlying step errors.
  - Cancellation still rethrows without emitting either event.
- `apps/api/src/services/book-event-bus.ts`
  - Added `addGlobalListener`, allowing listeners to receive `(label, event)` for every book.
  - Existing per-label listeners are unchanged.
- `apps/api/src/routes/book-events.ts` (new)
  - Adds `GET /books/events`, an always-on SSE stream for all books.
  - Wraps each event with its `label` so the client knows which book the event belongs to.
- `apps/api/src/app.ts`
  - Mounts `/books/events` **before** `createBookRoutes` so it is not captured by `/books/:label`.

### Electron

- `apps/desktop/src/main/ipc/notifications.ts` (new)
  - `notifications:show` → displays a native `Notification`.
  - `window:is-focused` → returns `BrowserWindow.fromWebContents(event.sender)?.isFocused()`.
- `apps/desktop/src/main/index.ts`
  - Registers `registerNotificationsIpc()` alongside the other IPC handlers.
- `apps/desktop/src/preload/index.ts`
  - Exposes `window.api.notifications.show()` and `window.api.notifications.isWindowFocused()`.
- `apps/desktop/src/preload/index.d.ts` and `apps/studio/src/vite-env.d.ts`
  - Added the matching `NotificationsApi` types.

### Studio

- `apps/studio/src/hooks/use-global-run-notifications.ts` (new)
  - Root-level hook mounted once in `__root.tsx`.
  - Subscribes to `/books/events`.
  - On `stage-complete` / `stage-error`:
    - If unfocused → OS notification with localized stage label and result.
    - If focused and on a different book/section → success/error toast with a `View` action that navigates to `/books/$label/$step`.
    - If already on that stage → no duplicate toast.
  - Works in web mode with `document.hasFocus()` as a fallback when the Electron bridge is unavailable.
- `apps/studio/src/routes/__root.tsx`
  - Mounts `useGlobalRunNotifications()` once.
- `apps/studio/src/locales/{en,es,fr,pt-BR,sq}.po`
  - Added and translated 4 strings:
    - `{stageLabel} completed`
    - `{stageLabel} completed in {label}`
    - `{stageLabel} failed`
    - `{stageLabel} failed in {label}`

---

## Files changed

**Backend**
- `apps/api/src/app.ts`
- `apps/api/src/routes/book-events.ts` (new)
- `apps/api/src/services/book-event-bus.ts`
- `apps/api/src/services/stage-runner.ts`
- `packages/types/src/progress.ts`

**Electron**
- `apps/desktop/src/main/index.ts`
- `apps/desktop/src/main/ipc/notifications.ts` (new)
- `apps/desktop/src/preload/index.ts`
- `apps/desktop/src/preload/index.d.ts`

**Studio**
- `apps/studio/src/hooks/use-global-run-notifications.ts` (new)
- `apps/studio/src/routes/__root.tsx`
- `apps/studio/src/vite-env.d.ts`
- `apps/studio/src/locales/{en,es,fr,pt-BR,sq}.po`

**Tests**
- `apps/api/src/services/book-event-bus.test.ts` (new)
- `apps/api/src/services/stage-runner.test.ts` (extended assertion for `stage-complete`)

---

## Edge cases

- **Cancellation**: `stage-error` is never emitted for a cancelled run; existing cancel behavior is preserved.
- **Same book, same stage**: the global hook skips toast/notification to avoid duplicating the existing chime, screen-reader announcement, and error toast.
- **Same book, different section**: success shows a toast because the existing per-book hook only chimes on success.
- **Different book**: both success and failure show a toast, since the per-book hook for that book is unmounted.
- **Web build (non-Electron)**: the OS notification path is skipped; the in-app toast still works via the global SSE stream.

---

## Notes

- Notifications fire at **stage** granularity, matching the existing completion chime.
- The `View` toast action navigates using the stage slug, which matches the existing `/books/$label/$step` route.
- All new user-visible strings are wrapped with Lingui and translated in all supported locales.


Eliezir:
Thanks for contributing this notification flow — the cross-book feedback is a useful addition. Before this can land, could you address these two items?

1. At `apps/studio/src/hooks/use-global-run-notifications.ts:43`, the global `EventSource` is created once and never recreated. It retries normal network disconnects, but a non-2xx or invalid SSE response permanently changes its state to `CLOSED` ([WHATWG specification](https://html.spec.whatwg.org/multipage/server-sent-events.html)). If the API or proxy later returns a fatal response such as `502` or `503`, cross-book notifications remain disabled for the rest of the session.

Please add a closed-state reconnect handler with capped backoff. A possible implementation around the existing `handleProgress` handler:

```ts
let eventSource: EventSource | null = null
let reconnectTimer: ReturnType<typeof setTimeout> | undefined
let reconnectAttempt = 0
let disposed = false

const scheduleReconnect = () => {
  if (disposed || reconnectTimer !== undefined) return

  const delay = Math.min(30_000, 1_000 * 2 ** reconnectAttempt++)
  reconnectTimer = setTimeout(() => {
    reconnectTimer = undefined
    connect()
  }, delay)
}

const connect = () => {
  if (disposed) return

  const source = new EventSource(`${BASE_URL}/books/events`)
  eventSource = source

  source.addEventListener("open", () => {
    reconnectAttempt = 0
  })

  source.addEventListener("progress", handleProgress)

  source.addEventListener("error", () => {
    if (
      disposed ||
      eventSource !== source ||
      source.readyState !== EventSource.CLOSED
    ) {
      return
    }

    scheduleReconnect()
  })
}

connect()

return () => {
  disposed = true
  clearTimeout(reconnectTimer)
  reconnectTimer = undefined
  eventSource?.close()
}
```

This only retries fatal closures; normal transport failures remain handled by the browser’s built-in retry behavior.

2. Make OS notifications optional. The new hook currently always shows OS notifications when the window is unfocused, with no setting to disable them. Please add a separate, `localStorage`-backed OS-notification preference in Settings, defaulting to enabled. This setting should gate only the Electron OS-notification call; the existing app-toast preference (the toast preferences are not merged into develop yet, but they are being built in the new inicial ui pr) and `/books/events` subscription should remain unchanged.

PR #839, opened 2026-09-02
Fix storyboard "Starting…" status, Codex image input, and CLI schema refs

elasticsounds:
The Storyboard stage now emits `step-start` for `web-rendering` before the render loop, so the sidebar shows it running immediately instead of "Starting…" until the first page finishes its render and visual-refinement loop (previously several minutes on slow or retry-heavy models, since the step row was only created on the first `step-progress`).

The Codex provider now accepts inline images: each one is written to the per-turn scratch directory and attached with `codex exec --image`, with a numbered `[Attached image N]` marker in the prompt, and the manifest declares `imageInput`; previously every image-bearing step (Metadata onward) failed with "Codex structured output cannot take inline images".

The shared CLI JSON-schema converter now inlines `$ref`s that do not point at a top-level definition, which zod-to-json-schema emits whenever one Zod instance is reused (e.g. `HeadingLevel` on outline entries and style clusters) and which OpenAI's structured-output validator behind `--output-schema` rejects; genuine recursive refs are kept.

Base64 image media-type detection moves into a shared helper used by both the Codex and Claude agent providers. Verified with typecheck, the api and llm test suites (including new tests for each change), and live Codex runs of an image prompt and the fixed Book Outline schema.


elasticsounds:
## Review summary

This PR now contains three related commits: the original storyboard/Codex compatibility fixes, Studio-driven Codex CLI sign-in, and the Gemini activity-agent model correction.

### What changed

- Storyboard now emits `step-start` for `web-rendering` before the render loop, so slow first-page renders do not leave the UI at `Starting…`.
- Codex structured-text calls now support inline images through per-turn scratch files and `codex exec --image`. Base64 media-type detection is shared with the Claude agent provider.
- CLI JSON Schema conversion now inlines reusable non-recursive `$ref` values that Codex structured output rejects, while preserving genuine recursive references.
- Codex CLI is resolved from `CODEX_EXECUTABLE`, PATH, common install paths, or the ChatGPT desktop bundle. The provider intentionally ships no Codex SDK or bundled executable.
- Settings can start and cancel `codex login`, display the fallback browser URL, detect completion, and run logout. The Codex CLI owns the OAuth callback and token storage; ADT never reads the tokens.
- CLI OAuth is advertised only in Electron/local development. State-changing login, cancel, and logout routes require `X-ADT-CLI-Action`, forcing browser CORS preflight and preventing drive-by CSRF.
- Provider health distinguishes API-key authentication from a local CLI login. Immediate login completion refreshes health correctly.
- The platform default remains `openai:gpt-5.4`; Codex remains opt-in. `CODEX_API_KEY` still outranks a local ChatGPT login.
- When a Google model is the book default, Generate Activity now resolves to `google:gemini-3.1-pro-preview`. The failed local request was already using `gemini-2.5-pro`; Google returned that it is unavailable to new users and recommended 3.1 Pro Preview.

### Suggested review focus

1. `apps/api/src/routes/providers.ts` and `apps/api/src/services/cli-login-service.ts`: runtime gating, CSRF header enforcement, process cancellation, and one-time status delivery.
2. `packages/llm/src/providers/codex/`: executable discovery, login URL parsing, JSONL handling, image cleanup, and credential precedence.
3. `apps/studio/src/hooks/use-cli-login.ts` and `CliLogin.tsx`: polling transitions, immediate completion, health invalidation, and API-key-versus-CLI state.
4. `packages/llm/src/providers/google/index.ts`: the activity-agent default changes independently of the selected structured-text model. Gemini 3.1 Pro Preview requires a paid API tier.

### Verification

- Full suite after the core Codex login/security changes: 255 test files, 3,314 tests passed.
- Latest Gemini delta: 37 targeted tests passed.
- `pnpm typecheck` passed.
- `pnpm lint` passed with 8 pre-existing unused-suppression warnings and no errors.
- `git diff --check` passed.

Issue #718, opened 2026-08-04
Debug Stats tab: failures are not counted and TTS cost is always $0.0000


gbergengruen:
### Preliminary Checks
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am filing a bug report.

### What happened?

Two separate blind spots in the debug panel's Stats tab:

**1. The Errors column misses outright failures.** The aggregate query in [`apps/api/src/routes/debug.ts`](../blob/develop/apps/api/src/routes/debug.ts) counts errors as:

```sql
SUM(CASE WHEN json_array_length(json_extract(data, '$.validationErrors')) > 0 THEN 1 ELSE 0 END) as errorCount
```

TTS failures set `success: false` and leave `validationErrors` unset, so they are counted as zero errors. The `llm_log` table has both a `success` and an `error_count` column, and the query uses neither.

**2. TTS cost is always \$0.0000.** [`apps/studio/src/components/debug/StatsTab.tsx`](../blob/develop/apps/studio/src/components/debug/StatsTab.tsx) computes cost from token counts with hardcoded per-million rates:

```ts
const cost = (inputTokens * 2.5 + outputTokens * 10) / 1_000_000
```

TTS records no tokens (it isn't a token-billed API), so every TTS row contributes nothing. ElevenLabs bills per character and Azure/OpenAI per character or request, none of which this formula can express.

### Expected behavior

The Errors column reflects calls that actually failed, and TTS spend is either estimated or clearly marked as not-estimated rather than displayed as \$0.00.

### Proposed solution

For the error count, switch to the recorded outcome (`$.success = 0`, or the `success` column directly).

⚠️ **This changes existing numbers for LLM rows too, and should be a deliberate decision rather than a silent fix.** Today a call that hit validation errors but eventually succeeded on retry counts as an error; keying on `success` would stop counting those. Arguably more correct — "Errors" should mean failures — but worth agreeing on before changing, and possibly worth splitting into two columns (failures vs. retried).

For cost: per-provider rates rather than one global token formula. This depends on the companion issue that records TTS character counts — without it there is no quantity to price. Until then, showing "—" for TTS would be more honest than \$0.0000.

### What area is this issue related to?

Other

### Additional context

Found while adding request-parameter logging for ElevenLabs TTS (PR #641). Both problems predate that PR and affect all providers.

Related: the Log tab shows the same failure blind spot in a different place (companion issue on the green "Success" dot).

PR #849, opened 2026-09-10
fix(speech): report Gemini TTS failures instead of hiding or papering over them

### LINKED ISSUE #846: Gemini TTS (gemini-2.5-flash-preview-tts): intermittent failures, inconsistent voice across generations, and Other blocks
## Summary

The `gemini-2.5-flash-preview-tts` TTS model (voice `Puck`) shows intermittent failures and voice inconsistency across calls, both in batch mode and per-sentence mode. This happens even with `temperature` and `seed` fixed, which suggests the issue is on Gemini's side rather than in ADT's pipeline.

## Environment

- Model: `gemini-2.5-flash-preview-tts`
- Voice: `Puck`
- Endpoint: `generativelanguage.googleapis.com/v1beta/models/{model}:generateContent`
- Modality: `AUDIO`
- Processing modes tested: Batch (full text) and per-sentence

## Prompts used

**Prompt 1 (accent/style):**
```
Habla en espanol rioplatense de Uruguay (Montevideo), no mexicano ni castellano peninsular.
Articula las vocales de forma natural y relajada, sin excesos.
Entonacion natural montevideana, relajada y conversacional.
Manten una prosodia neutra pero sin musicalidad mexicana.
Manten la voz estable y directa, con una altura de voz media-neutra constante, sin vibrato notable ni agudos marcados.
Ritmo conversacional uruguayo, uniforme entre oraciones.
```

**Prompt 2 (consistency):**
```
Genera SOLO el audio para este texto exacto. Actúa como un narrador profesional de audiolibros uruguayo.
REGLA CRÍTICA: Mantén un volumen, tono de voz, respiración y velocidad absolutamente
constantes y uniformes. No agregues dramatismo, emoción ni fluctuaciones. Tu voz debe
sonar plana, técnica, consistente y estable en todo momento.
```

**Test text (real excerpt used, ~180 words, Rioplatense Spanish):**
> "LA HISTORIA QUE NOS UNE En dos mil quince nos embarcamos en la tarea de poner en práctica nuestra prédica del aprender haciendo. [...] Otro de los objetivos es acompañar al docente en la integración de la novela con su propuesta didáctica."
> *(full text available on request; omitted here for length)*

## Steps to reproduce / Tests performed

| # | Mode | Temperature | Seed | Result |
|---|------|-------------|------|--------|
| 1 | Batch (full text) | 0.0 | 42 | ADT error: `Gemini TTS response did not include audio data` |
| 2 | Per-sentence | 0.0 | 42 | ADT error: `Gemini TTS response did not include audio data` |
| 3 | Batch (full text) | 0.4 / 0.8 | 42 | Audio is generated, but with different intonation and even **different voice** (male/female) across runs. Whisper processes twice but doesn't cut precisely: some sentences end up without audio. |
| 4 | Per-sentence | 0.4 / 0.8 | 42 | Random behavior: some sentences change intonation/voice (male↔female); others return `Gemini TTS response did not include audio data`. |
| — | Both | 0.4 / 0.8 | 42 | In other runs, the response indicates an `Other` block reason (no audio, no `inlineData`). |

## Expected behavior

- With `temperature: 0` and a fixed `seed`, the voice and result should be deterministic, or at least not fail systematically.
- The selected voice (`Puck`) should remain consistent across calls for the same book/project, without alternating between male and female timbre.
- The service should not block generation with an `Other` reason without reporting a specific cause.

## Actual behavior

- With `temperature: 0`, most calls (batch and per-sentence) fail with `Gemini TTS response did not include audio data`.
- With `temperature: 0.4` or `0.8`, audio is obtained in some cases, but:
  - The voice identity changes (male/female) between sentences or between runs of the same text.
  - Sporadic `did not include audio data` failures persist on individual sentences.
  - `Other` block reasons appear.

## Impact

- It is not possible to reliably generate audio for full books because voice consistency is not maintained across calls.
- Per-sentence processing worsens the problem (more calls → higher chance of failure or voice change).
- There is currently no stable alternative within the Gemini Developer API (AI Studio) that replicates the `Puck` voice with consistency guarantees; Cloud Text-to-Speech / `synthesizeLongAudio` does not accept Gemini AI Studio API keys and would require a separate GCP integration.
- **Token/cost spend is a particular concern**: every failed attempt (`did not include audio data`, `Other` block, or audio with inconsistent voice that must be discarded and regenerated) consumes Gemini tokens without producing a usable result. On top of that, Whisper reprocessing when cuts aren't precise duplicates compute cost without achieving audio consistency, whether the root cause lies in the Gemini models or in Whisper.

## Additional notes

- We ruled out the issue being in ADT's own code: the failure was reproduced with a client independent of ADT, using an equivalent payload (same model, voice, `generationConfig`), also getting HTTP 500 with the full page text.
- With short texts (~31 characters), generation works correctly, suggesting the failure is related to request size/load and/or instability in Gemini's backend under certain conditions.
- We tried configuring `batch_by_page: true` and `batch_max_chars: 300` to reduce the size of each call; this did not resolve the voice inconsistency across calls.

gbergengruen:
Fixes #846.

## What was upstream, and what was ours

The no-audio failures and the voice drift are Gemini's. An [upstream report](https://discuss.ai.google.dev/t/gemini-2-5-tts-model-not-working-at-low-temperatures/137850) pins the no-audio case precisely: Gemini's TTS models **fail at temperature ≤ 0.4 and work from 0.5**, returning `candidates: [{ finishReason: "OTHER" }]` with no content — and failing *slowly* (~148s, against ~21s for a success). Google's own docs separately list voice inconsistency and drift on long outputs as known preview limitations, and document neither `temperature` nor `seed` for TTS.

What was ours is that we made those worse, hid them, and recommended the setting that triggers them.

## Evidence, from the reporter's own book

Still on disk in `books/Uruguay_Ceibal_1930-el-viaje-2`:

- Ten of fifteen page-9 slices were effectively silent — three of them exactly 44 bytes, a WAV header with zero audio — and **every one was recorded as a success**.
- The `llm_log` row for that run reads `(page pg009, 60 entries)` for a **15-entry page**: the transcript was 5,603 chars with the page text repeated verbatim four times, at offsets 0/1401/2802/4203. Gemini narrated ~3 passes and stopped; because ids repeat, the silent tail overwrote the good slices.
- That run logged `durationMs=128275`, matching the upstream slow-failure signature.

Replaying the cached page audio and Whisper result through the new aligner: **11 of 52 ranges** would have produced empty or near-empty WAVs. Zero API calls needed — both inputs are still cached.

## Changes

| | |
|---|---|
| Dedupe page-group entries | The actual trigger. Loudly — a duplicate here means something upstream is wrong |
| Parse `promptFeedback.blockReason` / `finishReason` | The exact upstream payload was previously invisible; the reporter could only say "an Other block reason" because that was all our error could tell them |
| Never write empty slices | `computeEntryTimeRanges` reports alignment provenance; `generatePageSpeechFiles` returns unwritable slices as `unaligned` |
| Report Gemini failures | Removed the `provider !== "gemini"` guard on `step-error`; word highlighting no longer skipped book-wide over one failure |
| Don't retry a deterministic failure | Was 5 × ~2.4 min per item for the same answer — the reporter's token-spend complaint |
| Stop after the first proven failure | 50-page book: ~2 hours → ~2.4 min |
| Warn below the temperature floor | In the UI as you type, and before the request |
| Scope Gemini sampling to Gemini books | Temperature/seed were showing for OpenAI/Azure/ElevenLabs books |
| `speech.batch_max_chars` | The reporter set this exact key; it didn't exist and Zod dropped it silently |

| Remove the cross-provider fallback | Single-item regeneration used to re-narrate a failed entry with whatever other provider had a key — a different provider *and* a different voice, silently |
| Add a synthesis preamble | The Gemini prompt opened with `### PERFORMANCE` and never asked for speech; this book prepends 824 chars of pronunciation rules, so the notes were ~94% of it |

## No silent substitution

Per review discussion, this deliberately does **not** paper over failures:

- The flash↔pro model swap is **removed**, not extended to full runs. Handing back audio from a model the user didn't choose takes control of the output away from them and hides a fixable setting.
- Unaligned entries are reported, never quietly re-synthesized.
- Low temperature is warned about but still **honoured** — it's the user's setting. The abort is scoped to a cause that provably applies to every request; a content-specific refusal still fails only that one entry, even when a low temperature is configured at the same time.
- **The cross-provider fallback is gone too.** It was the same substitution one layer down: a failed entry re-narrated by another provider in that provider's voice, with nothing to tell the user. Holding an OpenAI key is not consent to narrate with OpenAI.

## Verification

3443 tests pass, `pnpm typecheck` clean, `pnpm lint` 0 errors, all five locales at 0 missing.

**Run against the reporter's own book** (`Uruguay_Ceibal_1930-el-viaje-2`, pages 9-10, 19 entries, `batch_by_page: true`), with `audio/es-UY` and `.cache/tts-page` cleared first:

| | |
|---|---|
| `temperature: 0` | **2 Gemini calls, `attempt=1` each** (95.9s and 155.3s — the upstream slow-refusal signature). Every entry recorded with `finishReason=OTHER` *and* the setting to change. Before this branch the same run was 10 calls: 5 retries per page for an answer that never changes. |
| Temperature cleared | 19/19 entries, **zero 44-byte WAVs** (smallest real slice 119 KB). Gemini answered in 44.7s and 51.5s. |
| Single-item regenerate at `temperature: 0` | HTTP 502 carrying Gemini's own reason. One Gemini call, **no OpenAI substitution** — and the existing entry was left untouched at `provider: gemini, voice: Kore`. |

The successful run also exercises the new prompt preamble end to end: this book sets `es-uy` instructions, so every request carried it and Gemini narrated normally.

UI walkthrough: the temperature warning appears as you type `0` and clears at `0.5`; the Sampling block is nested inside the Gemini card and absent for non-Gemini books; `batch_max_chars` shows its clamp warning below 120 and survives toggling page batching off and on.

## Still open after this PR

The voice-identity half of #846 — Gemini returning a different (sometimes male↔female) voice between calls — is upstream and **not fixed here**. `temperature`/`seed` don't control it, and this PR documents that rather than pretending otherwise. `batch_max_chars` reduces *within*-recording drift but adds requests, and the voice can differ between them; the helper text and `config.yaml` now say so. That half deserves its own issue rather than being closed with this one.


gbergengruen:
**Status: paused here, staying in draft. Picking this up tomorrow.**

Code is complete and green (3437 tests, typecheck clean, lint 0 errors, locales 0 missing). What's left is verification and two open decisions.

### Left to do

- [ ] **Finish the UI walkthrough.** Part 1 (settings — no keys, no run) is done. Part 2 needs a real page-batched run on `Uruguay_Ceibal_1930-el-viaje-2`: temperature 0 to see it fail fast and explain itself, then temperature cleared to confirm no 44-byte WAVs. Both books are reset to *before* the Speech stage, so TTS Normalization has to run first, and `audio/es-UY` + `.cache/tts-page` need clearing or the run reuses the old artifacts.
- [ ] Decide on the two items below.

### Open decisions

**1. The cross-provider fallback chain.** `getSingleItemFallbackAttempts` in `apps/api/src/routes/tts.ts` is the same silent substitution as the flash↔pro swap removed in this PR — arguably worse, since it re-narrates with a different provider *and* a different voice, so the user gets a narrator they never picked. Left in place with a `TODO(#846 follow-up)` because removing it is wider than this issue. By the principle this PR follows, it should probably go the same way.

**2. The `es-uy` instruction block.** `config/speech_instructions.yaml` prepends 824 characters of pronunciation rules and IPA to *every* Gemini request — ~94% of the prompt for a typical 50-char sentence, repeated on every call. Google's prompting guide warns that a prompt without a clear "synthesize speech" preamble gets classifier-rejected or read aloud, and `### PERFORMANCE` labels the notes without ever instructing synthesis. Plausible contributor to the `OTHER` blocks *and* pure repeated spend. Worth measuring before changing.

### Known gaps, deliberately not addressed here

- **What duplicated the catalog is still unknown.** Both books' current `text-catalog` nodes have zero duplicate ids, so the version carrying the 4× has been regenerated and the source isn't reproducible from these artifacts. The dedupe guard stops the damage and warns, but the upstream cause deserves its own issue.
- No live probe script. The ADT-side defects are covered offline and the upstream half is settled by the forum thread; the probe would only answer the two open questions above.
- `pipeline-dag.ts` (CLI path) has no page batching and no Gemini retry at all.
- TTS cost is never recorded (`buildTtsLogEntry` emits no `usage`), so the spend concern stays unmeasurable — #718.

PR #862, opened 2026-09-14
feat(publish): add Studio publishing and reader feedback flow

Eliezir:
Restores the full publishing Studio experience and the published-reader feedback runtime.\n\nIncludes the publishing stage, publication dashboard, reader comments and presence UI, and the temporary loader artwork test bench. Depends on #858.


Eliezir:
Closed — this PR's contents were redistributed into the existing stack, so it now has nothing unique in it (0 files changed).

| Where it went | What |
|---|---|
| #853 `feat/publish-cloudflare-ui` | all 119 `apps/studio/**` files — the publish stage, dashboard and run animation, `publication-feedback/`, `publications/`, the temporary `/dev/publish-loaders` route, styles and catalogs, plus the Studio data layer (`api/client.ts` publication/feedback/raw-page-image methods and the five hooks) |
| #857 `feat/publish-book-publication` | the `publish-service.ts` change that enables `features.comments` in the published bytes — moved back down from #858, where it had landed during the earlier rushed update |
| #858 `feat/publish-api-routes` | the 77 reader files — `apps/adt-runtime/**` comments and presence, plus the regenerated `assets/adt/**` interface translations and Tailwind build |

The Studio data layer went into #853 rather than #855 as originally sketched: #855 sits above #853 in the stack and its own content is the API-side `publish-worker-client.ts`, so splitting the hooks out would have left ~40 files in #853 with unresolvable imports. #853 now typechecks and passes its 605 studio tests on its own.

Three pre-existing failures in this branch were fixed on the way through:

- `usePageAspect.ts` called `image.decode()`, which does not exist outside a real browser — it threw synchronously past the `.catch()` and took 10 tests down with it. It now falls back to the `load` event where `decode()` is missing.
- `provisioner.test.ts` still documented one shipped migration while this stack ships three.
- Five `apps/publish-service/src/room*.ts` files ended on a blank line.

Verified on the stack tip: `pnpm typecheck` clean, `vitest run` 324 files / 4276 tests passing, `pnpm lint` 0 errors, `git diff --check` clean.

PR #867, opened 2026-09-15
chore(staging): validate Cloudflare publishing stack

Eliezir:
Temporary integration PR for the Cloudflare publishing stack. It points at the top stack commit and exists only because the staging workflow requires a PR targeting `develop`. Do not merge this PR; the individual stack PRs remain the review and merge path.


elasticsounds:
## Feature review — Cloudflare publishing & sharing

**Amazing feature.** This works almost seamlessly and flawlessly. All you need is a free Cloudflare account to publish and collect comments — it's really remarkable. The settings and publishing integration is great. Below is my feedback.

1. **Dark mode** — The Cloudflare settings page does not work in dark mode. Needs to be updated.

2. **More destinations (stubbed)** — On the publishing page you list Cloudflare; you should ALSO list **GitHub** with a "coming soon" pill, and **Local Network** as well. Do not implement GitHub or Local Network — just add them with the "coming soon" pill.

3. **"Publish" → "Share"** — Change "Publish" everywhere to "Share". Technically we are not publishing, we are sharing for comments.

4. **Shared-state indicator (sidebar)** — After the site is shared, the icon in the left-hand menu stays grey; it should turn coloured to indicate it is shared.

5. **Comment count indicator** — After a user leaves a comment, the publish icon needs an indicator — preferably with a number — to show how many comments have been made.

6. **Comment → storyboard navigation & resolve** — Clicking a comment in the Cloudflare view jumps to the storyboard, but it links only to the first page and the comment itself isn't visible. Ideally, clicking a comment should go to the correct storyboard page, and we should be able to see the comment and toggle it on/off. There is also no way to "resolve" a comment in the publishing panel or the storyboard panel. **Resolve** should effectively archive comments, indicating they are done.

7. **Comment numbering is per-page and misleading** — In the Cloudflare preview, the comment numbers on each page don't match the numbers in the right column. If you add two comments on a later page, then come back and add two comments on the first page, comment "3" will show as comment "1" on the first page — the indexing seems to restart at one on every page. I would remove the numbering from comments completely for now and just have "a comment".

8. **Page numbering is confusing** — The page number attached to each comment is confusing. Say you create a book from pages 17–20 of a PDF: it says the comment is on Page 17, but the preview's bottom numbering says "1". Also the storyboard says "15" while the preview says "19" while the slide says "1". We need clearer, definitive numbering for simplicity. I think we should label every comment definitively **by page and section — the same numbering used in the URL**. Please check.

9. **"Waiting on you" can't be expanded** — In "Waiting on you", you can't open it to see the other comments — it shortcuts you to the storyboard but you can't see the comment. You can only see the top 3 comments, not the others.

10. **Onboarding / discoverability of commenting** — When you first enter the cloud view, it's not obvious that there are "comment" and "view comment" buttons at the bottom. We need simple onboarding. My thought: a popup bubble that says "Leave a comment"; clicking it lets you leave a comment, and the bubble has an **X** so it's easy to close. ALTERNATIVELY, copy Google: put the comments button in the top right, and when you open the right panel it shows a large "Add comment" button in the middle for your first comment. ALSO, when the right panel is open there should be an "Add comment" shortcut button within that view that is permanently accessible.

11. **Shared link on the home screen** — The link to the Cloudflare page needs to be on the home screen with each book — we should be able to view the shared link in the popup window, and it should be visually indicated as shared on the book's panel itself.

PR #876, opened 2026-09-17
fix(cli): render skipped steps as skipped, not complete

### LINKED ISSUE #811: fix(cli): skipped steps render as complete (100% progress bar) on a failed run
## Summary

The CLI renders a **skipped** step exactly like a **completed** one — it fills the progress bar to 100%. A failed run can therefore display a fully green Speech stage at `100/100` when no speech was generated at all.

Discovered while reviewing #799. Not caused by that PR.

## Mechanism

[`cli.ts:63-67`](https://github.com/unicef/adt-studio/blob/ba720b13/packages/pipeline/src/cli.ts#L63-L67) handles both event types in one branch:

```ts
if (event.type === "step-complete" || event.type === "step-skip") {
  const bar = bars.get(event.step)
  if (bar) bar.update(bar.getTotal())
  return
}
```

`step-skip` is a distinct event in the progress schema ([`progress.ts:22`](https://github.com/unicef/adt-studio/blob/ba720b13/packages/types/src/progress.ts#L22)) and is emitted in ~15 places in the stage runner, so the information is available — the CLI just discards it.

## Observed behaviour

A run failed upstream, so the Speech stage was skipped. The CLI printed a filled bar reading `100/100` for Speech. The only indication anything was wrong was the "Failed stages" footer at the end.

## Contrast: the UI gets this right

[`use-book-run.ts:288-292`](https://github.com/unicef/adt-studio/blob/ba720b13/apps/studio/src/hooks/use-book-run.ts#L288-L292) maps the two events to different states:

```ts
} else if (d.type === "step-complete" || d.type === "step-skip") {
  ...
  const nextStepState: StepState = d.type === "step-skip" ? "skipped" : "done"
```

and `StepState` has a dedicated `"skipped"` member. So the fix is to give the CLI a skipped rendering — **not** to change the event or the emitters, which are already correct.

## Suggested direction

Render skipped steps distinguishably: leave the bar unfilled with a `skipped` label, or use a distinct marker/colour. Whatever is chosen, a skipped step must not be visually identical to a completed one.

Worth checking at the same time whether the "Failed stages" footer is the only failure signal in a long run, and whether it is prominent enough.

edubraqd:
## What this does

The CLI handled `step-complete` and `step-skip` in the same branch and filled the bar to its total for both, so a stage skipped after an upstream failure showed a full bar reading `100/100`. The event schema already tells the two apart and the studio UI maps `step-skip` to its own `skipped` state; only the CLI threw that away.

`step-skip` now leaves the bar empty and replaces the `value/total` column with `skipped`. The per-step format becomes a format function so it can read that flag from the bar payload; nothing changes for `step-complete` or `step-progress`.

```
   Extract        [██████████████████████████████] 1/1
   Speech         [░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░] skipped
   Progressed     [████████████░░░░░░░░░░░░░░░░░░] 4/10
```

Emitters are untouched, as the issue asks. The "Failed stages" footer question raised there is left open; this only fixes the misleading bar.

Fixes #811

## Testing

- `pnpm typecheck` clean.
- `vitest run packages/pipeline`: 67 files, 1257 tests, all passing. `cli.ts` has no tests of its own and `createCliProgress` is not exported, so no new test; the render above is from driving the same format function through `cli-progress` 3.12 with a captured stream.

Written with help from Claude Code; I reviewed and ran everything myself.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Eliezir:
Hey! It looks like this PR is targeting `main` instead of `develop`. `develop` is the repository's default branch, and targeting `main` is pulling a large number of unrelated changes into the diff.

Could you update the base branch to `develop`?

edubraqd:
Done, base is now develop. The branch was already on top of develop (43814f2), so the diff is just the one change in packages/pipeline/src/cli.ts. Sorry for the noise.

Eliezir (approved):

